PR #52, opened 2020-11-08
Update bike.cpp

kenands:
"Power Table for Wattbike pro" seems to be more near of power provided by kinomap with bike 500
(https://support.wattbike.com/hc/en-us/articles/115001881825-Power-Resistance-and-Cadence-Tables)
With kinomap:
- At resistance level 1 and 80 rpm, power was about 110w 
- At resistance level 15 and high rpm (up to 150) I reached 1200w


cagnulein:
awesome!

PR #189, opened 2021-03-06
M3i support

p3g4asus:
Please use [2-dot comparison](https://github.com/cagnulein/qdomyos-zwift/compare/master..p3g4asus:M3iSupport) to see the changes this PR introduces.
This PR adds the support to Keiser M3i bike. It is a connectionless bluetooth device that supports only data output and not command input. To receive data you have to continue scanning for BLE devices and parse their manufacturer data. This device requires 3 settings (included in the qml):

1. Bike ID: the ID of the bike configured to be sure to use the data of the correct bike just in case there are more than one in the same room
2. Speed buffer length: The speed of the bike is not sent to output by the bike but its self calculated using distance and time with a moving average filter whose size is this buffer length (default 60)
3. Use Qt search on android (default false): unfortunately qt BLE search API under android performs very poorly in finding all m3i advertisings. So I implemented the use of BLE scan android API (which works much better). You can however choose to use QT search anyway (highly discouraged). Android BLE scan requires `ACCESS_FINE_LOCATION` permission (added in the manifest and in `main` for runtime ask). The scan requires also two additional small java classes (added in `DISTFILES` inside project file) and an additional cpp class (added as additional android source and header in the project file).

This PR contains also:

- Some minor fixes, optimizations and code indentation style change (I am sorry for that but I prefer working with KR code indentation style) in `fitshowtreadmill` class
- A minor change in `bluetooth` module where I think its preferable to use a `QMutableListIterator` to traverse the `QBluetoothDeviceInfo` list to replace in the list the elements you find multiple times and keep it updated.
- An improvement in debug log system: the date is added in the `myMessageOutput` log handler inside `main` rather than in `debug` slot. This way even the logs produced with `qDebug()` will have date and time header.
- A fix in a function of treadmill.cpp (return type is missing)


p3g4asus:
Please ignore this PR. I amopening an other on. I made an important change that I have to include.

PR #190, opened 2021-03-06
M3i support

p3g4asus:
Please use [2-dot comparison](https://github.com/cagnulein/qdomyos-zwift/compare/master..p3g4asus:M3iSupport) to see the changes this PR introduces.
This PR adds the support to Keiser M3i bike. It is a connectionless bluetooth device that supports only data output and not command input. To receive data you have to continue scanning for BLE devices and parse their manufacturer data. This device requires 3 settings (included in the qml):

1. Bike ID: the ID of the bike configured to be sure to use the data of the correct bike just in case there are more than one in the same room
2. Speed buffer length: The speed of the bike is not sent to output by the bike but its self calculated using distance and time with a moving average filter whose size is this buffer length (default 60)
3. Use Qt search on android (default false): unfortunately qt BLE search API under android performs very poorly in finding all m3i advertisings. So I implemented the use of BLE scan android API (which works much better). You can however choose to use QT search anyway (highly discouraged). Android BLE scan requires `ACCESS_FINE_LOCATION` permission (added in the manifest and in `main` for runtime ask). The scan requires also two additional small java classes (added in `DISTFILES` inside project file) and an additional cpp class (added as additional android source and header in the project file).

This PR contains also:

- Some minor fixes, optimizations and code indentation style change (I am sorry for that but I prefer working with KR code indentation style) in `fitshowtreadmill` class
- A minor change in `bluetooth` module where I think its preferable to use a `QMutableListIterator` to traverse the `QBluetoothDeviceInfo` list to replace in the list the elements you find multiple times and keep it updated.
- An improvement in debug log system: the date is added in the `myMessageOutput` log handler inside `main` rather than in `debug` slot. This way even the logs produced with `qDebug()` will have date and time header.
- A fix in a function of treadmill.cpp (return type is missing)


cagnulein:
thanks @p3g4asus for your amazing work! i will release the the 2.4.9 and then i will merge this.

cagnulein:
@p3g4asus also could you please add the peloton conversion in your keiser file?

you should follow this conversion table

image

p3g4asus:
> 
> 
> @p3g4asus also could you please add the peloton conversion in your keiser file?
> 
> you should follow this conversion table
> 
> image

Sure I will. Is there an example bike class inside QZ to follow?

cagnulein:
another question, i don't know if you already did in this way: could you make the native android scan (as you already did) and the QT scan for iOS? if i saw correctly, you have now a static define to choose it. I guess this define should just follow the OS, so if android, using the native one, if anything else (iOS, raspberry...) use the QT one. What do you think?

cagnulein:
> Sure I will. Is there an example bike class inside QZ to follow?

check this https://github.com/cagnulein/qdomyos-zwift/blob/master/src/echelonconnectsport.cpp#L178 @p3g4asus

p3g4asus:
> 
> 
> another question, i don't know if you already did in this way: could you make the native android scan (as you already did) and the QT scan for iOS? if i saw correctly, you have now a static define to choose it. I guess this define should just follow the OS, so if android, using the native one, if anything else (iOS, raspberry...) use the QT one. What do you think?

With the current version of m3ibike, Under Android OS QT scan can be chosen instead of android scan but android scan will be performed by default. For all other OSes QT scan will be used.

p3g4asus:
> 
> 
> > Sure I will. Is there an example bike class inside QZ to follow?
> 
> check this https://github.com/cagnulein/qdomyos-zwift/blob/master/src/echelonconnectsport.cpp#L178 @p3g4asus

Thank you!

p3g4asus:
@cagnulein [peloton conversion added](https://github.com/p3g4asus/qdomyos-zwift/blob/M3iSupport/src/m3ibike.cpp#L554)

cagnulein:
> peloton conversion added

thanks. maybe in the next version we can linearize it with excel in order to have a realtime conversion.

cagnulein:
@p3g4asus after resolving the conflicts, now i see correctly the files here https://github.com/cagnulein/qdomyos-zwift/pull/190/files

cagnulein:
@p3g4asus could you please retest the master branch? just to check if i didn't commit any error solving the conflicts ;)

p3g4asus:
@cagnulein I just checked the merge and it looks to me just fine. No errors.
Could you please apply this patch to `m3ibike.cpp` that adds peloton resistance linearization? I was just about committing it. I didn't even know that resistance was a floating point number.
```diff
--- a/src/m3ibike.cpp	2021-03-06 13:44:44.000000000 +0100
+++ b/src/m3ibike.cpp	2021-03-06 13:33:13.000000000 +0100
@@ -26,13 +26,16 @@
     OP(17,18,60,V,D)\
     OP(19,20,70,V,D)\
     OP(21,22,80,V,D)\
     OP(23,24,90,V,D)
 
 #define SET_PELOTON_RESISTANCE(MINV,MAXV,ZWIFTV,KEISERV,DEST)\
-    else if (KEISERV>=MINV && KEISERV<=MAXV) DEST = (double)ZWIFTV;
+    else if (KEISERV>=MINV && KEISERV<=MAXV) {\
+        angular_coeff = 10.0/(MAXV+1-MINV); \
+        DEST = angular_coeff * KEISERV + ZWIFTV - angular_coeff * MINV; \
+    }
 
 KeiserM3iDeviceSimulator::~KeiserM3iDeviceSimulator() {
     if (dist_buff) delete[] dist_buff;
     if (dist_buff_time) delete[] dist_buff_time;
 }
 KeiserM3iDeviceSimulator::KeiserM3iDeviceSimulator() {
@@ -507,12 +510,13 @@
 }
 
 void m3ibike::processAdvertising(const QByteArray& data) {
     if (disconnecting)
         return;
     QSettings settings;
+    double angular_coeff;
     int buffSize = settings.value("m3i_bike_speed_buffsize", 90).toInt();
     debug(" << " + data.toHex(' '));
     if (parse_data(data, &k3)) {
         detectDisc->start(6000);
         if (!initDone) {
             initDone = true;
```

cagnulein:
@p3g4asus ok thanks! i'm adding a new bike. i will push it and then i will add this patch too. thanks

cagnulein:
done :)

cagnulein:
hi @p3g4asus could you please attach the ios and android logs for future references? Thanks!

p3g4asus:
Here they are!
My next steps concern further optnimizing android-java scan and studying objective c to implement a more reliable ios scan but it will take quite some time due to the (in my opinion) obfuscated nature of objective C
[android_qt.log](https://github.com/cagnulein/qdomyos-zwift/files/6099594/android_qt.log)
[ios.log](https://github.com/cagnulein/qdomyos-zwift/files/6099595/ios.log)
[android_java.log](https://github.com/cagnulein/qdomyos-zwift/files/6099596/android_java.log)

PR #209, opened 2021-03-23
VIRTUALBIKE Append 0 byte object after HR

ben75020:
Appending a byte with zeros fixes an issue with the "BKOOL application" : the application has an offset of "+1280" on HR PPM value.


cagnulein:
@ben75020 hi Ben, how are you? Are you still using BKOOL? Some users told me that BKOOL can't find QZ anymore. Is it possible?

PR #403, opened 2021-07-17
Allowing 2 or more instances of TemplateInfoSenderBuilder

p3g4asus:
This PR helps completing and deploying the work on chartjs by allowing two or more instances of TemplateInfoSenderBuilder. User templates are inside src/templates. Inner templates are put inside src/inner_templates. user templates settings are customizable in Settings qml. Inner templates settings are to be customized in the code. In this implementation  inner templates webserver uses a random free port that is retrieved in the chart qml to show chartjs template. This PR changes also the folder for external templates in Android for compatibility with the latest versions of Android 11. External templates are to be put in the android/data folder.


cagnulein:
Thanks! i will test it tomorrow i hope!

PR #444, opened 2021-09-25
Fix segfault of `bluetooth::stateFileRead`

d3m3vilurr:
a `device` isn't a `domyostreadmill`. but every treadmill have to be
subclass of `treadmill`
`setLastSpeed`, `setLastInclination` and `odometer` move to
upper class and change to virtual function


cagnulein:
Thanks!

PR #503, opened 2021-11-26
qt5package name change

doudar:
Modified the raspberry pi install instructions as the qt5-default package has been renamed qt5-assistant in the raspbian sources.


cagnulein:
More info here https://forum.qt.io/topic/132190/issue-with-sudo-apt-get-install-qt5-default

Thanks

PR #508, opened 2021-11-30
Adding support to dircon protocol

p3g4asus:
This pull request adds dircon protocol support. It uses a new submodule that is needed to advertise the virtual dircon device on the network. Credits to nitroshare for this module. I just slightly modified it to make it more compatible to android. The settings structure for virtual device has been changed. Now it has a bluetooth and a dircon section that can be enabled independently. Also virtual bike module was slightly changed to allow code reuse between dircon virtual device and bluetooth virtual device.
I am sorry but at the moment I am lacking tim to document my modification deeper. Please feel free to ask if you need additional info. I am sorry that I had no time to test it on ios devices.


cagnulein:
Can i delete from the PR the qHttpServerBin's files? I guess they're changing only because you have a different path, isn't it?

cagnulein:
@p3g4asus only an observation: i just compiled the first android build with Dircon and i noticed that the apk size had strongly decreased from 88.7Mb to 47.8Mb (with all the archs so, x86, x86_64, and both arm) :O
I saw that you updated also the openssl commit, if i'm right, maybe that could be the explanation?
Of course is not a problem, it's better of course, but my spider senses have been alerted :D

PR #529, opened 2021-12-27
Implementing dircon protocol for treadmill

p3g4asus:
This pull request implements the dircon protocol for treadmills. It introduces also a small fix in m3i speed calculation that solves a problem that can arise if the bike switches because of timeout during training (for example if the workout is stopped for more than 60 seconds).


cagnulein:
@p3g4asus  if you want to double check the https://github.com/cagnulein/qdomyos-zwift/pull/529/commits/37b439231994a22899cb3313bf06d60d61bafe43
i aligned it to the current master

p3g4asus:
Seems fine to me!

PR #530, opened 2021-12-27
Activate "NordicTrack" compatibility mode from the command line

pascal-bourut:
I added a new parameter (-nordictrack-10-treadmill) for the command line to enable NordicTrack compatibility.
`sudo /home/pi/qdomyos-zwift/src/qdomyos-zwift -no-log -no-gui -heart-service -nordictrack-10-treadmill`


cagnulein:
Does it work so? :)

pascal-bourut:
> Does it work so? :)

Yes, works great. That was indeed a pretty straightforward modification.
Thanks for your help.

Issue #582, opened 2022-01-15
[REQ] Add to WebSocket API Timer Control 


gungun974:
**Is your feature request related to a problem? Please describe.**
In the GUI I can start and stop (and reset) the timer.
However in the api I do not control it and I am forced once the bluetooth connection is made to start immediately.

**Describe the solution you'd like**
Add `start_timer` to start the timer in `templateinfosenderbuilder.cpp`
Add `stop_timer` to stop the timer in `templateinfosenderbuilder.cpp`
Add `reset_timer` to reset the timer in `templateinfosenderbuilder.cpp`
Add maybe if it's not to hard `set_timer` to set timer time in seconds in `templateinfosenderbuilder.cpp`

cagnulein:
@p3g4asus if you want take care of this :)

PR #584, opened 2022-01-15
Add on Websocket start / pause / stop commands

gungun974:
This pull request implements 3 commands to control `qdosmyos` by WebSocket for issue #582.
Start who when `qdosmyos` is on pause resume or otherwise reset timer (and distance...) and start from new.
Pause who only pause `qdosmyos` when pause is not active.
Stop who stop `qdosmyos` and reset all data.

One of my problem of my PR is with the current `bluetoothdevice.h` I can't know if the state is on stop or start so start will always if the state is not on pause reset all data. After this is not very bad cause we can presume on the external side with "workout" state if the state the machine is running or not (if speed is at zero for example).

Second and last problem of my PR is I don't know how communicate from `templateinfosenderbuild.cpp` to `homeform.cpp` to update the UI state when I stop / start / pause. For me this is not a problem cause I don't use the UI and for the user it will not be too cause this API is most reserve for developer.

Anyways I'm open to any changes or review.

### DIFF
--- a/src/templateinfosenderbuilder.cpp
+++ b/src/templateinfosenderbuilder.cpp
@@ -508,6 +508,7 @@ void TemplateInfoSenderBuilder::onStop(TemplateInfoSender *tempSender) {
     device->stop();
     device->setPaused(true);
     device->clearStats();
+    emit workoutEventStateChanged(bluetoothdevice::STOPPED);
     QJsonObject main;
     main[QStringLiteral("msg")] = QStringLiteral("R_stop");
     QJsonDocument out(main);
--- a/src/templateinfosenderbuilder.cpp
+++ b/src/templateinfosenderbuilder.cpp
@@ -476,6 +476,44 @@ void TemplateInfoSenderBuilder::onGetSessionArray(TemplateInfoSender *tempSender
     tempSender->send(out.toJson());
 }
 
+void TemplateInfoSenderBuilder::onStart(TemplateInfoSender *tempSender) {
+    if (!device->isPaused()) {
+        device->clearStats();
+        device->start();
+        emit workoutEventStateChanged(bluetoothdevice::STARTED);
+    } else {
+        device->start();
+        device->setPaused(false);
+        emit workoutEventStateChanged(bluetoothdevice::RESUMED);
+    }
+    QJsonObject main;
+    main[QStringLiteral("msg")] = QStringLiteral("R_start");
+    QJsonDocument out(main);
+    tempSender->send(out.toJson());
+}
+
+void TemplateInfoSenderBuilder::onPause(TemplateInfoSender *tempSender) {
+    if (!device->isPaused()) {
+        device->stop();
+        device->setPaused(true);
+        emit workoutEventStateChanged(bluetoothdevice::PAUSED);
+    }
+    QJsonObject main;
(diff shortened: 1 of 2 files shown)


cagnulein:
> One of my problem of my PR is with the current bluetoothdevice.h I can't know if the state is on stop or start so start will always if the state is not on pause reset all data. After this is not very bad cause we can presume on the external side with "workout" state if the state the machine is running or not (if speed is at zero for example).

there is for sure a bit in the 26 byte that the treadmill sent, that told us if the treadmill is in pause or stop. Generally speaking (i mean talking about all the treadmill that qz supports), i guess we should just use the isPaused() method and, maybe the speed value ( == 0 means that the treadmill is stopped). What do you think?

> Second and last problem of my PR is I don't know how communicate from templateinfosenderbuild.cpp to homeform.cpp to update the UI state when I stop / start / pause. For me this is not a problem cause I don't use the UI and for the user it will not be too cause this API is most reserve for developer.

there is no way actually, but we can leave this point for the future i guess.

@p3g4asus what do you think about this?

Thanks @gungun974

gungun974:
I forgot in my `stop` command to mark the global state of the device changed with `emit workoutEventStateChanged`

p3g4asus:
> @p3g4asus what do you think about this?

I think it is a secondary problem. I will have a look if the GUI and the websocket command can be connected in any way but for now they can be left separated.
Thanks @gungun974 for your hard work.

p3g4asus:
Nice work @gungun974. I think I would add to the JSON response to the start command a field that reports the state in which the device was found before sending it the command. I think it could be potentially useful. Wouldn't it?

gungun974:
It's could be worth. For me personally I don't need that cause I always fetch the "wourkout" event but for a specific case that could be good. 
After is it necessary to send only the pause state (I can't get if the machine is stop or running) or could we just send the same data that "workout" event already do ? 
(For this second options the edit is very simple for me personally cause I will just wrap the data response of the "workout" in is own function "prepareTelemetry" and use this same function in the start / pause / stop to ensure the same data formatage)

p3g4asus:
> After is it necessary to send only the pause state (I can't get if the machine is stop or running) or could we just send the same data that "workout" event already do ? 

I think the same content as workout data would be good too.

gungun974:
@p3g4asus what is the `sessionArray` in `TemplateInfoSenderBuilder` ? Cause I thought that the `workout` event returned the response to the websocket like the actions but its operation and let's say much more different than what I thought. (Like I don't even see a function where it writes to the socket with JSON...)

p3g4asus:
Session Array contains workout data collected in the current activity till the moment is requested. The send code for websockets can be found in webserverinfosender.cpp.

cagnulein:
@gungun974 tell me when you are ready to merge

gungun974:
Unfortunately @p3g4asus I don't have any time to add the JSON response cause I have some job contract to finish for the nexts weeks and if I want to eat I must to stop working on an simple treadmill :sweat_smile:.

But good already the fact of being able to control the start and all that is very good so yes I think @cagnulein you can merge now.

It was an honor for me to have been able to make this feature real :smile:.

cagnulein:
let's merge this, there will be time to improve it :)

PR #587, opened 2022-01-16
Improve the setting of the speed and the slope

december-soul:
This is my first implementation of the GPIO controlled treadmill.
It is a start but there is still something missing.
So it takes some time until my treadmill is set to speed. In the time no speed is transmitted to Zwift.
So changing the speed has to be done in another thread.

for testing it turned out that a trainings.xml file is a good possibility.
#525


cagnulein (approved):
Yes we need to put the whiles with the delay in a separate thread for sure. It could be quite easy to do. But let's do this by steps.
Thanks for the PR.

PR #594, opened 2022-01-20
adjust skierg metrics, and use one notification

siudej:
1. It seems only one subscription is needed.
2. Some metrics added.
3. Speed, cadence and watts forced to 0 when skierg is not active. Otherwise it keeps reporting old stale values.

TODO: Cleanup debug code.

Related issue: https://github.com/cagnulein/qdomyos-zwift/issues/573


cagnulein:
thanks!

PR #627, opened 2022-01-27
Treadmill gpio

december-soul:
Here is my version that I have now tested a few workouts.

I am not yet completely satisfied because a set the speed can take quite a bit longer.
So from 6 km/h to 12 km/h are with me, in 0.1 steps, 60 keystrokes a 400 ms. So over 24 seconds.
Although the treadmill accelerates anyway not soo fast but the time seems to me still very long. But maybe that is not wrong.
I have already thought about how I can do it differently.

Then it turned out that my treadmill doesn't like it when two buttons are pressed at the same time. Therefore, I have now protected the critical section with a semaphore. But it does not make it faster.

Unfortunately I had to reassign the pin numbers because the previous pins do not exist on my system (RasPi B2). The new GPIO pins are available on all boards.

### DIFF
--- a/src/gpiotreadmill.cpp
+++ b/src/gpiotreadmill.cpp
@@ -109,6 +109,17 @@ gpiotreadmill::gpiotreadmill(uint32_t pollDeviceTime, bool noConsole, bool noHea
     refresh->start(pollDeviceTime);
 }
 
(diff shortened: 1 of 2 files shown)


cagnulein:
Great! I like it. Don't you have other preset buttons for the speed on the treadmill? @december-soul

PR #647, opened 2022-02-05
Whatsonzwift

p3g4asus:
This branch implements the navigation on whatsonzwift.com to take zwift training programs and load them inside QZ. This branch introduces also some fixes to fitshow treadmill and m3i bike. It should fix anyrun compatibility activating the new settings inside the fitshow section.

### DIFF
--- a/src/settings.qml
+++ b/src/settings.qml
@@ -183,7 +183,8 @@ import Qt.labs.settings 1.0
 
             property double horizon_gr7_cadence_multiplier: 1.0
 
-            property int  fitshow_user_id: 0x006E13AA
+            property int  fitshow_user_id: 0x13AA
+            property bool  fitshow_anyrun: false
 
             property bool inspire_peloton_formula: false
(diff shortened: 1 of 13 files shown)


cagnulein:
#641

cagnulein:
#646

src/settings.qml
```diff
@@ -183,7 +183,8 @@ import Qt.labs.settings 1.0
 
             property double horizon_gr7_cadence_multiplier: 1.0
 
-            property int  fitshow_user_id: 0x006E13AA
+            property int  fitshow_user_id: 0x13AA
```
cagnulein, line 186: @p3g4asus is the id change something important?

p3g4asus: Not so much. In the last version the two most significant bytes are ignored because user id should be 2 bytes long. The other bytes are for different purposes (and they were for different purposes also for non anyrun treadmills).

cagnulein (approved):

PR #666, opened 2022-02-11
SS2K Shift Step calibration

rjbutler:
SS2K Shift Step calibration

Add calibration of SS2K shift step by resistance level using least squares method.  Sampling of data points in configuration TBD

I'm thinking it could work like this:
1. For each of the data points, QZ would set the resistance on the SS2K.  
2. The user would pedal and tap +/- to adjust the shift step for that data point.   setShiftStep is called with the new value.
3. When the resistances match, they'd tap OK and move on to the next data point.
4. When all 4 data points are set, calibrateShiftStep is called to calculate the slope and intercept which are then saved to settings.


cagnulein:
> @cagnulein Sure, that sounds good. I'll make the change. I can take a look at the settings page too. Are you thinking under Schwinn or SS2K?

SS2K for sure

>  It's possible there will be other bikes that need similar adjustments, so maybe SS2K? Also, for the same reason maybe allow the user to enter both resistance and shift step?

yes that's a good idea!

> 
> For settings, I looked at var and QList for the array but I didn't see you using it anywhere else. Would individual values be your preference? So something like ss2k_resistance_sample1, ss2k_shiftstep_sample1, ...?

Yes unfortunately there is no array, so yes, just use separate variables

> 
> I can write something up for the wiki page too.

Amazing! Thanks!

rjbutler:
I added the samples to settings.   I'll work on instructions for the wiki.

cagnulein:
Thanks @rjbutler I will put the default values of these new settings to 0 or anyway to values that they are leaving the behaviour as was before this mod, so that the compatibility will be kept.
What do you think?

rjbutler:
Yeah, I meant to do that before committing.  Sorry

PR #671, opened 2022-02-12
Fix for anyrun treadmills. They use big endian transmission and not l…

p3g4asus:
Fix to some problems related to anyrun treadmills. Added an option to use timer data from the treadmill rather than the one calculated internally.


cagnulein:
thanks!

cagnulein:
#646

cagnulein:
#641

PR #691, opened 2022-02-23
Improve functionality for Peloton App users

grantland:
* Cadence and Resistance Tiles - Improve readability for bikes using red, limegreen, and orange for < min, <= max, < max respectively
* Peloton Offset Tile
  * Rename to Peloton Remaining/Rem.
  * Add remaining time to make it easier to match with Peloton app
  * Swap remaining time with offset as remaining time is more useful
* Time to Next Tile
  * Add Current Row Elapsed Time as secondary text
  * Enable writing to skip forward/back to next/previous row

Note: Time To Next -/+ along with Peloton Remaining makes it super easy to catch up to the correct timestamp now!

Screenshot_1645774076

PR #713, opened 2022-03-05
Improve Peloton cues

grantland:
Changes:
* (before) `/api/workout/<workout_id>/performance_graph?every_n=10` has a resolution of 10s which means that cues will always be off unless on a strict 10s mark
* (after) `/api/ride/<ride_id>/details` has exact cues
* Fallback using old logic

Test Plan:
* Normal ride cues are perfectly sync'd
* Power Zones & Treadmill do not have any cues (same as before my changes)

### DIFF
--- a/src/peloton.cpp
+++ b/src/peloton.cpp
@@ -213,7 +213,13 @@ void peloton::instructor_onfinish(QNetworkReply *reply) {
     }
     emit workoutChanged(workout_name, current_instructor_name);
 
-    getPerformance(current_workout_id);
+
+    if (workout_name.toUpper().contains(QStringLiteral("POWER ZONE"))) {
+        qDebug() << QStringLiteral("!!Peloton Power Zone Ride Override!!");
+        getPerformance(current_workout_id);
+    } else {
+        getRide(current_ride_id);
+    }
 }
 
 void peloton::workout_onfinish(QNetworkReply *reply) {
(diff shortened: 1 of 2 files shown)


cagnulein:
Hi, treadmills have cues, I'm using them for treadmill workout. Not all of them.

cagnulein:
That's great for ride workout, I will check this tomorrow!

grantland:
> Hi, treadmills have cues, I'm using them for treadmill workout. Not all of them.

Ah, not sure if it was the class I tested on but the new API endpoint I was using didn't seem to support treadmills and the old logic is gated on the bluetooth device (which I don't have).

cagnulein:
Ok I will double check it also with @Janabouc

cagnulein:
@grantland Since I don't have a peloton account, if you're confident with this, I can merge this. What do you think?

janabouc:
I'm not familiar with the problem or solution. How can i help?

cagnulein:
@janabouc just with testing when I will merge this since I don't have a peloton account:)

grantland:
@cagnulein sounds good to me! did a full ride today and the cues were timed perfectly :)

cagnulein:
I will build this today for the rest of the people. @janabouc i will send you this today!

cagnulein:
@grantland i did a little change for #763 could you please double check that everything is ok on your side too?

grantland:
@cagnulein Are you referring to your changes at https://github.com/cagnulein/qdomyos-zwift/commit/02f08f387ea3c3530d5d16e07264cffe90f0509d? If so, I don't see anything in this PR that I made that utilizes `trainprogram::currentRowElapsedTime` or `trainprogram::currentRowRemainingTime` and I don't think anything else really changed in that commit. Do you maybe mean PR #691?

cagnulein:
yes i was referring to that changes. Hah yes maybe i linked to the wrong PR :) anyway, just double check it that I didn't break anything releated to your changes :)

grantland:
Looking at the changes it seems like you're really only adding an offset if `rampElapsed` is set and that's only set in `zwiftworkout`, but not `peloton`, so I think it should be fine!

cagnulein:
Yes, exactly! Perfect, thanks!

Issue #118, opened 2021-01-12
Settings Profile (Peloton, Zwift...)


cagnulein:
https://www.facebook.com/groups/149984563348738/permalink/154451332902061/

PR #714, opened 2022-03-05
Implement Profiles

grantland:
Adds a new feature to support multiple profiles on the same device

* Add a new sidebar menu "Profiles: <profile_name>"
* Enables saving current settings as a new Profile
* Enables quickly switching between Profiles
* Enables deleting a Profile by pressing and holding
* Updates existing settings export filenames to be `yyyyMMdd` to be sorted by recent first

|Sidebar|Profiles|
|-|-|
|Screenshot_1646514307|Screenshot_1646805737|

### DIFF
--- a/src/homeform.cpp
+++ b/src/homeform.cpp
@@ -3714,7 +3714,7 @@ void homeform::saveSettings(const QUrl &filename) {
     QSettings settings;
     QSettings settings2Save(
         path + QStringLiteral("settings/settings_") +
-            QDateTime::currentDateTime().toString().replace(QStringLiteral(":"), QStringLiteral("_")) +
+            QDateTime::currentDateTime().toString("yyyyMMddhhmmss") +
             QStringLiteral(".qzs"),
         QSettings::IniFormat);
     auto settigsAllKeys = settings.allKeys();
--- a/src/homeform.cpp
+++ b/src/homeform.cpp
@@ -4,6 +4,7 @@
 #include "keepawakehelper.h"
 #include "material.h"
 #include "qfit.h"
+#include "simplecrypt.h"
 #include "templateinfosenderbuilder.h"
 #include "zwiftworkout.h"
 
@@ -3709,23 +3710,38 @@ QString homeform::getAndroidDataAppDir() {
 }
 #endif
 
+quint64 homeform::cryptoKeySettingsProfiles() {
+    QSettings settings;
+    quint64 v = settings.value(cryptoKeySettingsProfilesTag, 0).toULongLong();
+    if (!v) {
+        QRandomGenerator r = QRandomGenerator();
+        r.seed(QDateTime::currentMSecsSinceEpoch());
+        v = r.generate64();
+        settings.setValue(cryptoKeySettingsProfilesTag, v);
+    }
+    return v;
+}
+
 void homeform::saveSettings(const QUrl &filename) {
     Q_UNUSED(filename)
     QString path = getWritableAppDir();
 
     QDir().mkdir(path + QStringLiteral("settings/"));
     QSettings settings;
-    QSettings settings2Save(
-        path + QStringLiteral("settings/settings_") +
-                settings.value("profile_name").toString() +
-                QStringLiteral("_") +
-                QDateTime::currentDateTime().toString("yyyyMMddhhmmss") +
-                QStringLiteral(".qzs"),
-        QSettings::IniFormat);
+    QSettings settings2Save(path + QStringLiteral("settings/settings_") + settings.value("profile_name").toString() +
+                                QStringLiteral("_") + QDateTime::currentDateTime().toString("yyyyMMddhhmmss") +
+                                QStringLiteral(".qzs"),
+                            QSettings::IniFormat);
     auto settigsAllKeys = settings.allKeys();
     for (const QString &s : qAsConst(settigsAllKeys)) {
-        if (!s.contains(QStringLiteral("password")) && !s.contains(QStringLiteral("token"))) {
-            settings2Save.setValue(s, settings.value(s));
+        if (!s.contains(cryptoKeySettingsProfilesTag)) {
+            if (!s.contains(QStringLiteral("password")) && !s.contains(QStringLiteral("token"))) {
+                settings2Save.setValue(s, settings.value(s));
+            } else {
+                SimpleCrypt crypt;
+                crypt.setKey(cryptoKeySettingsProfiles());
+                settings2Save.setValue(s, crypt.encryptToString(settings.value(s).toString()));
+            }
         }
     }
 }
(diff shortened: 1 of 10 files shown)


cagnulein:
Wow thanks! I didn't check the code yet, but the current implementation doesn't save password on the settings file for security reason. Are you imagining this as a limit? Maybe we can store password crypted?

grantland:
TBH I only thought about this from an Android perspective, but not other platforms.

For Android, this shouldn't be an issue as we're using internal app storage which isn't globally readable.

For others, this might be an issue if the storage paths are globally readable. I'm not sure encryption would be a good idea since the keys would be both in the open source repository and in the binary. 

Do you know if it's possible to persist additional settings instances the same way the default settings are persisted?

cagnulein:
I'm also thinking about the settings file as something that the users can export on their friends...maybe we can store the password using a key generated on the local device? So no one can know it. What do you think?

grantland:
This actually doesn't interfere with the existing settings export so passwords and tokens will still be stripped on export in `/settings` while Profiles will retain these values in `/profiles`. Do you think my changes are still a problem in this case?

cagnulein:
No your changes will be ok, I was just wondering that massively use the setting load/save mechanism without saving password will lead the users to issues (peloton logins ko,...)
So I was thinking that before adding profiles we need to save passwords. Don't you think?

grantland:
Those values will be saved now since `saveProfile` does not strip them :)

cagnulein:
Yeah I know but we can't save them in this way. I have already handled legal issues and my only defence weapon was that I was not sharing passwords in any way. So we need to be very safe on this. So at least we need to encrypt them with a local key (for example the Bluetooth MAC address or something very hard coded in the device) or we need to tell the user to enter back their password (without saving them in the profile)
Which way do you prefer?

cagnulein:
I'm going to bed right now, almost 11pm here in Italy:)
Have a great day!

grantland:
That makes sense, but I think we can work around this restrictions with "sharing" passwords by not "sharing" them at all.

Instead of "exporting" to a user-accessible dir like `saveSettings()` does with`QStandardPaths::DownloadLocation`/`QStandardPaths::DocumentsLocation`, we could store the profiles in a similar directory that `QSettings` stores application settings by default and we can maintain the same security of the saved profiles as we do with `QSettings settings`.

From the [`QSettings` documentation](https://doc.qt.io/qt-5/qsettings.html#locations-where-application-settings-are-stored), it seems like various platforms persist application settings in the following locations:
* Unix: `$HOME/.config/<organization_name>/<application_name>.conf`
* macOS/iOS: `$HOME/Library/Preferences/<organization_domain>.<application_name>.plist`
* Windows: `HKEY_CURRENT_USER\Software\<organization_name>\<application_name>`
* Android: `/data/data/com.yourAppName/files/.config/<organization_name>/<application_name>.conf` (note: this is not documented, but I found them in my emulator)

So to do this, we could store profiles in [`QStandardPaths::AppConfigLocation`](https://doc.qt.io/qt-5/qstandardpaths.html#StandardLocation-enum) as it is a similar location to `QSettings settings` on each platform:
* Unix: `~/.config/<APPNAME>`
* macOS/iOS: `~/Library/Preferences/<APPNAME>`/`<APPROOT>/Library/Application Support`
* Windows: `C:/Users/<USER>/AppData/Local/<APPNAME>`
* Android: `<APPROOT>/files`

How would this sound?

cagnulein:
I like this @grantland ! It sounds the perfect solution!

grantland:
Updated

Also it turns out that `QStandardPaths::AppConfigLocation` on Android is `/data/data/com.yourAppName/files/settings`

src/homeform.cpp
```diff
@@ -3726,18 +3731,45 @@ void homeform::saveSettings(const QUrl &filename) {
 }
 
 void homeform::loadSettings(const QUrl &filename) {
-
     QSettings settings;
     QSettings settings2Load(filename.toLocalFile(), QSettings::IniFormat);
     auto settings2LoadAllKeys = settings2Load.allKeys();
     for (const QString &s : qAsConst(settings2LoadAllKeys)) {
-        if (!s.contains(QStringLiteral("password")) && !s.contains(QStringLiteral("token"))) {
```
grantland, line 3734: FYI I removed this to enable loading settings from the same method as there's no real reason to omit this on load as `saveSettings()` should already be omitting these values

grantland:
Please hold off on landing this, I'm having some weird errors on the settings page and I'm not sure if my changes are causing them.

grantland:
We're safe, my issue is not related to these changes

cagnulein:
> Updated
> 
> 
> 
> Also it turns out that `QStandardPaths::AppConfigLocation` on Android is `/data/data/com.yourAppName/files/settings`

so it's not the same path of the settings builtin in the qt? if so we need to generate a md5 something for the password because that path is accessible from the user with cx file explorer

grantland:
It is the same path. Anything on `/data/data/com.yourAppName` is app scoped and inaccessible by the user and other apps. Since `QSettings settings` is under `/data/data/com.yourAppName/files/.config/<organization_name>/<application_name>.conf` and `QStandardPaths::AppConfigLocation` is under `/data/data/com.yourAppName/files/settings` and both are subdirs of `/data/data/com.yourAppName`, both are safe.

cagnulein:
#118 
ok i will give it a try!

cagnulein:
last thing @grantland i saw that you're using user_nickname setting to profile name. If you're thinking this about user profile i guess it's ok but i know that the majority of the users will use this to switch quickly to different scenario (treadmill to bike, or zwift to peloton and more). So i guess we should use a new setting instead of using the nickname, what do you think? I can do this by myself if you agree. Thanks

cagnulein:
@janabouc this will change your QZ experience for sure I guess! :)

grantland:
yeah sounds good, created a new `setting.profile_name` to keep it separate from `setting.user_nickname` and put that setting on the `profile.qml`:
Screenshot_1646805737

cagnulein:
i'm testing this on PC and of course i can get the password from the file, i'm too scaried about security issues. I'm adding to this PR the ability to store passwords crypted. I will add the commits here

grantland:
Are you also able to see the password in `HKEY_CURRENT_USER\Software\<organization_name>\<application_name>` on PC?

grantland:
Re https://github.com/cagnulein/qdomyos-zwift/pull/714#issuecomment-1059846178 it seems like PC is the only one where `QSettings setting` doesn't match `QStandardPaths::AppConfigLocation`. The easiest thing in my mind would be to disable it for PC users.

Alternatively, we can also find out how to write to a similar dir as `HKEY_CURRENT_USER\Software\<organization_name>\<application_name>` on PC.

Otherwise the other OSs store in similarly safe dirs.

Also, I'm not really sure how well cryptography would be if we have to store the decryption keys in a similar place as `QStandardPaths::AppConfigLocation`, right?

cagnulein:
what i'm doing is creating a unique id randomly if not present and store it in the settings. Of course an user could read it, but at least this key will not exported in the same file of the profiles and settings and it should be safer. So unwanted password sharing should be covered. Anyway i almost did this, i will commit this shortly so you can check the code

cagnulein:
done @grantland put an eye too on my code. I tested on PC and it seems to work, now i'm trying to check it on android and ios

janabouc:
When I reopened after clicking Save once, the new Profile was displayed as active but wasn't in the list. image
image

janabouc:
So after restart, the newly created profile's name is displayed at the top of the Settings page [Profile: new one] but if you click Profile and go to the selection page, the current selected profile is only in the edit name field. I would expect it be in the list of profiles and highlighted but it is not. Instead Default is highlighted. image

janabouc:
When I saved twice the name appeared in the list. When i clicked the new name ("Save Twice") I got this "Profile Loaded. Would you like to quit?" messageimage.

grantland:
> When I tried to create a new profile on ipad one click of Save did not display new profile until i restarted. Clicking Save twice made it appear without restarting.

> When I reopened after clicking Save once, the new Profile was displayed as active but wasn't in the list.

> So after restart, the newly created profile's name is displayed at the top of the Settings page [Profile: new one] but if you click Profile and go to the selection page, the current selected profile is only in the edit name field. I would expect it be in the list of profiles and highlighted but it is not. Instead Default is highlighted.

> When I saved twice the name appeared in the list. 

I'm not too familiar with QT/QML, but this feature works similar to how settings export/import works for me and is what I copied most of the functionality from. It sometimes takes pressing more than once and the top setting/profile will always be selectd by default.

Not sure if @cagnulein can help chime in on why this happens and if both profiles and setings need changes.

> When i clicked the new name ("Save Twice") I got this "Profile Loaded. Would you like to quit?" message

This is expected as like changing/importing settings you need to restart aftewards and I thought this would make it easier and more obvious.

cagnulein:
yes @grantland this problem is not something that you added, it's probably there since day one :) so i will check this in the next days. Thanks @janabouc for the feedback. I opened a ticket for Profiles bug #725

cagnulein:
@janabouc i guess i addressed all the bug in the 2.10.46

PR #724, opened 2022-03-09
Add src/secret.h to .gitignore

grantland:
Prevent the ability to accidentally commit secrets

### DIFF
--- a/src/homeform.cpp
+++ b/src/homeform.cpp
@@ -36,15 +36,24 @@ using namespace std::chrono_literals;
 #include <QtAndroid>
 #endif
 
-#include "secret.h"
+#if __has_include("secret.h")
+#   include "secret.h"
+#else
+#   define STRAVA_SECRET_KEY test
+#   if defined(WIN32)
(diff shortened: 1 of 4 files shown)


cagnulein:
thanks!

src/homeform.cpp
```diff
@@ -36,15 +36,24 @@ using namespace std::chrono_literals;
...
+#   define STRAVA_SECRET_KEY test
+#   if defined(WIN32)
+#       pragma message("DEFINE STRAVA_SECRET_KEY!!!")
+#   else
+#       warning "DEFINE STRAVA_SECRET_KEY!!!"
+#   endif
 #endif
+
+#ifndef STRAVA_CLIENT_ID
+#   define STRAVA_CLIENT_ID 7976
```
grantland, line 51: I thought it might be a good idea to move this actual value to `secret.h` to make it more obvious that Strava won't work but left it to retain your workflow. Do you think this would be a good idea to move or was there a reason you left this out of `secret.h`?

cagnulein: yes you're totally right!

Issue #718, opened 2022-03-07
[REQ] Adding mid levels of difficulty to Peleton settings


gilliancorbo:
This app has been incredibly helpful in getting my fitness up; thank you so much Roberto Viola! It’s been so helpful that I feel like I’m finally able to progress from the lower level of difficulty in Peleton classes, however the average may be too much of a jump. Is there any way to select something that is halfway beetween lower and average difficulty using auto resistance or do I just need to make that adjustment manually and disable auto resistance until I’m fit enough to tackle the average resistance? 

Maybe a future update could include 5 ranges of difficulty instead of the 3? I’m using an EX-5 with QZ on my iPhone and the Peleton app on my iPad

cagnulein:
@grantland will you find useful too? i was thinking about interpolating the current 3 levels

grantland:
I've actually been working on something a little different that could solve this: https://github.com/grantland/qdomyos-zwift/commit/b24e6e7b29d5b0f8e67995ef45ae5e8411219786

It's more in line with how the Peloton Bike+ functions where each cue change updates the auto-resistance to the same % between prev-lower/pre-upper to post-lower/post-upper.

For instance if lower-upper is 40-50 and you're set at 42 (20% of 10), the next cue change to 30-40 will automatically change your resistance to 32 (20% of 10).

This would also allow you to adjust yourself on the fly as you can use the knob or on-screen controls to change to 35 (50% of 10, same lower-upper of 30-40) and the next cue change to 50-70 will automatically change your resistance to 60 (50% of 20).

I still need to work out some bugs due to echelon <-> peloton conversion is causing auto-resistance drifts and I'm also assuming you'd want this behind a new setting I'd need to add as well :)

cagnulein:
@gilliancorbo will you be ok with the solution proposed by @grantland ? I'm not familiar with Bike+ behaviours ;)

> I still need to work out some bugs due to echelon <-> peloton conversion causing auto-resistance drifts and I'm also assuming you'd want this behind a new setting I'd need to add as well :)

yes thanks :)

cagnulein:
@gilliancorbo Mmm you could be right... @grantland did you try the gears tile too?

grantland:
I haven't used it, but from looking at the gears code it doesn't take into account the the difference between lower-upper and it's just another offset of resistance based on lower, similar to difficulty. 

[...]

PR #777, opened 2022-04-19
Update resistance from trainprogram based on previous resistance %

grantland:
It's more in line with how the Peloton Bike+ functions where each cue change updates the auto-resistance to the same % between prev-lower/pre-upper to post-lower/post-upper.

For instance if lower-upper is 40-50 and you're set at 42 (20% of 10), the next cue change to 30-40 will automatically change your resistance to 32 (20% of 10).

This would also allow you to adjust yourself on the fly as you can use the knob or on-screen controls to change to 35 (50% of 10, same lower-upper of 30-40) and the next cue change to 50-70 will automatically change your resistance to 60 (50% of 20).

Also addresses #718

### DIFF
--- a/src/bike.cpp
+++ b/src/bike.cpp
@@ -5,14 +5,31 @@
 
 bike::bike() { elapsed.setType(metric::METRIC_ELAPSED); }
 
+void bike::changeResistanceRange(int8_t lower, int8_t upper) {
+    // 0.0 is floor and 1.0 is ceil
+    double lastResistanceOffsetPerc = 0.0;
+    if (RequestedLowerResistance >= 0) {
+        lastResistanceOffsetPerc = (Resistance.value() - RequestedLowerResistance) / (RequestedUpperResistance - RequestedLowerResistance);
+    } else {
+        lastResistanceOffsetPerc = 0.0;
+    }
+
+    RequestedLowerResistance = lower;
+    RequestedUpperResistance = upper;
+
+    double calculatedResistance = RequestedLowerResistance + (lastResistanceOffsetPerc * (RequestedUpperResistance - RequestedLowerResistance));
+    changeResistance(calculatedResistance);
+}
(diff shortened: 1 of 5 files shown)


grantland:
Notes:
* I still need to work out some bugs due to echelon <-> peloton conversion causing auto-resistance drifts
* I'm also assuming you'd want this behind a new setting I'd need to add as well :)

src/bike.cpp
```diff
@@ -5,14 +5,31 @@
 
 bike::bike() { elapsed.setType(metric::METRIC_ELAPSED); }
 
+void bike::changeResistanceRange(int8_t lower, int8_t upper) {
+    // 0.0 is floor and 1.0 is ceil
+    double lastResistanceOffsetPerc = 0.0;
+    if (RequestedLowerResistance >= 0) {
+        lastResistanceOffsetPerc = (Resistance.value() - RequestedLowerResistance) / (RequestedUpperResistance - RequestedLowerResistance);
```
grantland, line 12: I think the drift is caused by the fact that I'm using `Resistance.value()` as it's the absolute value emitted from the bike which is most likely an `int`. Due to peloton <-> echelon conversion, `Resistance.value()` might be +/-3 in Peloton units from what `calculatedResistance` is and we lose the granularity.

To fix this, we might need to track whether or not the user changed `Resistance.value()` by turning the knob or using the UI. If the user did not change it we should use `RequestedResistance`, otherwise use `Resistance.value()`

cagnulein:
thanks! i will check the code in the next days, i will be on travel today/tomorrow

src/bike.cpp
cagnulein: one of the most common request was getting the event of the knob movement, in order to change the gear tile. Maybe we can create this event and you can use somehow on this? what do you think?

PR #796, opened 2022-05-01
DIRCON Client #782

cagnulein:
#782


cagnulein:
closed in favour of #3869

PR #811, opened 2022-05-17
first commit that works loading a GPX without the speed and time

cagnulein:
[b.zip](https://github.com/cagnulein/qdomyos-zwift/files/8706893/b.zip)


cagnulein:
https://jsfiddle.net/rmx9v5kw/8/

PR #813, opened 2022-05-24
Initial implement start/stop control of KingSmith R2 when press start/pause/stop buttons

d3m3vilurr:
This patch is not perfect.
If a user doesn't do workout, treadmill will stop automatically,
but an UI of the application couldn't update anything.

Also in the timing issue, sometimes changeing `ControlMode` or `runState` command
just return `Error` packet.
I think, sending message too early (before finishing handshake) could
make the problem.

Fully implement needs passing the signal to homeform or MainApp.
And also it needs more complex state machine for handling buttons.
(eg; press the start button & press the pause button before
finish start treadmill action, pause action never call)


cagnulein:
ok before merging i will wait a more complete implementation. Let me know if I can help you on this somehow

d3m3vilurr:
@cagnulein imo, patchset is dirty hack. but it could explain the concept.

cagnulein:
@d3m3vilurr just to understand: which is the main purpose of this patch? I mean I already managed the pause/stop button of others treadmill using the speed as a control for homeform. all the mechanism specific to the treadmill should be in the treadmill module. Am I missing something?

d3m3vilurr:
Is it possiable that QZ UI receive running status from treadmill?
(I mean, is it possible that treadmill update UI buttons? I thought it can't, because a bluetoothdevice doesn't have homeform instance.)

before the patch, homeform's start/pause and stop button will works independently and it's maintained by UI itself.
after the patch, homeform's buttons will be coupled to treadmill's current status.
(so before; treadmill can be running mode even if UI shown `pause`. but after patch `pause` button only can be shown treadmill is running)

cagnulein:
@d3m3vilurr did you see this https://github.com/cagnulein/qdomyos-zwift/blob/bfb7b747792a73314cc439ed9eb3950b92d2959f/src/homeform.cpp#L2262 ? maybe you should use the same interface inside your module so no changes on the homeform module will be required. What do you think?

d3m3vilurr:
I'm not sure that is perfect or not.
because [button clicking](https://github.com/cagnulein/qdomyos-zwift/blob/bfb7b747792a73314cc/src/homeform.cpp#L310-L311) will make the problem.
these actions will change own status (of homeform instance) then just change the request flag variable of the `bluetoothdevice`
but if device couldn't send the message (unimplement start/stop handling) or returned error message(not finished initializing step or other reason), an application shows wrong buttons.

idk why kingsmith r2 doesn't have the these `auto` methods. (maybe timing problem when i copied from r1 pro)
I'll make the another pr of it

and also, if you want it, i'll split the pr to implement handling start/stop message (maybe mix first and third patch) and others

cagnulein:
argh I didn't test this yet :(

d3m3vilurr:
wow thank you merged this old mr!

d3m3vilurr:
an app is still installed in my phone. just recently i didn't use zwift & treadmill.
it's my free time issue.(yeah.... actually it's just one of excuses. ik..). i'll reuse when i restart workout :)

Issue #856, opened 2022-07-10
Units of class members unknown


drmason789:
BluetoothDevice has some protected members:

```
 metric elapsed;
    metric moving; // moving time
    metric Speed;
    metric KCal;
    metric Distance;
    uint8_t FanSpeed = 0;
    metric Heart;
    int8_t requestStart = -1;
    int8_t requestStop = -1;
    int8_t requestIncreaseFan = -1;
    int8_t requestDecreaseFan = -1;
    double requestFanSpeed = -1;
    double m_difficult = 1.0;
    metric m_jouls;
    metric elevationAcc;
    metric m_watt;
    metric WattKg;
    metric WeightLoss;
    metric Cadence;
    metric Resistance;
    metric METS;
    QGeoCoordinate coordinate;
    metric Inclination;
    metric HeartZone;
    metric PowerZone;

```

Bike has some others.

Could we have comments on these to define what they are and what the units are? 

e.g.

```
    /**
     * @brief Heart The heart rate reported by the device. Unit: beats per minute
     */
    metric Heart;
    
    /**
     * @brief Cadence The time taken for a crank to complete a full revolution. Unit: Radians per second
     */
    metric Cadence;
    
    /**
     * @brief Resistance The resistance requested from the device. Unit: 0..get_MaxResistance()
     */
    metric Resistance;
```

cagnulein:
I'm a bit overwhelmed by requests as you can imagine but if you want we can
make this in this way: ask me when you have some doubts and you can create
a PR for this. It will be much appreciated for the whole community. What
do you think?




Il giorno dom 10 lug 2022 alle ore 19:12 David Mason <
***@***.***> ha scritto:

drmason789:
I'm happy to do the code edits and PR. 

Could we start with these ones? I'm not sure their clear definition and units. Here are some guesses that are probably wrong:

Speed - Flywheel speed. Units: radians per second
Distance - Cumulative flywheel revolutions multiplied by the wheel circumference. Units: meters
Heart - the heart rate. Units: beats per minute
Cadence - the speed at which the crank is turning. Unit: Radians per second
Resistance - the currently requested resistance level. Expected range: 0 to getMaximumResistance()

cagnulein:
> Speed - Flywheel speed. Units: radians per second

km/h

> Distance - Cumulative flywheel revolutions multiplied by the wheel circumference. Units: meters

km

> Heart - the heart rate. Units: beats per minute

yes

[...]

PR #858, opened 2022-07-13
#856 documentation on some members

drmason789:


cagnulein:
Thanks! Do you plan to continue this or do you want to merge this right now?

drmason789:
I'll continue it.

cagnulein:
@drmason789 did you check my observation above?

drmason789:
@cagnulein I don't see an observation. Is it possibly still in draft?

cagnulein:
oh sorry @drmason789 you were right, you should see them now

drmason789:
Thanks, I'll fix those in my next block of work.
What is "pace"? I thought it would be something to do with steps per time unit on a stepper or treadmill. Is it lap time?

src/bluetoothdevice.h
```diff
@@ -145,7 +165,15 @@ class bluetoothdevice : public QObject {
     metric m_watt;
     metric WattKg;
     metric WeightLoss;
+
+    /**
+     * @brief The speed at which the crank is turning. Units: revolutions per minute (RPM)
+     */
```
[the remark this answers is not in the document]
drmason789: Perhaps BluetoothDevice::lastCrankEventTime should be moved to class bike, or renamed to something more generic, e.g. lastActionEventTime.

src/bluetoothdevice.h
```diff
@@ -53,42 +53,38 @@ class bluetoothdevice : public QObject {
...
-     * @brief currentPace Gets the current pace. Units: ?
-     * @return
+     * @brief currentPace Gets the current pace. Units: time per km
      */
     virtual QTime currentPace();
 
     /**
-     * @brief currentInclination
-     * @return
+     * @brief currentInclination The current inclination. Expected range: 0 degrees (horizontal) to 90 degrees (vertical)
```
cagnulein, line 65: actually some treadmill can have also -15%. The max I guess it will never go more than 25% but of course any value could be accepted. Every device module would limit this to its max.

src/bluetoothdevice.h
```diff
@@ -103,24 +99,22 @@ class bluetoothdevice : public QObject {
 
     /**
      * @brief jouls Gets a metric object to get and set the number of joules expended. Units: joules
-     * @return
      */
     metric jouls();
 
     /**
      * @brief fanSpeed Gets the current fan speed. Units: revolutions per second
-     * @return
```
cagnulein: mmm the unit depends to the device. On domyos, for example, it's just a pure number without unit

src/bluetoothdevice.h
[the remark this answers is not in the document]
cagnulein: i'm using that name for all the types because a lot of users want to use a treadmill, for example, as a bike. So lastCrankEventTime is actually used to simulate a cadence as a bike from a treadmill. So i would like to leave this as it is

src/bluetoothdevice.h
```diff
@@ -36,64 +46,304 @@ class bluetoothdevice : public QObject {
...
     virtual void setAverageAzimuthNext300m(double azimuth) { azimuthAvgNext300m = azimuth; }
+
+    /**
+     * @brief lastCrankEventTime The time of the last crank event. Units: 1/1024s
+     */
     virtual uint16_t lastCrankEventTime();
+
+    /**
+     * @brief VirtualDevice ???
+     */
```
cagnulein, line 189: the virtual bridge to zwift for example, or to any 3rd party app.

src/bluetoothdevice.h
```diff
@@ -36,64 +46,304 @@ class bluetoothdevice : public QObject {
...
     virtual void clearStats();
+
+    /**
+     * @brief bluetoothDevice The bluetooth device information.
+     */
     QBluetoothDeviceInfo bluetoothDevice;
+
+    /**
+     * @brief disconnectBluetooth Disconnect from bluetooth (the device or the client software ??)
+     */
```
cagnulein, line 226: the device itself

src/bluetoothdevice.h
```diff
@@ -36,64 +46,304 @@ class bluetoothdevice : public QObject {
...
+    /**
+     * @brief setGPXFile Sets the file for GPS data exchange.
+     * @param filename The file path.
+     */
     void setGPXFile(QString filename);
+
+    /**
+     * @brief currentGPXBase64 ???
+     */
     QString currentGPXBase64() { return gpxBase64; }
```
cagnulein, line 306: returns the Base64 encode for the current GPX used.

cagnulein (commented):
checked

src/bluetoothdevice.h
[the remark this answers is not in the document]
drmason789: Acknowledged.

src/bluetoothdevice.h
drmason789: I've changed the wording to use inclination as vertical:horizontal instead of degrees, as well as being vague about limits.

cagnulein (approved):

drmason789:
I prefer to leave this as is. Further refinements could be made as better detail is determined.

cagnulein:
Thanks!

PR #883, opened 2022-08-06
Add support for the Trixter X-Dream V1 bike 

cagnulein:
#855

@drmason789


drmason789:
@cagnulein I wasn't expecting this to happen so fast. I thought you were going to "add a virtual bike".

I'm now testing with Linux and the real bike. Not successful yet.

cagnulein:
adding the virtual bike it's a thing of 2 minutes :) I just will be more
comfortable with a PR because I already have everything set up :)

Roberto Viola
http://robertoviola.cloud



Il giorno sab 6 ago 2022 alle ore 14:36 David Mason <
***@***.***> ha scritto:

Issue #887, opened 2022-08-08
[SUGGESTION] Improve use of polymorphism in bluetooth class


drmason789:
The bluetooth class has 10s of fields, each dedicated to a particular class that handles a particular device.
It also has a function device() which looks through all of these for the first one that isn't nullptr. 
device() is called frequently.

When it comes time to deallocate the object for the device, again the whole list of fields is traversed, deleting the non-null ones.

Additionally, there is a function with a huge if else if else if else ... construct.

I suggest 
- replace the vast majority of these device fields with a single "bluetoothdevice * bluetoothDevice;
- have the device() function simply return the bluetoothdevice* field
- where these fields were searched and deleted, just delete and nullify the new field
- divide the enormous if else if else construct into functions dedicated to each device. In the cases where the detection checks if another type of device has already been found, use a dynamic_cast on device().

I have already done this and have it building and running, but I can't test it as I don't have any of the devices. I'll commit to a branch so you have it to copy and paste if you want to use it.

cagnulein:
Yes sure, create a PR! thanks 
I never have time to improve the code :(

drmason789:
I don't seem to be able to target a new branch in your repo. Is it okay to target your repo's master branch for this PR?

drmason789:
I will make a new PR with only the first 3 items.

drmason789:
Reworded issue:

The bluetooth class has 10s of fields, each dedicated to a particular class that handles a particular device.
It also has a function device() which looks through all of these for the first one that isn't nullptr.
device() is called frequently.

When it comes time to deallocate the object for the device, again the whole list of fields is traversed, deleting the non-null ones.

I suggest
- replace the vast majority of these device fields with a single "bluetoothdevice * bluetoothDevice;
- have the device() function simply return the bluetoothdevice* field
- where these fields were searched and deleted, just delete and nullify the new field

drmason789:
I'm wondering why this:

```
 if (device() && device()->VirtualDevice()) {
        if (device()->deviceType() == bluetoothdevice::TREADMILL) {

            delete static_cast<virtualtreadmill *>(device()->VirtualDevice());
        } else if (device()->deviceType() == bluetoothdevice::BIKE) {
            delete static_cast<virtualbike *>(device()->VirtualDevice());
        } else if (device()->deviceType() == bluetoothdevice::ELLIPTICAL) {
            delete static_cast<virtualtreadmill *>(device()->VirtualDevice());
        }
[...]

PR #888, opened 2022-08-08
Bluetooth polymorphism #887

drmason789:
Some refactoring. Could have broken a few things, but I can't test as I have none of these devices.

Actually, I'd rather just put these changes into a branch so they're saved.


drmason789:
I seem to have a merge problem with some comment changes from another issue. 
Apparently I'm not a git expert yet.

drmason789:
#887

cagnulein:
I didn't check yet the code, but since a lot of code lines are involved in this one, I guess it will be useful to create some tests function to check that all the cases are ok.

https://doc.qt.io/qt-6/qtest-overview.html

What do you think?
I mean I have only 2 devices of the hundreds compatibile with qz, so my tests will be very limited

drmason789:
It might seem like a substantial amount of work, but it's not actually that much to me, and I realised before I started that it would be a risk for you. So I understand if you don't want to merge it in. 

Although I am a fan of unit testing, I don't want to put the time in on this project as it is currently because I'm really not a fan of this Qt/C++ environment. Everything seems to take so long compared with Visual Studio, C# and .NET.

You've probably seen from my videos that I've got an emulator for the bike I've been adding to QZ, and if you've looked through the repo for that, I've got numerous tests (including some Google Tests for the trixterxdreamv1client class I've put in QZ), some of which use data in binary files originally read from the actual device, so I can test without having the device connected.

I think the QZ code could be refactored to better facilitate unit testing. For example, using dependency injection for the "data source". Currently each "bluetoothdevice" object is taking care of reading from wherever the data comes from (a real bluetooth device, wifi, serial port etc), populating the metrics, and then something I've not discovered yet passes the metrics via bluetooth to the cycling application (Zwift, etc). 

For tests, there needs to be a means to bypass the device and feed data collected from the device, or suitable fake data, into the object being tested. 

In my repo for the Trixter bike, I have the "business logic" reading data from what is basically a data source interface, which feeds a supply of data objects that implement a specific interface. When using the real device, I plug in an implementation that reads from the serial port. For testing, I use a test implementation that either constructs data objects in order to tests specifically what the test is trying to test, or generated from a binary file stored from data read from the real bike.

For each bluetoothdevice subclass in this project, I think I would look at what is feeding it data (I've not looked at this yet), e.g. is there a signal emitted to indicate "new data packet from bluetooth" and simply provide a way to plug in a different object that emits the same signal, but supplies data read from a file.

I think you'd also need a way for users to store data and donate it to the project. Maybe they could opt to send usage data by internet to your server. :-)

cagnulein:
> It might seem like a substantial amount of work, but it's not actually that much to me, and I realised before I started that it would be a risk for you. So I understand if you don't want to merge it in. 

I would like to merge it, but only if I can test properly :)

> 
> Although I am a fan of unit testing, I don't want to put the time in on this project as it is currently because I'm really not a fan of this Qt/C++ environment. Everything seems to take so long compared with Visual Studio, C# and .NET.

It's so strange that you say so, because I think the same of .NET :D but I guess it's just how each person habits to work :) anyway I don't want to put the all test to you, I was just explaining my idea about this :)

> 
> You've probably seen from my videos that I've got an emulator for the bike I've been adding to QZ, and if you've looked through the repo for that, I've got numerous tests (including some Google Tests for the trixterxdreamv1client class I've put in QZ), some of which use data in binary files originally read from the actual device, so I can test without having the device connected.

Yes absolutely, I blindly trust of your work.

> I think the QZ code could be refactored to better facilitate unit testing. For example, using dependency injection for the "data source". Currently each "bluetoothdevice" object is taking care of reading from wherever the data comes from (a real bluetooth device, wifi, serial port etc), populating the metrics, and then something I've not discovered yet passes the metrics via bluetooth to the cycling application (Zwift, etc). 

Agree. I'm just writing the code in the fastest way because QZ is an hobby for me that I eating my life. So I really appreciate when someone is joining my cause on this. But of course I don't want to force you!


> For tests, there needs to be a means to bypass the device and feed data collected from the device and stored, or suitable fake data, into the object being tested. 

I have already all the debug log for each case, so it's just a problem of parsing it. But as I said, I don't have the time to do so. Or at least not now.


> In my repo for the Trixter bike, I have the "business logic" reading data from what is basically a data source interface, which feeds a supply of data objects that implement a specific interface. When using the real device, I plug in an implementation that reads from the serial port. For testing, I use a test implementation that either constructs data objects in order to tests specifically what the test is trying to test, or generated from a binary file stored from data read from the real bike.

Yes it's what I have also in mind now if I would start again QZ from scratch :)

> For each bluetoothdevice subclass in this project, I think I would look at what is feeding it data (I've not looked at this yet), e.g. is there a signal emitted to indicate "new data packet from bluetooth" and simply provide a way to plug in a different object that emits the same signal, but supplies data read from a file.

Yes my idea is generating the Bluetooth receive signal from file. It shouldn't be too hard, but the timing of the Bluetooth will play an important role. Probably also the timestamp in the test would be generated in the same way.


> I think you'd also need a way for users to store data and donate it to the project. Maybe they could opt to send usage data by internet to your server. :-)

I don't like this :) I mean the privacy for me it's important and I don't want to store users data

Anyway thanks for the collaboration so far!

drmason789:
I recommend cancelling this PR, there are smaller steps that could be made.

cagnulein:
you can update in the future instead of cancelling it?

drmason789:
I could still update this one, but I need to roll it all back and start over.

cagnulein:
i guess you cancel this by yourself if you want :) Let me know otherwise

Issue #895, opened 2022-08-12
[BUG] Resistance tile labelling indicates %, but it's not really


drmason789:
The resistance tile is labelled to indicate its numbers refer to a % resistance:

![image](https://user-images.githubusercontent.com/29954900/184444379-5cf34dbe-02e1-428a-a9d2-22e58442e221.png)

However, it isn't really.

The bluetoothdevice sublass I am testing this with overrides maxResistance() to return 250 (see the Resistance box on the right of the bottom window in the video), which is the device's maximum resistance. However, the bike::changeResistance(int8_t) slot is being sent values up to 127, the upper limit of the argument's data type of int8_t.

https://user-images.githubusercontent.com/29954900/184446323-83d4891c-2b06-4d56-9733-46dc08e184fd.mp4

Perhaps it could be made a rule that bike::changeResistance(int8_t) will be sent a percentage from 0 to 100, and it should adjust according to whatever maxResistance() returns for the class.

Or perhaps the tile could calculate an increment, scale = 0.01 * bike::maxResistance() and send its (the tile's) current value * scale to bike::changeResistance(int8_t).

cagnulein:
you're right, this is not a %. It's a raw value from each bike. So I'm just removing the %

drmason789:
I see you've removed the % sign. 

There is still the problem if the device supports more than 127 resistance levels, the int8_t parameter on the changeResistance slot is too small.

cagnulein:
hah sorry I didn't read it carefully. So we need to change the int8 to int16 everywhere

drmason789:
I could do this if you prefer.

cagnulein:
hah ok thanks! In these days I have my wife at the hospital so I'm quite busy

drmason789:
Perhaps I should 

`#define resistance_t int16_t`

and use that. There are many locations to change.

cagnulein:
Yes sounds good. Also all the slots and signals must be changed...

drmason789:
I've got this done. 

Perhaps another issue with power_t, pelotonResistance_t, heartRate_t, ...

drmason789:
I've done the testing I can by merging into a branch of my branch for #855 - it seems to work. 

https://user-images.githubusercontent.com/29954900/184517624-60e3088a-2757-451b-84e6-74a34829447a.mp4

drmason789:
@cagnulein Do you prefer to be the one that closes issues?

[...]

PR #898, opened 2022-08-14
#895 introduced "typedef int16_t resistance_t;" and used throughout code

drmason789:


drmason789:
#895

src/bluetooth.cpp
```diff
@@ -1746,12 +1746,12 @@ void bluetooth::connectedAndDiscovered() {
 
                 connect(ftmsAccessory, &smartspin2k::debug, this, &bluetooth::debug);
 
-                connect(this->device(), SIGNAL(resistanceChanged(int8_t)), ftmsAccessory,
-                        SLOT(changeResistance(int8_t)));
-                connect(this->device(), SIGNAL(resistanceRead(int8_t)), ftmsAccessory,
-                        SLOT(resistanceReadFromTheBike(int8_t)));
-                connect(ftmsAccessory, SIGNAL(resistanceRead(int8_t)), this->device(),
-                        SLOT(resistanceFromFTMSAccessory(int8_t)));
+                connect(this->device(), SIGNAL(resistanceChanged(int16_t)), ftmsAccessory,
```
cagnulein, line 1749: here we need to use Q_DECLARE_METATYPE and qRegisterMetaType https://stackoverflow.com/questions/19397567/passing-structure-to-signal-in-qt

cagnulein:
@drmason789 https://github.com/cagnulein/qdomyos-zwift/pull/898/files#r948206063

src/bluetooth.cpp
drmason789: Okay, but resistance_t, which I have missed the replacement to here, is just a #define to one of the standard int types, not a struct, which is that that documentation refers to. Aren't these SIGNAL and SLOT macros obsolete? Are the connected slots not working for this code?

drmason789: Perhaps you were thinking ahead, perhaps thinking of actually having a resistance struct, which encapsulates a resistance level, and a resistance type (e.g. peloton, percentage, device), with converter functions like toPeloton(), toPercentage(), toDevice().

cagnulein: > Okay, but resistance_t, which I have missed the replacement to here, is just a #define to one of the standard int types, not a struct, which is that that documentation refers to. 

but we can simply transform it to a typedef and it should work.
i can test it myself but not sooner than the next week


> Aren't these SIGNAL and SLOT macros obsolete? 

nope. at least i'm not aware of this 

> Are the connected slots not working for this code?

this code is working for sure, it's just a matter of cleaning the code :)

cagnulein: > Perhaps you were thinking ahead, perhaps thinking of actually having a resistance struct, which encapsulates a resistance level, and a resistance type (e.g. peloton, percentage, device), with converter functions like toPeloton(), toPercentage(), toDevice().

nope, i mean it makes sense but i was thinking simply about the signal slot thing ;)

drmason789:
@cagnulein Are you still wanting Q_DECLARE_METATYPE and qRegisterMetaType now that I've switched to a typedef?

cagnulein:
@drmason789 i'm waiting the possibility to test it :) my wife is at the hospital and i don't have time to test this yet

cagnulein:
@drmason789 tested. Seems fine to me. I just added some assert to the connects just to add a runtime check. Thanks!

Issue #921, opened 2022-09-02
[SUGGESTION] Simple code reduction in bluetooth.cpp


drmason789:
In my issue #887, I remove most of the device fields, and change each of these device detecting if statements to store the new device object as a bluetoothdevice* and do the actions that are performed at the end of each block (notably if the discoveryAgent isn't active, emit searchingStop, and adding the new device to the template managers) in once place at the end of the for loop.

```
} else if (!nordictrack_2950_ip.isEmpty() && !nordictrackifitadbTreadmill) {
                discoveryAgent->stop();
                nordictrackifitadbTreadmill = new nordictrackifitadbtreadmill(noWriteResistance, noHeartService);
                emit deviceConnected(b);
                connect(nordictrackifitadbTreadmill, &bluetoothdevice::connectedAndDiscovered, this,
                        &bluetooth::connectedAndDiscovered);
                connect(nordictrackifitadbTreadmill, &nordictrackifitadbtreadmill::debug, this, &bluetooth::debug);
                // nordictrackifitadbTreadmill->deviceDiscovered(b);
                // connect(this, SIGNAL(searchingStop()), cscBike, SLOT(searchingStop())); //NOTE: Commented due to #358
                if (!discoveryAgent->isActive()) {
                    emit searchingStop();
                }
                userTemplateManager->start(nordictrackifitadbTreadmill);
                innerTemplateManager->start(nordictrackifitadbTreadmill);
            }
```

As you are understandably unkeen on merging a PR of that magnitude, would you consider a change in bluetooth.cpp like this:

Before the large if statement:

`bluetoothdevice * newDevice = nullptr;`

then for each if block, e.g. 

```
} else if (!nordictrack_2950_ip.isEmpty() && !nordictrackifitadbTreadmill) {
                discoveryAgent->stop();
                newdevice = this->nordictrackifitadbTreadmill = new nordictrackifitadbtreadmill(noWriteResistance, noHeartService);
                emit deviceConnected(b);
                connect(nordictrackifitadbTreadmill, &bluetoothdevice::connectedAndDiscovered, this,
                        &bluetooth::connectedAndDiscovered);
                connect(nordictrackifitadbTreadmill, &nordictrackifitadbtreadmill::debug, this, &bluetooth::debug);
                // nordictrackifitadbTreadmill->deviceDiscovered(b);
                // connect(this, SIGNAL(searchingStop()), cscBike, SLOT(searchingStop())); //NOTE: Commented due to #358
[...]

Issue #911, opened 2022-08-23
[REQ] Automated testing


drmason789:
@cagnulein  You've suggested this before [here](https://github.com/cagnulein/qdomyos-zwift/pull/888#issuecomment-1209305409)

Suggestions for making a start:
- new folder alongside src called tst
- project file tst/qdomyos-zwift-tests.pro
- folders for the tests for each device

drmason789:
I can make a start by importing my Google tests for the trixterxdreamv1client class.

cagnulein:
ok thanks, I was thinking about QT tests because so it will be all integrated on the IDE but I never did a test class so I'm a newbie on this.

drmason789:
I would convert them to Qt tests. I just want to check you're okay with the file system organisation before doing anything.

cagnulein:
great! thanks! yes the fs proposed is perfect to me

cagnulein:
A first level of testing could be injecting each Bluetooth name in the discovery functions checking if a device will be created 

Every fitness machine module must be its Bluetooth name lists to test

drmason789:
[Connecting the projects](https://dragly.org/2014/03/13/new-project-structure-for-projects-in-qt-creator-with-unit-tests/)

drmason789:
It seems that with QtTest you can only have 1 test class per project, otherwise QtCreator won't recognise the test class.

Consequently I've copied the structure of the Qt source code and used a tree of projects using the SUBDIRS template, with the tests at the leaf nodes. 

The project that includes the main application and the test projects is in the root directory, qdomyos-zwift.pro.

No classes from the main project instantiated in the tests yet.

drmason789:
I believe I'm going to need to change something so that the current qdomyos-zwift executable produced from the code in src must be a lib output, for the test project to link its outputs to.

I have a couple of ideas to explore:
1. Define much of the src/qdomyos-zwift.pro in a pri file that is shared between two replacement pro files that include this and differ only in that one specifies TEMPLATE = app and the other TEMPLATE = lib 
2. Split the current /src code into an app for the UI and a lib project for the devices and other "business". Then have the tests link to the lib project.

drmason789:
I've implemented approach 1 and have a test case that creates an activiotreadmill object working.

Next up, testing the bluetooth name recogniser.

drmason789:
[...]

PR #922, opened 2022-09-02
#921 consolidation of repeated code in bluetooth.cpp

drmason789:
#921


src/bluetooth.cpp
```diff
@@ -1708,11 +1523,13 @@ void bluetooth::deviceDiscovered(const QBluetoothDeviceInfo &device) {
                 chronoBike->deviceDiscovered(b);
                 // NOTE: Commented due to #358
                 // connect(this, SIGNAL(searchingStop()), chronoBike, SLOT(searchingStop()));
-                if (!discoveryAgent->isActive()) {
+            }
+
+            if(newDevice) {
+                if (emitSearchingStop && !discoveryAgent->isActive())
```
drmason789, line 1529: In order to avoid changing behaviour I have this emitSearchingStop variable. But I wonder, am I simply refactoring a bug? Should it always emit searchingStop if the discoveryAgent isn't active?

cagnulein: that signal is an inheritance of the past and i think it's quite useless nowadays. Anyway, in order to don't break things, your solution seems fine... We are always in time to remove it in the future ;)

src/bluetooth.cpp
```diff
@@ -1458,8 +1298,7 @@ void bluetooth::deviceDiscovered(const QBluetoothDeviceInfo &device) {
                 // connect(echelonConnectSport, SIGNAL(inclinationChanged(double)), this,
                 // SLOT(inclinationChanged(double)));
                 flywheelBike->deviceDiscovered(b);
-                userTemplateManager->start(flywheelBike);
-                innerTemplateManager->start(flywheelBike);
+                emitSearchingStop = false;
             } else if ((b.name().toUpper().startsWith(QStringLiteral("MCF-"))) && !mcfBike && filter) {
                 this->stopDiscovery();
                 mcfBike = new mcfbike(noWriteResistance, noHeartService, bikeResistanceOffset, bikeResistanceGain);
```
drmason789, line 1304: Missing assignment of newDevice

drmason789:
Transferring from:

https://github.com/cagnulein/qdomyos-zwift/pull/950#issuecomment-1264459600

https://github.com/cagnulein/qdomyos-zwift/pull/950#issuecomment-1264552934

cagnulein:
thanks bot

drmason789:
This will be rendered obsolete by the eventual PR for #911.

cagnulein:
hah ok, so let's leave this in hold

cagnulein:
ping

drmason789:
Closing as will absorb into other work.

drmason789:
Closing as will absorb into other device detection work. What this changes would be tested by tests added in #1221.

Issue #927, opened 2022-09-06
[REQ] bluetooth::startDiscovery and stopDiscovery


drmason789:
While aligning the starting and stopping of discovery of non-bluetooth devices with the that of the bluetooth device discovery in another issue #855, and I saw that the bluetooth discovery is started in different ways on different platforms, but this is not consistent across all "starts".

For example,  bluetooth::bluetooth and bluetooth::finished have:

```
    if (!trx_route_key && !bh_spada_2 && !technogym_myrun_treadmill_experimental) {
#endif
        discoveryAgent->start(QBluetoothDeviceDiscoveryAgent::LowEnergyMethod);
#ifndef Q_OS_IOS
    } else {
        discoveryAgent->start(QBluetoothDeviceDiscoveryAgent::ClassicMethod |
                              QBluetoothDeviceDiscoveryAgent::LowEnergyMethod);
    }
#endif
```

but bluetooth::restart has:

```
    if (onlyDiscover) {
        onlyDiscover = false;
        discoveryAgent->start();
        return;
    }
```

and at the end

`discoveryAgent->start();`

_Originally posted by @drmason789 in https://github.com/cagnulein/qdomyos-zwift/issues/889#issuecomment-1214200327_

Apparently these should be the same. https://github.com/cagnulein/qdomyos-zwift/issues/889#issuecomment-1214211876_

drmason789:
I would have done this for #889 but that could be a while away.
- put the full code from bluetooth::bluetooth and bluetooth::finished into a new method bluetooth::startDiscovery() and call it from all the places that currently call discoveryAgent->start().
- add a bluetooth::stopDiscovery() which calls discoveryAgent->stop(). Replace all discoveryAgent->stop() with a call to that. At a later time, other actions (e.g. stopping a non-bluetooth device discovery process) can be added here in a single place.

@cagnulein Do you approve?

drmason789:
I've actually already done this  for #855, but that could take a while.

cagnulein:
yes absolutely, it's actually a bug. Do a dedicated PR for this one and i will merge it asap ;)

PR #930, opened 2022-09-07
#927 route starting and stopping of bluetooth discovery through new functions startDiscovery() and stopDiscovery()

drmason789:
#927

### DIFF
--- a/src/bluetooth.cpp
+++ b/src/bluetooth.cpp
@@ -15,11 +15,6 @@ bluetooth::bluetooth(bool logs, const QString &deviceName, bool noWriteResistanc
                      uint32_t pollDeviceTime, bool noConsole, bool testResistance, uint8_t bikeResistanceOffset,
                      double bikeResistanceGain) {
     QSettings settings;
-    bool trx_route_key = settings.value(QStringLiteral("trx_route_key"), false).toBool();
-    bool bh_spada_2 = settings.value(QStringLiteral("bh_spada_2"), false).toBool();
-    bool technogym_myrun_treadmill_experimental =
-        settings.value(QStringLiteral("technogym_myrun_treadmill_experimental"), false).toBool();
-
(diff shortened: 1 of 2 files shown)


src/bluetooth.cpp
```diff
@@ -172,16 +155,34 @@ void bluetooth::finished() {
...
+                                  QBluetoothDeviceDiscoveryAgent::LowEnergyMethod);
+        }
+    #endif
+}
+
+
+void bluetooth::stopDiscovery()
+{
+    if(this->discoveryAgent)
+        this->discoveryAgent->stop();
```
drmason789, line 185: This is changed behaviour...previous code was not doing the nullptr check.

cagnulein: let's put a qdebug if the discoveryagent is a null pointer in order to track future bug, thanks

src/bluetooth.cpp
```diff
@@ -172,16 +155,34 @@ void bluetooth::finished() {
...
+}
+
+void bluetooth::startDiscovery() {
+
+    #ifndef Q_OS_IOS
+        QSettings settings;
+        bool technogym_myrun_treadmill_experimental =
+            settings.value(QStringLiteral("technogym_myrun_treadmill_experimental"), false).toBool();
+        bool trx_route_key = settings.value(QStringLiteral("trx_route_key"), false).toBool();
+        bool bh_spada_2 = settings.value(QStringLiteral("bh_spada_2"), false).toBool();
```
drmason789, line 168: I moved these bools into the #ifndef block because they're not used outside of it.

cagnulein: sounds good

src/bluetooth.cpp
drmason789: Would you prefer one that always happens, or all call to debug(..) which checks the log field ?

cagnulein: In the else case as you did, thanks!

PR #940, opened 2022-09-13
New Apple Watch 9 Metrics for treadmill (for now are avaiable only for outdoor workotus)

cagnulein:
runningStrideLength
A quantity sample type that measures the distance covered by a single step while running. iOS 16.0+
iPadOS 16.0+
macOS 13.0+ Beta
Mac Catalyst 16.0+ Beta
watchOS 9.0+
Declaration
static let runningStrideLength: HKQuantityTypeIdentifier Discussion
These samples use length units (described in HKUnit) and measure discrete values (described in HKQuantityAggregationStyle). During outdoor running workouts, the system automatically records running stride samples on Apple Watch SE and Series 6 and later.

https://developer.apple.com/documentation/healthkit/hkquantitytypeidentifier/3929732-runningstridelength/


cagnulein:
https://developer.apple.com/forums/thread/714484#714484021

Issue #948, opened 2022-09-20
[SUGGESTION] QSettings keys as QStringLiteral constants rather than repeated hardcoded string literals


drmason789:
**Is your feature request related to a problem? Please describe.**
The QSettings object  is accessed via hard-coded string literals. The programmer needs to know what these are and what they mean.  It would be convenient if they were available as constants somewhere with comments describing what they are available in QtCreator's intellisense.

**Describe the solution you'd like**
A new class, e.g. QZSettings, which contains QStringLiteral members defining the general QZ settings keys.
Default values also defined as constants.
Device-specific settings keys also in the new class OR in the device's bluetoothdevice subclass
Code documentation on each i.e. /** @brief etc.
Replace calls into QSettings that use hardcoded string literals with references to these constants.

Even better: static strongly-typed accessors where the getter applies the default.

**Describe alternatives you've considered**
Replace that whole QSettings class with one that provides an integer index for your settings key, and replaces the text lookup with an O(1) array lookup so we can trim some milliseconds. Also notifies subscribers when a change occurs so clients can cache the values instead of looking them up repeatedly.

**Who's going to do it?**
I'll do it. The first one...not the alternative.

drmason789:
@cagnulein By "I'll do it" I mean I'll do the change if you agree to it.

cagnulein:
Yes thanks, it's something that I have to do but how could this works into the qml? Also I was scary about doing some error and causing problem to the users

drmason789:
I'd leave the QML alone. The access to the settings in the C++ would still go through the QSettings object so I believe there would be not change of function in the QML.

cagnulein:
Yeah I was saying so because if we put the constants for all the settings, it will be nice to have a constant for whole code, otherwise we will have the string for each settings in the c++ new module and in the QML.
But ok, let's do the c++ first, in case I will do the QML then

drmason789:
I've used a namespace instead of a class to contain the constants. I was thinking in C# when I suggested the class. Could still do that though. An instance attached to a QSettings instance providing strongly typed accessor functions that transparently handle the default values would shrink the calling code.

PR #950, opened 2022-09-21
Use of named constants for settings keys and default values

drmason789:
#948 

@cagnulein I'm yet to make a start on the inline documentation. 

I've done this change in simple steps so you can review each change incrementally. To make the qzsettings.h file, I started with the settings.qml file and regex-replaced it to much of what it is now, incrementally so the default values can be observed in the review.

I found a case that appeared to be a wrong key name used: tile_nextrowtrainprogram_order in the C++ where there was tile_nextrowstrainprogram_order (rows plural) in the settings.qml file.

I've left the old hardcoded defaults in the settings.value calls beside the replacement reference to a constant as I will go over each of these and see if they are different from the ones defined in settings.qml that have been transferred to qzsettings.h.
I wonder if we could do something smart here, e.g. set the defaults from a QSettings object that's been directed away from the stored values of the various profiles.


drmason789:
"Compile Linux Desktop" action took 2mins longer than previously.

cagnulein:
> I found a case that appeared to be a wrong key name used: tile_nextrowtrainprogram_order in the C++ where there was tile_nextrowstrainprogram_order (rows plural) in the settings.qml file.

OPS!

> I've left the old hardcoded defaults in the settings.value calls beside the replacement reference to a constant as I will go over each of these and see if they are different from the ones defined in settings.qml that have been transferred to qzsettings.h.

you did right, it will be useful at least at the beginning.
I was thinking an automatic way to check if all your work is right and I came with a simple idea.

Using the same for code that we have here https://github.com/cagnulein/qdomyos-zwift/blob/94830044f6344448c637dc1520b5257c35601377/src/main.cpp#L347 we can print all the setting in the actual master, and then we can do the same on this PR. The result must be the same if all the work is ok. What do you think?

> I wonder if we could do something smart here, e.g. set the defaults from a QSettings object that's been directed away from the stored values of the various profiles.

mmm I think your actual implementation is smart enough :)

cagnulein:
another idea that I had some time ago is adding to the settings some attribute, for example a description in order to show it in the settings UI, or a measure unit.... But it's something that we can do in a next step

drmason789:
I've been using tooltips for the settings in #855 / #899 but these are hardcoded into the UI QML.

cagnulein:
@drmason789 hah I didn't noticed! The problem about the tooltip is that it works only with mouse. It would be better to add a question mark icon button or something similar

drmason789:
> we can print all the setting in the actual master, and then we can do the same on this PR. The result must be the same if all the work is ok. What do you think?

The code you referenced is looping over all the keys that exist in the QSettings and reading their values. 

What I'm looking to test is  for mismatches between the default specified in calls to QSettings.value(key, default) and the default specified in the settings.qml file. I've attempted to avoid creating differences between settings.qml and qzsettings.h by starting qzsettings.h from settings.qml and incrementally converting the QML property definitions to C++ using regular expression search and replace. I've committed each of these steps separately for side-by-side comparison in review. I encountered 1 problem where the QML has a bool property with value "value" instead of "true" or "false". 

To test that the defaults constants match the settings, I believe I would need to create a QSettings object that has no connection to an existing profile, read out all the values, and compare with the constants. If we do that, I think, why not just read the defaults out of the object and store them instead of checking the hardcoded ones are correct? That way the defaults, for the keys that are actually in the settings.qml file, are only defined in 1 place (the settings.qml file).

cagnulein:
> The code you referenced is looping over all the keys that exist in the QSettings and reading their values.

After a run, the qsettings are getting the default values, so it will be a really easy job to do on both master and this, isn't it?

drmason789:
I think to achieve what you are looking to do, I would need to have  line for each key constant to read the value from the QSettings object and debug log it. This is simple, I can do that.

cagnulein:
> What I'm looking to test is for mismatches between the default specified in calls to QSettings.value(key, default) and the default specified in the settings.qml file. I've attempted to avoid creating differences between settings.qml and qzsettings.h by starting qzsettings.h from settings.qml and incrementally converting the QML property definitions to C++ using regular expression search and replace. I've committed each of these steps separately for side-by-side comparison in review. I encountered 1 problem where the QML has a bool property with value "value" instead of "true" or "false".

Hah ok, this I guess this it's impossible to do in automatic way.

cagnulein:
> If we do that, I think, why not just read the defaults out of the object and store them instead of checking the hardcoded ones are correct? That way the defaults, for the keys that are actually in the settings.qml file, are only defined in 1 place (the settings.qml file).

Because you can do it after the qml loaded, because it's only when the settings.qml is loaded the first time that the setting file will be created.

drmason789:
@cagnulein Do you have an idea for where you'd like the check done? Once #911 is done it could go in a unit test but that's not here yet. In main.cpp near the code you referred to, perhaps?

drmason789:
Actually, it's not really a check, it's just a log. There could be more settings in the QSettings object than are known about in the QZSettings namespace.

src/homeform.cpp
```diff
@@ -3921,8 +3921,8 @@ void homeform::networkRequestFinished(QNetworkReply *reply) {
 
 void homeform::callbackReceived(const QVariantMap &values) {
     qDebug() << QStringLiteral("homeform::callbackReceived") << values;
-    if (!values.value(QStringLiteral("code")).toString().isEmpty()) {
-        strava_code = values.value(QStringLiteral("code")).toString();
+    if (!values.value(QZSettings::code).toString().isEmpty()) {
+        strava_code = values.value(QZSettings::code).toString();
```
drmason789, line 3925: @cagnulein Do you think this setting should be changed to strava_code ?

cagnulein: We can't change the name of the settings otherwise we will break everything

drmason789:
I will do a sweep of the settings keys. Those I think have self-explanatory names, I will remove the inline documentation (currently empty). For others I will provide some details.

drmason789:
@cagnulein 

I'm about to sweep the settings key constant usages as follows:
- search for references to the constants using QtCreator
- check the calls to settings.value, and check that the old specified default (in comment) matches the corresponding default constant
- check that the call isn't using some other value eg from a variable obtained from the command line

https://user-images.githubusercontent.com/29954900/192628463-42f0ea49-ea19-4e01-a5ad-aab62e3df398.mp4

drmason789:
@cagnulein I'm finding a few places where QLatin1String is used instead of QStringLiteral.

For the default for "user_email" homeform.cpp uses "" in some places and QStringLiteral in others.

Based on this [discussion](https://forum.qt.io/topic/78540/qstringliteral-vs-qlatin1string) I'll use QStringLiteral throughout for consistency and quick conversion to QString.

drmason789:
@cagnulein There's a conflict on default_continuous_moving. 

settings.qml : false
bluetoothdevice, elliptical and treadmill : true
trainprogram : false

Current code in this PR has the default of false for all of them by way of the constant.

Do we have a precise definition of what this setting represents?

drmason789:
https://github.com/cagnulein/qdomyos-zwift/blob/43d6e6c233886ec53ae1e93f4476efff037659e8/src/powerzonepack.cpp#L11-L17

Here I am going to assume this is a comparison with the default value and use the default value constant in both places.

drmason789:
For this one, the default password in settings.qml is "username"

https://github.com/cagnulein/qdomyos-zwift/blob/43d6e6c233886ec53ae1e93f4476efff037659e8/src/settings.qml#L65-L66

but when used, it's "password":

https://github.com/cagnulein/qdomyos-zwift/blob/43d6e6c233886ec53ae1e93f4476efff037659e8/src/powerzonepack.cpp#L28-L32

drmason789:
homeform::sortTiles has some conflicting defaults for tile order. Not sure what to do here other than use the hardcoded values instead of the constants. I'm wondering why some are 0, and some are numbers like 24,28,29, and 30.

drmason789:
cadence tile is disabled by default for treadmills, enabled for others.

cagnulein:
> @cagnulein I'm finding a few places where QLatin1String is used instead of QStringLiteral.
> 
> For the default for "user_email" homeform.cpp uses "" in some places and QStringLiteral in others.
> 
> Based on this [discussion](https://forum.qt.io/topic/78540/qstringliteral-vs-qlatin1string) I'll use QStringLiteral throughout for consistency and quick conversion to QString.

I'm worried about this because the email can contain only Latin1 characters, why don't leave it as it is?

> @cagnulein There's a conflict on default_continuous_moving.
> 
> settings.qml : false bluetoothdevice, elliptical and treadmill : true trainprogram : false

good call, this setting should be off by default.

> Some kind of "ignore that the user has stopped doing work" flag?

yes exactly

> Here I am going to assume this is a comparison with the default value and use the default value constant in both places.

yes right!


> For this one, the default password in settings.qml is "username"
> 
> https://github.com/cagnulein/qdomyos-zwift/blob/43d6e6c233886ec53ae1e93f4476efff037659e8/src/settings.qml#L65-L66
> 
> but when used, it's "password":

yeah in this case there is no problem at all, because the user must change the password. you can use the "password" default value (so it will match with the settings.qml)

> homeform::sortTiles has some conflicting defaults for tile order. Not sure what to do here other than use the hardcoded values instead of the constants. I'm wondering why some are 0, and some are numbers like 24,28,29, and 30.

yeah that numbers are tricky to understand :D it's just the default order of the tiles. Unfortunately, we can't change them because most of the users are using that values, and changing them means receiving hundreads of mail about sorting the tiles. I know that for you and me sounds a stupid little things, but I assure you that this is not for the majority of the QZ users :)

So can we leave them as it is using maybe an override of the constructor?

> cadence tile is disabled by default for treadmills, enabled for others.

same as the previous point

drmason789:
>yeah in this case there is no problem at all, because the user must change the password. you can use the "password" default value (so it will match with the settings.qml)

You say to use the "password" default value so that it will match with settings.qml, but settings.qml has "username" as the default password. Perhaps I should change settings.qml to have "password" as the default value.

drmason789:
>settings.qml : false bluetoothdevice, elliptical and treadmill : true trainprogram : false

>good call, this setting should be off by default.

Okay, so I'll change this to default to off in the 3 for which it defaults to on. But might this change behaviour for users? I'm not currently sure if previous runs of QZ would write "false" to the config. Something I will need to check.

drmason789:
> yeah that numbers are tricky to understand :D it's just the default order of the tiles. Unfortunately, we can't change them because most of the users are using that values, and changing them means receiving hundreads of mail about sorting the tiles. I know that for you and me sounds a stupid little things, but I assure you that this is not for the majority of the QZ users :)

I'll leave the tiles alone. However,  as an outside observer, these seem like magic numbers shoved in. How were they calculated and what effect will inserting other tiles have ? As I said, I'll just leave them alone.

cagnulein:
> You say to use the "password" default value so that it will match with settings.qml, but settings.qml has "username" as the default password. Perhaps I should change settings.qml to have "password" as the default value.

Hah sorry, so the opposite. Don't change the setting.qml because changing it I saw that could leads to file corruption on iOS

cagnulein:
> Okay, so I'll change this to default to off in the 3 for which it defaults to on. But might this change behaviour for users? I'm not currently sure if previous runs of QZ would write "false" to the config. Something I will need to check.

You're right also on this one...maybe we need to keep the default settings as they are for this setting

cagnulein:
> I'll leave the tiles alone. However,  as an outside observer, these seem like magic numbers shoved in. How were they calculated and what effect will inserting other tiles have ? As I said, I'll just leave them alone.

They are just index number. Starting from 0 (the top left tile) to the max number of the tiles available.

So there isn't a real algorithm behind this, it's just that if we touch these, we have to fight with thousand of angry users :D

I don't if you want to put some constants instead but I don't see any real surrogate of them

drmason789:
Summary at this point:
Tiles: I will leave them with the existing numbers
pzp_password: I will switch the default to the constant "username" to match the settings.qml
continuous_moving: in the implementations where the default is true instead of settings.qml's false, I will leave the hardcoded values.
email items: I will try to prove that QStringLiteral is safe, but may end up reverting them. Things to determine: what actually happens to the value "" as it goes through various functions on its way to the email? What do the email functions take in, and do they do the conversion themselves?

cagnulein:
> Summary at this point:
> 
> Tiles: I will leave them with the existing numbers
> 
> pzp_password: I will switch the default to the constant "username" to match the settings.qml
> 
> continuous_moving: in the implementations where the default is true instead of settings.qml's false, I will leave the hardcoded values.
> 
> email items: I will try to prove that QStringLiteral is safe, but may end up reverting them. 

Ok to everything 

> Things to determine: what actually happens to the value "" as it goes through various functions on its way to the email? What do the email functions take in, and do they do the conversion themselves?

What do you mean? The email setting is used has the destination to the workout summary. So if it "" no email will be sent.

drmason789:
>Things to determine: what actually happens to the value "" as it goes through various functions on its way to the email? What do the email functions take in, and do they do the conversion themselves?

>What do you mean? The email setting is used has the destination to the workout summary. So if it "" no email will be sent.

I'm wondering what the problem is with using QStringLiteral("") instead of QLatin1String("") or whatever results from "" as the default for the user_email setting. You say you are "worried about this because the email can contain only Latin1 characters" but both MIME and the other form of email text go through  smtpClient.sendMessage which converts to utf8 so I think unicode shouldn't matter, unless the recipient needs the content to be in Latin1 characters. 

https://github.com/cagnulein/SmtpClient-for-Qt/blob/3fa4a0fe5797070339422cf18b5e9ed8dcb91f9c/src/smtpclient.cpp#L468-L476

Also, the email address, even if the "" is actually used which I believe it currently isn't (because, as you say, in that case no email is sent), if combined with other strings that are not specifically QLatin1String (unless the char* strings concocted on the fly are automatically converted to QLatin1String).

So I think it's most likely safe to use a precompiled QStringLiteral("").

cagnulein:
@drmason789 Yes probably is safe at 99% :) it's that 1% that make me crazy :)

We have to do test cases for everything we changed :) but ok let's go on this, but please run some test on this.

Thanks

drmason789:
@cagnulein The default value for ios_peloton_workaround is quite inconsistent...is the default dependant on the device?

cagnulein:
> @cagnulein The default value for ios_peloton_workaround is quite inconsistent...is the default dependant on the device?
> 
> 

It has a meaning only on iOS, and it must be TRUE

drmason789:
@cagnulein I notice that in the large, device recognising if statement in bluetooth.cpp, when a relatively small number of devices are discovered, the bluetooth name is written to "bluetooth_lastdevice_name" and in some environments, the address to "bluetooth_lastdevice_address".

Should this in fact be done for all devices? If so I think this could be added to #921.

cagnulein:
Yes that cuple contains the last name and address of the last connected device (only connected not discovered).
Actually that one is used only for the Schwinn bike iOS workaround if I remember correctly

drmason789:
@cagnulein I notice that virtualrower_forceresistance is only used once, and it is read into a variable that is not used. 

Would you agree to me removing this setting and the unused bool in virtualrower.cpp ?

cagnulein:
> @cagnulein I notice that virtualrower_forceresistance is only used once, and it is read into a variable that is not used. 
> 
> 
> 
> Would you agree to me removing this setting and the unused bool in virtualrower.cpp ?
> 
> 
> 
> 
> 
> 

We can't remove it from the settings.qml because of the corruption of the settings file but if you want you can omit it in the new class file, yes.

drmason789:
>>Yes that cuple contains the last name and address of the last connected device (only connected not discovered).
> Actually that one is used only for the Schwinn bike iOS workaround if I remember correctly

It's done for these:

domyos rower
Kingsmith R2 treadmill
Kingsmith R1 Pro treadmill
Shua A5 treadmill
True Treadmill
Schwinn IC4 Bike

but I don't see it actually being used anywhere. 

I'm wondering if this:

```
                settings.setValue(QZSettings::bluetooth_lastdevice_name, b.name());
#ifndef Q_OS_IOS
                settings.setValue(QZSettings::bluetooth_lastdevice_address, b.address().toString());
#else
                settings.setValue(QZSettings::bluetooth_lastdevice_address, b.deviceUuid().toString());
#endif
```

should be done for all the devices in that large if statement, that are actually bluetooth devices. 
With your confirmation, I could put this into the #921 changes where the processing that applies to many devices is standardised in 1 place.

cagnulein:
I'm from the phone right now, I would like to check better the code before answering. Thanks

drmason789:
I see now, the setting is set from those devices, but it's only actually used for the Schwinn bike, in code that's only in iOS builds.

drmason789:
https://github.com/cagnulein/qdomyos-zwift/blob/9498aa5501af47df1599e6afc231719354a277a8/src/settings.qml#L358

but

https://github.com/cagnulein/qdomyos-zwift/blob/9498aa5501af47df1599e6afc231719354a277a8/src/virtualbike.cpp#L30

https://github.com/cagnulein/qdomyos-zwift/blob/9498aa5501af47df1599e6afc231719354a277a8/src/virtualtreadmill.cpp#L15

drmason789:
https://github.com/cagnulein/qdomyos-zwift/blob/9498aa5501af47df1599e6afc231719354a277a8/src/settings.qml#L359
but
https://github.com/cagnulein/qdomyos-zwift/blob/9498aa5501af47df1599e6afc231719354a277a8/src/dirconmanager.cpp#L149

drmason789:
https://github.com/cagnulein/qdomyos-zwift/blob/9498aa5501af47df1599e6afc231719354a277a8/src/settings.qml#L397
but
https://github.com/cagnulein/qdomyos-zwift/blob/9498aa5501af47df1599e6afc231719354a277a8/src/homeform.cpp#L3190

https://github.com/cagnulein/qdomyos-zwift/blob/9498aa5501af47df1599e6afc231719354a277a8/src/settings.qml#L410
but
https://github.com/cagnulein/qdomyos-zwift/blob/9498aa5501af47df1599e6afc231719354a277a8/src/homeform.cpp#L3239

Also these commented out ones that have 0 instead of 10:

https://github.com/cagnulein/qdomyos-zwift/blob/9498aa5501af47df1599e6afc231719354a277a8/src/settings.qml#L24-L25
but
https://github.com/cagnulein/qdomyos-zwift/blob/9498aa5501af47df1599e6afc231719354a277a8/src/virtualbike.cpp#L461-L465
https://github.com/cagnulein/qdomyos-zwift/blob/9498aa5501af47df1599e6afc231719354a277a8/src/virtualrower.cpp#L188-L192

cagnulein:
> https://github.com/cagnulein/qdomyos-zwift/blob/9498aa5501af47df1599e6afc231719354a277a8/src/settings.qml#L358
> 
> but
> 
> https://github.com/cagnulein/qdomyos-zwift/blob/9498aa5501af47df1599e6afc231719354a277a8/src/virtualbike.cpp#L30
> 
> https://github.com/cagnulein/qdomyos-zwift/blob/9498aa5501af47df1599e6afc231719354a277a8/src/virtualtreadmill.cpp#L15

dircon MUST be true! Thanks!

cagnulein:
> 36866

dircon port MUST be 36866

cagnulein:
both tts value could stay false, thanks
the zwift_erg_filter* are used in the bike.cpp

cagnulein:
about the bluetooth_lastdevice_* yes it has to be in ALL the devices, confirmed.

drmason789:
> both tts value could stay false, thanks 

"stay false" means they are already false. They are true, it's the settings.qml value that is false.
I will take your statement to mean the defaults specified in the settings.value statements should match the "false" in settings.qml.

>the zwift_erg_filter* are used in the bike.cpp

What I was looking for here was a correction to the commented out code that might be put back in action one day. If this commented out code will never be used for these device types, then perhaps it should be removed altogether.
Unless you specify otherwise I'll change it to use the default constant copied from settings.qml.

cagnulein:
> stay false" means they are already false. They are true, it's the settings.qml value that is false.
> I will take your statement to mean the defaults specified in the settings.value statements should match the "false" in settings.qml.

Yes exactly:)

cagnulein:
> What I was looking for here was a correction to the commented out code that might be put back in action one day. If this commented out code will never be used for these device types, then perhaps it should be removed altogether.
> Unless you specify otherwise I'll change it to use the default constant copied from settings.qml.

It shouldn't be activated again that piece of software, so we can remove safely the lines commented

drmason789:
I've handled the last bluetooth device settings in #962 (i.e. setting it for all devices), and in this PR I've just factored out the method for doing it to ensure consistency. If #962 is merged into master, I could then sync master with this PR to cover it for all devices. 

Todo:
- I have decided to revert the email defaults, for later analysis
- Revert all the tile defaults in homeform.cpp
- Consider comments for the settings key constants I've left for later consideration

I also notice that when one mouse hovers over the QStringLiteral constants in QtCreator, one doesn't see the value, just the macro. I believe I should determine if there is actually a benefit to using QStringLiteral for compile time constants and these may be better off using QLatin1String all around, or just char* values.

src/homeform.cpp
```diff
@@ -4032,13 +4032,13 @@ void homeform::sendMail() {
...
 
-    bool miles = settings.value(QStringLiteral("miles_unit"), false).toBool();
+    bool miles = settings.value(QZSettings::miles_unit, QZSettings::default_miles_unit).toBool();
     double unit_conversion = 1.0;
     QString weightLossUnit = QStringLiteral("Kg");
     double WeightLoss = 0;
 
     // TODO: add a condition to avoid sending mail when the user look at the chart while is riding
-    if (settings.value(QStringLiteral("user_email"), "").toString().length() == 0 || !bluetoothManager->device()) {
+    if (settings.value(QZSettings::user_email, QZSettings::default_user_email).toString().length() == 0 || !bluetoothManager->device()) {
```
drmason789, line 4041: I've left this one using the constant because it's just used to determine if there is in fact an email address.

cagnulein:
so this should be ready for merge @drmason789 ?

drmason789:
> so this should be ready for merge @drmason789 ?

Yes, if the latest build is successful.

You might want to satisfy yourself that the settings dumps to the debug log are okay. There are some settings that aren't in the constants, and some defined in the constant list that aren't always there.

Also, the content of qzsettings.cpp is just there for this test, and is an ongoing maintenance item - each time  a setting is added:
- to get the benefit of sharing the constant, and having the list available in intellisense, the new setting key and its default must be added to the qzsettings.h file.
- to update this test, the key and its default must be added to the array in qzsettings.cpp and the row count updated.

Perhaps a notice is required in the settings.qml file and the qzsettings.h file indicating this.

cagnulein:
for reference, the settings on windows registery are located in `Computer\HKEY_CURRENT_USER\SOFTWARE\Roberto Viola\qDomyos-Zwift`

now i'm trying to remove all of them, start the master version, copy the log and do the same with this PR. If they match I guess we're ok

cagnulein:
[master version setting debug log.txt](https://github.com/cagnulein/qdomyos-zwift/files/9713843/master.version.setting.debug.log.txt)

cagnulein:
[reg pr exported.zip](https://github.com/cagnulein/qdomyos-zwift/files/9714046/reg.pr.exported.zip)
[pr version setting debug log.txt](https://github.com/cagnulein/qdomyos-zwift/files/9714047/pr.version.setting.debug.log.txt)

cagnulein:
ok there is only one difference

master:
`"template_user_ids" QVariant(QStringList, ("user_QZWS", "user_qz", "user_vlc"))`

pr:
`"template_user_ids" QVariant(QStringList, ("user_QZWS", "user_vlc", "user_qz"))`

cagnulein:
I guess we can merge this, i will double check it later today the changes in the files, just to be sure! Thanks @drmason789

src/main.cpp
```diff
@@ -312,33 +312,33 @@ int main(int argc, char *argv[]) {
 
         // Android 10 doesn't support multiple services for peripheral mode
         if (QOperatingSystemVersion::current() >= QOperatingSystemVersion(QOperatingSystemVersion::Android, 10)) {
-            settings.setValue(QStringLiteral("bike_heartrate_service"), true);
+            settings.setValue(QZSettings::bike_heartrate_service, true);
```
cagnulein, line 315: @drmason789 there is a reason why here there is no default constant used?

cagnulein:
@drmason789 the android build grows from 92mb to 114mb :O i thought the opposite.

drmason789:
I'll try moving the assignment of the values in the cpp file.

cagnulein:
@drmason789 let me try this immediately ;)

drmason789:
> @drmason789 let me try this immediately ;)

Okay

cagnulein:
> Okay

@drmason789 i wrote it without thinking enough..what is your idea about putting in the cpp? create static functions?

drmason789:
Change the QZSettings namespace to a class and use class members.

See [here](https://cookierobotics.com/032/)

cagnulein:
hah ok i'm trying this approach https://github.com/cagnulein/qdomyos-zwift/tree/const-char-instead-of-qstring-for-qzsettings

drmason789:
I just tried my idea and the Windows build went down from 320Mb to 258Mb. 

I think your approach is better for the programmer (the main point of this issue) because the value can be seen in Qt by mouse hovering.

But I wonder what the compiler does with the char* constants: since QSettings.value doesn't have an overload specifically for char*, does the compiled code create a QString object at runtime each time one of these constants is used, or does the compiler create an implicit QString somewhere and pass the same one every time?

cagnulein:
@drmason789 probably is using qstring so it's not good for performance. could you please create a pr with your solution? i would like to try it
thanks

drmason789:
I tried using 
static const QString xxx = "xxx"; 

instead and the executable size went to 278Mb from the 258 for the strings in cpp solution.

cagnulein:
@drmason789 ah ok so i guess the result is the same of what o have done.

drmason789:
Which do you prefer?

1) QStringLiterals defined in the C++ file
2) QStrings instantiated from char * in h file



Item1) has the smaller output, but the programmer won't be able to see the value by mouse hovering, which might just be inconvenient for the defaults.
Item 2) has larger output, but nowhere near as large as using QStringLiterals in h file. Programmer can see exactly how the value is defined with mouse hover.

drmason789:
@cagnulein  I notice that in the workflow file, the Android build is commented out, pending a [github issue](https://github.com/jurplel/install-qt-action/issues/63) that has been closed:

https://github.com/cagnulein/qdomyos-zwift/blob/ee8c66d3e880d2a741682d458c4b21727f029158/.github/workflows/main.yml#L191-L193

Could it perhaps be restored?

cagnulein:
@drmason789 it's closed but not resolved. You can try to run it on your fork in order to check if it's ok

PR #972, opened 2022-10-05
#948 Set string values in cpp file to lower output size

drmason789:
#948


drmason789:
Another thing I notice with this, changing to the static const class members from static const char*, is the lack of thousands of lines of compiler warnings about unused variables.

PR #976, opened 2022-10-10
Adding ESLinker YPOO Mini Change treadmill

nzbrian:
This works on my YPOO Mini Change, at least enough to show speed, distance, KCals. It was showing 0.8 as the speed randomly after being on for a minute, when it was still at 0. It can't go less that 1k/h so I just ignored the 0.8. Uses different fields for KCals than the others. Doesn't have inclination, and the field that the other two use seems to just echo distance, so I've made the inclination read just return 0.


cagnulein:
thanks @nzbrian seems everything ok, i will merge this later today!

Issue #953, opened 2022-09-25
[REQ] Improve ANT+ support


jwatral:
Current implementation uses outdated library and has hard to resolve bugs (like #167).
As discussed with @cagnulein on Facebook Group, we could improve the ANT+ feature using other library from Garmin.
More details to be provided by @cagnulein.

cagnulein:
@jwatral This is our starting page https://www.thisisant.com/developer/ant/ant-in-android

My idea is porting this library inside qz.
So for starting you can create a simple Java app that, for example, send fake power and speed to your watch.

Also will be good to read hr by ant+

If it works we can add it to qz quite easily

jwatral:
@cagnulein I did some digging and here are the results:
- the ANT+ SDK - is only for reading sensor data via multiple ANT+ profiles
- the ANT SDK does supports custom communications, allows sending data via ANT (but without ANT+ profiles)
- there are some paid membership options as well, like this:
> Note that ANT+ Members can also [download the member early release Android ANT+ SDK version P.B2. MO1](https://www.thisisant.com/developer/ant/%7Bpage_562%7D), which also contains support for FE-C.

FE-C would be extremely useful for us, but I'm not sure if this will cover the server side or only the client one.

Your current implementation is based on ANT SDK (android_antlib_4-14-0.jar), which to my current understanding, is the only one we can really use, without buying membership. What's even more interesting - in the official download page, there seems to be an older version (4-12-0) so really no point in changing the lib.

I've contacted their support and requested an account - let see if they will able to help.

But maybe I missed smth?
Btw, are you already ANT member?

cagnulein:
Thanks @jwatral a lot of information that I didn't know!

> But maybe I missed smth?

Mmm not really, I guess I were just misinformed 

> Btw, are you already ANT member?

Nope unfortunately 

So our best hope is to fix the current implementation?
Maybe it's an easy task: if you want you can easily take my Java file to use in a single project to play with it.

What do you think?

jwatral:
>So our best hope is to fix the current implementation?

I think so, yes. But because the power and cadence are working fine, I have high hopes on this ;-)

> Maybe it's an easy task: if you want you can easily take my Java file to use in a single project to play with it.

Sure, will do that. Is there an easy way to build the android apk locally? I could then install it on the galaxy S7 and test it right away.

cagnulein:
> I think so, yes. But because the power and cadence are working fine, I have high hopes on this ;-)

[...]

PR #992, opened 2022-10-19
fixed speed value conversion in SpeedChannelController #167

### LINKED ISSUE #167: [BUG] ANT+ Speed Conversion
Beta 2.3.11.  It appears the speed transmitted via ANT+ from QZ is incorrect.  
When testing with ipSensorMan the app shows 64.2m/s (I'm guessing this is meter per second) which would convert to 143mph!  
From BKOOL, it reports a speed 4 times less than QZ.

jwatral:
Fixes #167


cagnulein:
Let me know when you feel confident to merge. I guess there is no space for going worst than the actual version :)

jwatral:
I think, we're good to go. Let's build a new QZ version and test it IRL.
I just wonder, where would be the best place to document, you have to set 100mm as wheel circumference in settings of ant+ speed sensor.
I think, we should also close #953 as there's no better option for ANT library as it seems after conversation with Garmin team.

cagnulein:
> I think, we're good to go. Let's build a new QZ version and test it IRL. 

ok

> I just wonder, where would be the best place to document, you have to set 100mm as wheel circumference in settings of ant+ speed sensor. 

I can add a note in the setting under the ant+ settings in qz. what do you think?

> I think, we should also close #953 as there's no better option for ANT library as it seems after conversation with Garmin team.

done!

PR #1024, opened 2022-11-04
Test Videos with higher fps Rates

Bepo7012:


Bepo7012:
Yes you can merge this. I can't imagine an impact to existing "normal" videos, the recalculation is only done if the difference between gpx and video is greater 60 seconds which should not happen except of videos that are fastened by default.

Bepo7012:
And: It's working like a Charm :-) Played around with a few videos yesterday, each was nicely played. for one video i had to use 240 fps becuse it was originally already recorded with 60fps.

cagnulein:
@Bepo7012 great! you're now our Steven Spielberg :D

Bepo7012:
:-D

Issue #1085, opened 2022-12-18
[REQ] Support for Computrainer with USB connection


jra650:
**Is your feature request related to a problem? Please describe.** 
No it would be a new connection type and a new supported trainer.

**Describe the solution you'd like**
Computrainers are the old original smart trainer and are still in use by lots of riders.  They can connect to Android Phones via USB with a FTDI to USB cable, There are many people looking for an Android solution for Computrainers. There is open source code from Golden Cheetah which connects to the Computrainer https://github.com/GoldenCheetah which can provide Computrainer specs and output signal information.

Ant Tester shows the Computrainer on my old Samsung Note 8.  QZ (qdomyos-zwift) does not recognize the Computrainer on the Android phone.  I am looking for a solution that will broadcast the Computrainer signal via ANT+ or BLE to another device for use with Zwift, Rouvy etc and receive a signal back to make the Computrainer a controllable trainer via this solution.

**Describe alternatives you've considered**
There is a windows solution (https://perfprostudio.com/WebHelp/CTBridge/index.htm?page=Setup.html.) but this requires a laptop just to rebroadcast the Computrainer signal.

**Additional context**
![20221218_104005](https://user-images.githubusercontent.com/83240285/208312236-f8f85284-675f-424a-b5fc-4164c7955a06.jpg)

cagnulein:
Perfect, this code is usable without changing anything https://github.com/GoldenCheetah/GoldenCheetah/blob/cabe078453cbe8e136b8adf8d4187a5be878671b/src/Train/Computrainer.cpp

Tomorrow I will implement it

cagnulein:
@jra650 
send me an email to roberto.viola83 at gmail.com indicating this ticket in order to add you to the internal tester. Thanks

cagnulein:
first code commit done. I'm ready to test @jra650

jra650:
Roberto – Thank you.  Yes I can test this if I can get a link to a downloadable app. - James

cagnulein:
as I said above send me an email with your google play account email

jra650:
Oh ok Sorry missed that.  I purchased the app on the ***@***.*** account. - James

cagnulein:
@jra650 you have to send me an email, github hides the email :)

cagnulein:
@jra650 I thought I wrote the steps, my fault.
[...]

PR #1086, opened 2022-12-18
Computrainer

cagnulein:


cagnulein:
@drmason789 just for your info: i took the internal code of this PR from GoldenCheetah project and check how they use the serial port: they don't use at all the qserialport module. Probaly because they saw similar issue too I guess

cagnulein:
#1085

PR #1092, opened 2022-12-21
Usb serial for android

cagnulein:
@drmason789 maybe you're interested in this too. It works like a charm on my test device


cagnulein:
#1085

PR #1110, opened 2023-01-02
issues with ant+ speed sensor fixed by migration to speed and cadence

### LINKED ISSUE #1015: [BUG] ANT+ speed sensor unreliable
### Describe the bug
The ant+ speed sensor from QZ has some weird stability issues - it either disconnects or broadcasts 0.0 values periodically.

### To Reproduce
1. Install QZ on ANT+ enabled device (in my case both Samsung Galaxy S7 and A5)
2. Enable `Ant+ Cadence` option in `Ant+ Options` in `Settings`
3. Restart QZ
4. Turn on stationary bike (in my case Domyos EB 900) -> QZ picks it up and connects
5. Start `Bide indoor` activity on Garmin Watch (in my case Forerunner 255)
6. Wait till watch picks up speed sensor
7. Set circumference to `100` in sensor options on the watch
8. Start pedaling and observe speed value readings on the watch

### Expected behavior
Speed value reading is correct and stable

### Actual behavior
Galaxy S7 - speed sensor disconnects every ~20 seconds, leaving ~3-4 second gaps in speed value reading (see screenshot from garmin connect)
Galaxy A5 - each second, the speed value oscillates between showing correct value, and 0.0  

image

### Smartphone (please complete the following information):
 - Device: Samsung Galaxy S7 and A5
 - OS: Android 8.0 (on both)
 - QZ Version: 2.11.83

### Append a debug log

//TODO
  
### Additional context
The problem was uncovered after fixing speed value reading in #167 and shortly described in #953 

### Analyses so far

**Notes:**
- *The debugging of QZ was done using Qt 5.15.2 package with `Qt Creator`*
- *The dummy app is an Android app created in the newest version of Android Studio, with Ant related classes copied over directly from QZ codebase. The dummy app is run on devices directly from Android Studio*

**Question** - maybe that's an issue with the watch?
**Outcome** - No, I've tried to read the speed value using yet another samsung phone with installed [ANT+ Plugin Sampler](https://play.google.com/store/apps/details?id=com.dsi.ant.antplus.pluginsampler&hl=en&gl=US) app. Same issue can be observed, with some additional details. During the "disconnection" the estimated timestamp stops to increment for a few seconds and the receiver changes status to `SEARCHING`

**Question** - is the behavior the same while QZ is used on different Android devices?
**Outcome** - no:
- on Galaxy S7 - speed sensor disconnects every ~20 seconds, leaving ~3-4 second gaps in speed value reading (see screenshot from garmin connect)
- on Galaxy A5 - each second, the speed value oscillates between showing correct value, and 0.0  

**Question** - do you observe the same behavior in the dummy app?
**Outcome** - no, no matter what I do, the dummy app behaves correctly and speed sensor is stable.

**Question** - so maybe it's something with passing the speed value from Qt/c++ to Android/java?
**Outcome** - no, I've tired to disable the `antObject` management completely from `KeepAwakeHelper` and `bluetooth.cpp` and add dummy increments of timestamp and revolutions in `SpeedChannelController` - it didn't help. 

**Question** - is there something interesting in the logs?
**Outcome** - no, nothing obvious so far (both in core and android logs)

**Question** - why is the power sensor, ok and speed not?
**Outcome** - I don't know yet. I've ruled out the calculation difference (dummy revolutions and time increments didn't help). The other thing that's different is different broadcast frequency 4.04Hz vs 4Hz and number of channel pages

**Question** - Have you tried to completely disable power and heart sensor channel controllers and only leave SpeedChannelController running? 
**Outcome** - Yes, it didn't change anything

**Question** - Have managed to obtain stable speed readings from QZ app in any setup?
**Outcome** - Yes, if I start the channels from MyActivity and set some dummy speed values the speed readings are fine BUT, as soon as the QZ connects to the stationary bike, the problem starts occurring (still with dummy value set in `onCreate` in MyActivity.

**Question** - Have you tried to enable `fake device` and check if that helps with the issue? 
**Outcome** - Yes, it helps - the speed reading is rock stable using with `fake device` enabled and `Speed` variable set to a constant `22` in `fakebike.cpp`. What's more - the `virtual device` option does not have any influence on that issue.

**Question** - Have you tried to profile android app using Android Studio profiler? 
**Outcome** - Yes, there wasn't nothing obviously wrong with threads / cpu and memory usage. Looks like, we need a deeper look, including dmesg

### What else we can try
That's a good question actually. I'm a bit lost at the moment. I'll have another look at speed sensor channel paging docs and try to spot a potential timing issue. I'll document all further developments in this ticket.
@cagnulein - any other ideas, what we can try?

jwatral:
Fixes #1015

### DIFF
--- a/src/android/src/ChannelService.java
+++ b/src/android/src/ChannelService.java
@@ -116,6 +116,9 @@ public class ChannelService extends Service {
             if (null != powerChannelController) {
                 powerChannelController.cadence = cadence;
             }
+            if (null != speedChannelController) {
+                speedChannelController.cadence = cadence;
+            }
         }
 
         int getHeart() {
--- a/src/android/src/SpeedChannelController.java
+++ b/src/android/src/SpeedChannelController.java
@@ -16,6 +16,7 @@
 package org.cagnulen.qdomyoszwift;
 
 import android.os.RemoteException;
+import android.os.SystemClock;
 import android.util.Log;
 
 import com.dsi.ant.channel.AntChannel;
@@ -32,15 +33,17 @@ import java.util.Random;
 
 public class SpeedChannelController {
     // The device type and transmission type to be part of the channel ID message
-    private static final int CHANNEL_SPEED_DEVICE_TYPE = 0x7B;
+    private static final int CHANNEL_SPEED_DEVICE_TYPE = 0x79;
     private static final int CHANNEL_SPEED_TRANSMISSION_TYPE = 1;
 
     // The period and frequency values the channel will be configured to
-    private static final int CHANNEL_SPEED_PERIOD = 8118; // 1 Hz
+    private static final int CHANNEL_SPEED_PERIOD = 8086; // 1 Hz
     private static final int CHANNEL_SPEED_FREQUENCY = 57;
 
     private static final String TAG = SpeedChannelController.class.getSimpleName();
-    public static final int SPEED_SENSOR_ID = 0x9e3d4b65;
+    public static final int SPEED_SENSOR_ID = 0x9e3d4b33;
+
+    private static final double MILLISECOND_TO_1_1024_CONVERSION = 0.9765625;
 
     private AntChannel mAntChannel;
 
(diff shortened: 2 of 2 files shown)


cagnulein:
oh wow! did you fixed it?!

cagnulein:
@jwatral i guess I fixed (at least it works perfectly on my garmin edge now!)
Check this https://github.com/cagnulein/qdomyos-zwift/pull/1550

PR #1111, opened 2023-01-02
Automated testing with Google Test #911

drmason789:
PR for #911 

@cagnulein 

This changes the file system so that there is now an overall project file qdomyos-zwift.pro in the root directory, and the content of src/qdomyos-zwift.pro is now shared between the app and lib forms, src/qdomyos-zwift.pro and src/qdomyos-zwift-lib.pro by way of the qdomyos-zwift.pri file.

Each time a file is added to the src/qdomyos-zwift.pro file in other PRs or branches, the new content of that .pro file will need to be put into the .pri file in this PR. So I hope this review doesn't go on for long. I believe we could merge the changes in the src project separately to avoid this.

I don't think the current organisation is ideal. I think it would be better to separate the device functionality into the lib project, the UI functionality into a new app project, and have the app and test projects link to the lib project. I've made an attempt at this and didn't find it all that hard, although in my way there's an IOS only section I am not in a position to work with at the moment.

I've updated the CI workflows to accomodate the new structure, but I notice the Android workflow build is not working. I've made it get further by checking out the googletest repo, and having it start its build in the root rather than the src folder. I'm not sure yet, where to go from here. Perhaps you can spot it immediately.


drmason789:
@cagnulein I've used Pascal cased class names in the test project. I find these easier to read, but if you want consistency, I am happy to make them totally lower case.

cagnulein:
thanks @drmason789 amazing work!
i will start the review tomorrow! hopefully I will finish this, this week

cagnulein:
ok let's start this from the android build. looks like it's missing the llvm-ar
the ar should be /usr/local/lib/android/sdk/ndk-bundle/toolchains/llvm/prebuilt/linux-x86_64/bin/ but it's probably there with another name. i will do some test later today. It should be easy to fix

cagnulein:
strange, llvm-ar is present https://github.com/cagnulein/qdomyos-zwift/actions/runs/3827888223/jobs/6512880975
i'm going to run now, i will check this later

cagnulein:
so strange, build works on my mac...I'm still investigating

drmason789:
> so strange, build works on my mac...I'm still investigating

Is the executable on executable search path on your Mac, but not in whatever environment that's running in on github?

cagnulein:
no i checked it's in the same folder of the clang++ on github. 
maybe the file not found is not about the llvm-ar executable. i have to check this

cagnulein:
with my last commit @drmason789 i guess I fixed it, let's see if it can finish the run. I will start tomorrow the review

cagnulein:
@drmason789 fixed! now i'm fighting with the m3i android module

cagnulein:
hi @drmason789 i added androidextras to the google test project but it still complains about qandroidjini header file...I guess we are missing some android define somewhere in the google tests. Do you have any idea?

drmason789:
@cagnulein I've made some modifications and now have new errors in github CI, and I'm dealing with 

image

in my Windows and Linux environments - only for Android builds. Running qmake at any level doesn't fix it.

cagnulein:
yeah i saw this, this morning. thanks for the previous commit. i'm still thinking about the fix of this

drmason789:
I think more from qdomyos-zwift.pri needs to go in the defaults.pri file in the root, but I'm not sure how much of the content actually applies to the test project too - I still need to work this out.

drmason789:
Currently there seems to be a problem downloading from the ubuntu server.

drmason789:
The tests are out of date due to changes in bluetooth.cpp for some devices, e.g the trx* devices. I don't have time to fix these at the moment.

cagnulein:
hah right i should update them
too. i will try to do when i will merge this

cagnulein:
OK I checked the modifications and are fine to me. I hope to run a quick test by end of the weekend on the ios and android app just to make sure that we don't have any regression and then we're good to go

drmason789:
I don't have the equipment to do iOS builds.

cagnulein (approved):

cagnulein:
> I don't have the equipment to do iOS builds.

no problem I will do it starting from tomorrow, thanks again for your help!

cagnulein:
it builds fine also on iOS. I will do just a small test on a real device later today but I guess we're good to go

Issue #1150, opened 2023-01-13
[BUG] installation procedure seems broken


pssara:
**Describe the bug**
the installation on a Linux system and a raspberry pi is described here https://github.com/cagnulein/qdomyos-zwift/blob/master/docs/10_Installation.md
After I complete the steps of cloning the repository and updating the submodules, I change the directory to the src and execute qmake.
qmake fails to find any projects.

![image](https://user-images.githubusercontent.com/13958895/212411491-57c4063e-2900-46ff-bce2-5db56da12b36.png)

cagnulein:
do a qmake without entering on src 

this is due to a pr that we merge 2 days ago 

@drmason789

pssara:
thanks, it still doesn't compile.
In file included from powerzonepack.h:21,
                 from peloton.h:5,
                 from homeform.h:8,
                 from nordictrackifitadbbike.cpp:2:
trainprogram.h:113:13: error: ‘QRecursiveMutex’ does not name a type; did you mean ‘QBasicMutex’?
     mutable QRecursiveMutex schedulerMutex;
             ^~~~~~~~~~~~~~~
             QBasicMutex

drmason789:
If it is only required to run the app, perhaps change to the src directory and specify the qdomyos-zwift.pro file in the qmake command.

cagnulein:
@pssara did you fix it?

pssara:
@cagnulein Thanks Roberto, I compiled it just fine. There was an additional dependency, the rasbian should be at the bullseye version. (Qt 15.5)

cagnulein:
@pssara ok if you want make a PR about these, thanks!

PR #1156, opened 2023-01-14
Build corrections 1150

drmason789:
#1150 I've made some changes to make it build and run under Linux (Ubuntu 22.04).

I've kept these instructions just for building the app.

I won't be in a position to test on the Raspberry Pi for about 1 week.


cagnulein:
thanks!

Issue #1189, opened 2023-01-20
[REQ] Specify profile to use via cmd-line


Rossbro2:
**Is your feature request related to a problem? Please describe.**
Not a problem per-se, mostly a user-experience improvement.  You can change profiles after QZ is started, but it takes extra clicks and a reboot of the app to accomplish this.  It would be much quicker & easier to be able to specify which profile you'd like loaded via cmd-line instead of requiring users to start QZ, click the menu, select a profile to use, then restart the app.

**Describe the solution you'd like**
Ability to start the app with a specified profile instead of limiting the user to whatever profile was last used.

i.e.:
`qdomyos-zwift.exe -profile Kyle`
or
`qdomyos-zwift.exe -profile SomeoneElse`

**Describe alternatives you've considered**
- Not requiring a reboot after selecting a different profile (as Roberto mentioned, this would require a lot of rework and a rewrite)

**Additional context**
See [this comment](https://www.facebook.com/groups/149984563348738/posts/536715541342303/?comment_id=536824697998054&reply_comment_id=571129897900867&notif_id=1674193764526084&notif_t=group_comment_mention) on facebook

@cagnulein 👀

cagnulein:
done @Rossbro2 when the build is completed, you will see the new version here https://github.com/cagnulein/qdomyos-zwift/actions/runs/3968799815 on the bottom

Rossbro2:
> done @Rossbro2 when the build is completed, you will see the new version here https://github.com/cagnulein/qdomyos-zwift/actions/runs/3968799815 on the bottom

Whoa, that was fast!  Thanks, @cagnulein! I'll give it a test and close after confirming!

Rossbro2:
@cagnulein  hmm I've tried several different things and it's still not starting with a different profile specified in the cmd-line.  Am I doing something wrong?  Has this been tested?

cagnulein:
@Rossbro2 sorry i forgot to mention that you need to add the extension so

-profile name.qzs

cagnulein:
i added the extension directly in the code here https://github.com/cagnulein/qdomyos-zwift/actions/runs/3973598085

let me know and in case send me a debug log

now run with -profile name

cagnulein:
@Rossbro2 any news?

Rossbro2:
Unfortunately, it still doesn't work :( here are some screenshots:

[...]

PR #1190, opened 2023-01-20
Specify profile to use via cmd-line #1189

cagnulein:


cagnulein:
#1189

Rossbro2:
Closed #1189 as this is working as expected - thanks @cagnulein!

cagnulein:
great! thanks for the feedback and for the coffees @Rossbro2 !

Issue #1222, opened 2023-01-28
[REQ] Decouple bluetoothdevice classes from UI (i.e. IOS lockscreen code)


drmason789:
Ulterior motive: decouple device logic from UI to make it simpler to split main project into app and lib projects, and also to inject a test stub in place of the IOS lockscreen code to test that it's actually being called, on non IOS operating systems, and also, to avoid repeated code.

There is code repeated in various sublcasses of bluetoothdevice that puts some information on the lockscreen of various IOS devices. This code is disabled in non-IOS builds with compiler directives.

e.g.

https://github.com/cagnulein/qdomyos-zwift/blob/099bbf7d9c2c4f1dc038d43e2040e4a02f59cc24/src/yesoulbike.cpp#L170-L180

The classes that have this are:
activiotreadmill
bhfitnesselliptical
chronobike
computrainerbike
concept2skierg
cscbike
domyosbike
domyoselliptical
domyostreadmill
echelonconnectsport
echelonrower
echelonstride
fakebike
fakeelliptical
faketreadmill
fitplusbike
fitshowtreadmill
flywheelbike
ftmsbike
ftmsrower
horizongr7bike
horizontreadmill
inspirebike
keepbike
kingsmithr1protreadmill
kingsmithr2treadmill
lifefitnesstreadmill
m3ibike
mcfbike
mepanelbike
nordictrackelliptical
nordictrackifitadbbike
nordictrackifitadbtreadmill
npecablebike
pafersbike
proformbike
proformelliptical
proformellipticaltrainer
proformrower
proformtreadmill
proformwifibike
proformwifitreadmill
renphobike
schwinnic4bike
shuaa5treadmill
skandikawiribike
smartrowrower
snodebike
solebike
soleelliptical
solef80treadmill
stagesbike
strydrunpowersensor
tacxneo2
technogymmyruntreadmill
truetreadmill
trxappgateusbbike
trxappgateusbtreadmill
ultrasportbike
wahookickrsnapbike
yesoulbike

The classes that don't have it are:
bowflext216treadmill
bowflextreadmill
domyosrower
eliterizer
elitesterzosmart
eslinkertreadmill
fitmetria_fanfit
heartratebelt
iconceptbike
nautilusbike
nautiluselliptical
nautilustreadmill
octaneelliptical
octanetreadmill
paferstreadmill
smartspin2k
spirittreadmill
sportsplusbike
sportstechbike
technogymmyruntreadmillrfcomm
[...]

PR #1262, opened 2023-02-11
#1222 factored out lockscreen functionality into a separate class

drmason789:
Phase 2 of #1222, following PR #1241 

Introduced QZLockscreen and QZLockscreenFunctions "interfaces".

QZLockscreenFunctions contains Peloton workaround code and the other update that is done for numerous devices, and manages a reference to a QZLockscreen object. 

In IOS, this is an IOS-only implementation which calls the original lockscreen class, which now implements QZLockscreenFunctions.

Ultimately, the test project would inject test implemenations of these interfaces to test the functions are called in a non-IOS environment. Also, it allows separation of the UI code into a separate project.


drmason789:
This currently has changes to over 180 files, but it's really just the 175 of #1241 with  few extras.

drmason789:
Merged into #1241

PR #1288, opened 2023-02-20
#1287 unit test fixes

### LINKED ISSUE #1287: [BUG] Unit test problems
1. Exclusion devices are not "detected" using their expected name, instead using the name of the device they are supposed to exclude. This is fixed in #1223 and resulted in some additional test passes.
2. Enabling/disabling configurations for classes that define more than 1 are not actually used.
3. The CSC Bike detected using JOROTO-BK- prefix doesn't account for casing and settings

Not quite bugs:
1. When the wrong type of device is detected, show the name of the type of the object that was returned instead.
2. Skip the exclusion tests where there are no exclusions defined, just as the tests for non-detection for invalid names are skipped.

drmason789:
#1287 
- This reduces the number of actual tests by skipping the devices that don't have excluding devices defined.
- Uses the exclusion device name for detection of the exclusion device, not the excluded device name.
- Actually use multiple configurations when they are defined
- Failure messages say more about what was being tested and what happened instead
- Split CSCBike test in 2, 1 for the condition using the configuration, the other avoiding that and using the JOROTO-BK- prefix (case insensitively).
- Changed the testName property of the test data classes to return the user friendly provided name rather than the version stripped to the be the generated google test name, the latter not being used anymore and the former being more user friendly.


drmason789:
@cagnulein Assuming the checks all pass (I'm waiting for them as I type this), I'm ready to merge this.

Issue #1248, opened 2023-02-05
[BUG] Can't authorize Strava upload from Chromebook


laurenweinstein1:
While Strava upload authorization works fine from an Android tablet, it fails (upon clicking the Authorize button on the resulting Strava page after clicking "Connect to Strava" in QZ), with a:

This site can't be reached
127.0.0.1 refused to connect

The gist of the problem is the isolation of Android in the Chromebook, as described here:

https://stackoverflow.com/questions/47525471/chromeos-chrome-cant-reach-127-0-0-18080

It is possible to determine the actual local ip address (which typically appears to be 100.115.92.2) but it is not obvious how to get this used properly by the authorization interaction between QZ and Strava.

laurenweinstein1:
More:

https://support.google.com/chromebook/thread/50901565/how-to-access-localhost-from-chrome-browser-to-android-webserver?hl=en

laurenweinstein1:
I've found a workaround for this issue. It appears that (for some bizarre reason) while Chrome has problems in this situation, Firefox doesn't. 

So, a workaround is to install Firefox on the Chromebook from the Play Store, then set it to be the default browser. Then do the Strava connect from QZ and authorize via Firefox (only needs to be done once). This goes fine. Then (in my case), switch the default browser on the Chromebook back to Chrome.

Testing shows FIT data going up to Strava correctly.

cagnulein:
i guess a perfect solution will be running the browser from the qz directly, without open chrome

cagnulein:
using https://doc.qt.io/qt-5/qtwebview-index.html QWebView directly. I will experiment with this later

cagnulein:
@laurenweinstein1 I started working on this, I have to test this before send this to you

cagnulein:
@laurenweinstein1 i'm gonna upload the new version today with this fix, could you please try again with the 2.12.71 with the same scenario?

laurenweinstein1:
Did a deauth/reauth sequence. This version appeared to work fine with my Chromebook. Thanks!

cagnulein:
yeah!

PR #1296, opened 2023-02-21
Strava Auth Webview

cagnulein:
Can't authorize Strava upload from Chromebook #1248


cagnulein:
note for me: add a setting to have also the previous logic as a fallback

Issue #1308, opened 2023-02-23
qthttpserver for windows


cagnulein:
https://github.com/cagnulein/qdomyos-zwift/pull/252

PR #1309, opened 2023-02-24
Qthttpserver for windows #1308

cagnulein:
#1308


cagnulein:
@victorypoint could you check if this binary windows version works? I added the possibility to cast the floating window also from the windows version! Thanks

cagnulein:
I mean, test just your classic scenario, not the floating window :)

victorypoint:
@cagnulein, this build, 2.13.32 seems to run fine on 3 of my Zwift PCs with different hardware. I just tested fake treadmill mode but no problems encountered. I didn't see an option to turn on floating window so that wasn't tested.

cagnulein:
Wow thank you very much @victorypoint 

The floating window is testable with a web browser pointing on the 

Http://ipofthepc:7000/floating/floating.htm

The 7000 port is random and you can check which one is checking the log (is changing at every boot of qz). But that's not necessary to test it because I already did it

Thanks!

PR #1337, opened 2023-03-04
Nordictrack_C7.5

Takiora:
Hi @cagnulein I didn't forget about issue #1277 and have found some times to experiment directly on the elliptical.
Some functions didn't appear in the previous logs, like the inclination value when you click on the console panel.
I have also rewritten some if-else if-else to make it easier to read.
In forceResistance, for proform_hybrid_trainer_xt, looks like case 15 and 16 are missing so I added them.
A part that took me some time to understand is that if you want to calculate a uint16_t from a newValue.at(x), you need to make a double conversion (uint16_t)((uint8_t)(newValue.at(1)). If you write (uint16_t)(newValue.at(1)), you'll get a random number...
Everything works fine except the watts value. That value looks too high (more than 150watts) but I'm not sure how it's calculated and what is a normal or attended value.

### DIFF
--- a/src/nordictrackelliptical.cpp
+++ b/src/nordictrackelliptical.cpp
@@ -1026,7 +1026,7 @@ void nordictrackelliptical::characteristicChanged(const QLowEnergyCharacteristic
 
     // wattage = newValue.at(12)
 
-    if (proform_hybrid_trainer_xt)) {
+    if (proform_hybrid_trainer_xt) {
         uint8_t c = newValue.at(18);
         if (c > 0)
             Cadence = (c * cadence_gain) + cadence_offset;
--- a/src/nordictrackelliptical.cpp
+++ b/src/nordictrackelliptical.cpp
@@ -397,7 +397,91 @@ void nordictrackelliptical::forceResistance(resistance_t requestResistance) {
             writeCharacteristic((uint8_t *)res22, sizeof(res22), QStringLiteral("resistance22"), false, true);
             break;
         }
-    } else if (!proform_hybrid_trainer_xt) {
+    } else if (proform_hybrid_trainer_xt) {
+        const uint8_t res1[] = {0xff, 0x0d, 0x02, 0x04, 0x02, 0x09, 0x06, 0x09, 0x02, 0x01,
+                                0x04, 0x32, 0x02, 0x00, 0x4a, 0x00, 0x00, 0x00, 0x00, 0x00};
+        const uint8_t res2[] = {0xff, 0x0d, 0x02, 0x04, 0x02, 0x09, 0x06, 0x09, 0x02, 0x01,
+                                0x04, 0xa3, 0x04, 0x00, 0xbd, 0x00, 0x00, 0x00, 0x00, 0x00};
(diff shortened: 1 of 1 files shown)


src/nordictrackelliptical.cpp
```diff
@@ -397,47 +397,39 @@ void nordictrackelliptical::forceResistance(resistance_t requestResistance) {
             writeCharacteristic((uint8_t *)res22, sizeof(res22), QStringLiteral("resistance22"), false, true);
             break;
         }
-    } else if (!proform_hybrid_trainer_xt) {
+    } else if (proform_hybrid_trainer_xt) {
```
cagnulein, line 400: why did you modifiy this case? You should change the nordictrack_elliptical_c7_5 case only above

src/nordictrackelliptical.cpp
```diff
@@ -1197,74 +1248,74 @@ void nordictrackelliptical::btinit() {
             QThread::msleep(400);
             writeCharacteristic(noOpData10c, sizeof(noOpData10c), QStringLiteral("init"), false, false);
             QThread::msleep(400);
-        } else if (!proform_hybrid_trainer_xt) {
-            writeCharacteristic(initData10, sizeof(initData10), QStringLiteral("init"), false, false);
+        } else if (proform_hybrid_trainer_xt) {
```
cagnulein, line 1251: again why you changed this condition? you should change only the nordictrack_elliptical_c7_5 case

cagnulein:
check my comments @Takiora

src/nordictrackelliptical.cpp
Takiora: I think it is better to have an if structure like this:
if (nordictrack_elliptical_c7_5) {
} else if (proform_hybrid_trainer_xt) {
] else if (new_futur_nordictrack_proform_elliptical) {
} else {
}

src/nordictrackelliptical.cpp
Takiora: Same reason as before.

src/nordictrackelliptical.cpp
cagnulein: yes I agree with you but you modififed the content of the proform_hybrid_trainer_xt instead of the content of nordictrack_elliptical_c7_5 case. Am I wrong?

Takiora: I didn't touch the content of the proform_hybrid_trainer_xt, I have swapped the content of proform_hybrid_trainer_xt with the content of the else with a simple copy/paste to change !proform_hybrid_trainer_xt into proform_hybrid_trainer_xt. GitHub makes this simple swap so hard to read...
The only thing I have added in proform_hybrid_trainer_xt was case 15 and case 16 in switch (requestResistance) because they are missing from your original file.

cagnulein: hah sorry, yes github makes me think that you swapped them :)

cagnulein:
> Everything works fine except the watts value. That value looks too high (more than 150watts) but I'm not sure how it's calculated and what is a normal or attended value.

wattage is calculated using speed and inclination and your weight. 150W at which speed? if it's 9km/h it could be true.

PR #1379, opened 2023-03-26
Opening a ZWO file on Android from other folder leads to no file opened

cagnulein:
https://github.com/mahdize/CrossQFile/issues/3

### DIFF
--- a/src/CrossQFile.cpp
+++ b/src/CrossQFile.cpp
@@ -1,0 +1,165 @@
+#include "CrossQFile.h"
+#include <QFileInfo>
+#ifdef __ANDROID__
+#include <jni.h>
+#include <QtAndroidExtras/QtAndroid>
+#include <QtAndroidExtras/qandroidjnienvironment.h>
+#endif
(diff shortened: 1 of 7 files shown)


cagnulein:
on android we can't check anymore the extension because it returns something like this
`content://com.android.providers.downloads.documents/document/126`

reference https://stackoverflow.com/questions/58715547/how-to-open-a-file-in-android-with-qt-having-the-content-uri

cagnulein:
@p3g4asus did you face something similar in the past? I tried using the QFile directly with the "content" string. It works, but it reads only the first part of the file (for example if a file is 1024 bytes, it reads only the first 70 bytes with the qfile::readall function. It works flawlessy if the file is inside the QZ folder.

I'm trying to use the https://github.com/mahdize/CrossQFile library but i have a JNI GetMethodID error

cagnulein:
solved by this https://github.com/cagnulein/qdomyos-zwift/pull/1868

PR #1391, opened 2023-04-02
Concept2 Log

cagnulein:


cagnulein:
Screenshot 2023-04-02 at 19 09 00
@jonathancolledge

cagnulein:
@jonathancolledge integration done. I will add some more metrics, but most of the job is done. Now I'm waiting the review from Concept2.

I did a post on the indoor facebook group page, but no one commented it, so I don't think it has so interesting :)

Anyway when I will receive the go from Concept 2 I have to:
- change the url in the code
- add the new client id and scret it from the official portal
- add the metrics

cagnulein:
actual status:

<div class="gE iv gt" style="font-size: 0.875rem; padding: 20px 0px 0px; cursor: auto; color: rgb(34, 34, 34); font-family: &quot;Google Sans&quot;, Roboto, RobotoDraft, Helvetica, Arial, sans-serif; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; background-color: rgb(255, 255, 255); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;"><br class="Apple-interchange-newline">

Beatrice Sims (Support) | Beatrice Sims (Support) | lun 3 apr, 16:31 (9 giorni fa) |   |  
-- | -- | -- | -- | --
Beatrice Sims (Support)
a Roberto | a Roberto
a Roberto

</div></div></div></div></div></div>Beatrice Sims (Support)
lun 3 apr, 16:31 (9 giorni fa)
a Roberto

##- Please type your reply above this line -##
Your request (457672) has been updated. To add additional comments, reply to this email.

Beatrice Sims (Concept2)

Apr 3, 2023, 10:31 AM EDT

Hi Roberto,
Great, thanks. Could you please log each of the following workouts?
1 minute fixed time
200m fixed distance
Short interval workout with at least 2 intervals
 
Then we can verify that all the data is coming through as expected. In the meantime, you can also use the same steps you did for log-dev in the live [log.concept2.com](http://log.concept2.com/) environment to create your live API key. (Just replace log-dev with log in the URLs) By default it will have read-only access. Once we've verified the workouts listed above, we will add write access to your live key.



@jonathancolledge actually I don't see any interesting on the group or from email to this, so since I have to spend more time to complete this, I'm pausing this in order to bring it back when I have more time.

lpecan:
Assuming the API is pushing a 2 minute call out at "moderate" and then a 2 minute call out at "hard" then it would be useful to have a representation of your average pace during that interval while rowing, as well as being able to see a current average pace for that interval. Essentially an auto-"lap" that reflects as an interval in the ultimate c2 upload.

At the same time, would be useful to have a graphical representation of the avg pace current lap during the event as a tile.

cagnulein:
right @lpecan I was thinking if also this could be meaningful for bike and treadmill workouts. what do you think?

lpecan:
Yeah, I think that's a good idea generally--though I hadn't previously
really thought about it.

cagnulein:
ok so i guess i should do that one first, and then your request here, will be closed automatically :)

lpecan:
Makes sense. Do you want me to just delete the comment here? I assume you are still working on the main issue here regarding connecting to the c2 logbook

cagnulein:
Screenshot 2024-08-18 at 15 17 32

cagnulein:
answered back to concept2, let's see if it's all ok

lpecan:
> answered back to concept2, let's see if it's all ok

Happy to test if it's helpful

cagnulein:
> > answered back to concept2, let's see if it's all ok
> 
> Happy to test if it's helpful

Thanks, now it's 2 weeks and the concept2 guys didn't answer me back yet :(

PR #1415, opened 2023-04-15
Inclination improvement for Videos

Bepo7012:


cagnulein:
@Bepo7012 i did this in order to change the speed of a fake bike (using the target power tile) https://github.com/cagnulein/qdomyos-zwift/commit/1ff6295aaef664e436785fcb5b96d8de9878200e

update it from master

Actually I would like to build only things that can be insert to the QZ beta without issue because I don't have so much time to build a beta only for you (it's also hard to manage in the test flight app because also other people will receive your build, instead google play can have different channel).

So I'm reviewing this PR in order to have a generic modifications.

Check the comment in the PR

src/fakebike.cpp
cagnulein, line 1: you can remove this differences

src/qmdnsengine
cagnulein, line 1: this shouldn't be necessary, please align also this subrepo to yours from master

cagnulein (changes requested):

src/fakebike.cpp
Bepo7012: Done

src/qmdnsengine
Bepo7012: Done

cagnulein:
@Bepo7012 i don't understand your changes on the fake bike, you don't need to. Once the power is set it stays with also the power calculated. I tested it before committing it

Bepo7012:
> @Bepo7012 i don't understand your changes on the fake bike, you don't need to. Once the power is set it stays with also the power calculated. I tested it before committing it

@cagnulein  Sure, it worked nicely for speeding up. But slowing down the speed sometimes get's "clicky" and speeding up the fakebike if the video is on a high inclination results in the need to give very much power. I thought my change would make this better which was wrong too. another issue that can happen is for sure, that the video get's jumpy because the maxSpeed from the trainpogram is only applied when changing the power and not on a regular basis. But that's not a real problem at all.

I have reverted the Change, but now the current inclination is ignored in the call of the speed calculation to solve the need for many clicks to get a desired speed if the inclination is currently high or very low.

Slowing down the Bike still is sometimes strange (reducing power gives higher speed at first), but this also is not a real problem.

src/trainprogram.cpp
```diff
@@ -317,7 +320,9 @@ double trainprogram::TimeRateFromGPX(double gpxsecs, double videosecs, double cu
         double avgSpeedForLimit = avgSpeedFromGpxStep(currentStep + 1, 5);
         if (avgSpeedForLimit > 0.0) {
             bike *dev = (bike *)bluetoothManager->device();
-            dev->setSpeedLimit(avgSpeedForLimit * 3);
+            // bepo70: Video is recorded with Factor 2, Replay allows Factor 2 max, so set the speed Limit to 4 times speed to
```
cagnulein, line 323: what if the video was recorded with a classic factor of 1? I mean we should probably implement something generic. What do you think?

Bepo7012: Totally agreed. I have added the Recording Factor in the calculation of maximum Speed allowed.

cagnulein: thanks, but now you will have a different return on your code since, the default recording factor is 1 not 2. Maybe we should get the recording factor from a tag in the gpx file? what do you think?

Bepo7012: no, i have the same return because my videos have recording factor 2. The Recording factor is already calculated by the ration of gpx -length to video length, so there is no need to extra save it in the gpx file.

cagnulein: sorry, you're right!

cagnulein (approved):

PR #1435, opened 2023-04-25
UI theme

cagnulein:


blautob:
Hi Roberto,

just tried this out. Like the color picker, etc. One major part is missing for me (or maybe i do it wrong):
I would like to have different custom color modification per profile.
So i have a bike profile - top bar is red and a tread profile - top bar is blue.
cheers

cagnulein:
@blautob yes exactly. change the color for a profile, save it, load another one, change the color and save it. when you load a profile now you will see the selected color :)

blautob:
Got it to work! Thanks.

PR #1508, opened 2023-06-07
QT6, let's see how it goes

cagnulein:


drmason789:
@cagnulein I see you are using Qt 6.5.1 in the workflow file...this is now not even offered in the installer:

image

although I can get it by checking the Archive filter check box. 

They've made 6.8 the LTS release. [link](https://www.qt.io/blog/qt-6.8-released)

cagnulein:
I'm using 6.8.2, maybe I did a mistake in the actions

Let me check it
Thanks

Roberto Viola
http://robertoviola.cloud



Il giorno lun 10 mar 2025 alle 00:32 David Mason ***@***.***>
ha scritto:

.github/workflows/main.yml
```diff
@@ -575,7 +575,7 @@ jobs:
       - name: Install Qt Android
         uses: jurplel/install-qt-action@v3
         with:
-          version: '5.15.0'
+          version: '6.5.1'
```
drmason789, line 578: 6.5.1

.github/workflows/main.yml
```diff
@@ -688,7 +688,7 @@ jobs:
       - name: Install Qt iOS
         uses: jurplel/install-qt-action@v3
         with:
-          version: '5.15.2'
+          version: '6.5.1'
```
drmason789, line 691: 6.5.1

drmason789:
Linux, Android, iOS builds.

cagnulein:
> Linux, Android, iOS workflows.

Yes for now I'm using qt 6 only for windows to solve the Bluetooth incompatibility with windows 11

drmason789:
I had some build errors on the test project but I have them running in Windows now.

drmason789:
Are you accepting PRs into this branch?

src/devices/m3ibike/m3ibike.cpp
```diff
@@ -589,12 +595,11 @@ bool m3ibike::isCorrectUnit(const QBluetoothDeviceInfo &device) {
...
-        QHashIterator<quint16, QByteArray> i(datas);
-        while (i.hasNext()) {
-            i.next();
-            if (parse_data(i.value(), &k3) && (!valid_id(id) || k3.system_id == id)) {
-                return true;
+        QMultiHash<quint16, QByteArray> datas = device.manufacturerData();
+        QHash<quint16, QByteArray> uniqueDatas;
+        for (auto it = datas.begin(); it != datas.end(); ++it) {
+            if (!uniqueDatas.contains(it.key())) {
+                uniqueDatas[it.key()] = it.value();
```
drmason789, line 602: After this change, the function never returns true for QT_VERSION >= 5.12.0, so the M3I bike is not detected.

drmason789: ...thereby causing a test failure:
image

drmason789: While I don't have an M2I bike to test it on, assuming the scenario the test sets up is still valid, you would need to put back the check of the values:

```
for (auto it = datas.begin(); it != datas.end(); ++it) {
            if (!uniqueDatas.contains(it.key())) {
                uniqueDatas[it.key()] = it.value();
				
                if (parse_data(it.value(), &k3) && (!valid_id(id) || k3.system_id == id)) {
                    return true;
                }
            }
        }
```

But I'm not sure why you are trying to avoid duplicate keys? The keys are unique already, but I see that in Qt6 this data is now in a QMultiHash so there can be multiple values for each key. To access all the values for each key you need use datas.values(key).

cagnulein: done thanks!

src/inner_templates/chartjs/main_ws_manager.js
cagnulein, line 1: @p3g4asus Hi! How are you? I'm writing only to share what I found today (since you wrote this module!). On Qt6 we can't anymore use websockets and webserver on the same port, we have to use 2 different ports. I don't know which is the limitation, anyway now it's working again. Have a great day!

p3g4asus: Hi @cagnulein ! I am fine thanks. That's a weird limitation. I cannot find any reference to this limitation online. Happy to know that it works also in qt6 again.

PR #1577, opened 2023-07-25
Nordictrack Usb Treadmill

### LINKED ISSUE #54: NordicTrack T7.5s no data in app
Hi I'm trying to get this working on a T7.5s

The install went smoothly but no data is coming through.

I've attached the logs from the adb-debug.bat

Thanks

[logcat.log](https://github.com/cagnulein/QZCompanionNordictrackTreadmill/files/12116671/logcat.log)
[logcat.txt](https://github.com/cagnulein/QZCompanionNordictrackTreadmill/files/12116673/logcat.txt)
[2023-07-21_logs.txt](https://github.com/cagnulein/QZCompanionNordictrackTreadmill/files/12116678/2023-07-21_logs.txt)

cagnulein:


ciarancoffey:
Hi Roberto, do you need any help with this branch? I have a S45i treadmill which does work qdomyos-zwift, but if this a solution to having a barebones app that just starts, stops, sets the speed and incline and at a bonus can talk to zwift I'm happy to put some time into it.

cagnulein:
yes for sure but it will require a lot of work because we need to understand how to send the correct frames to the usb. are you ok with this?

cagnulein:
also this could be very useful https://github.com/nategreco/uBike

ciarancoffey:
> yes for sure but it will require a lot of work because we need to understand how to send the correct frames to the usb. are you ok with this?

I should be, hopefully.
I'll go at my treadmill with a screw driver and see what connections are where.
Is the aim to find out how the tablet talks to the controller?

cagnulein:
no actually that part should already done. it's usb. now we should just need to send the right command to it. so it should be only software. let me resume this PR

PR #1732, opened 2023-10-26
ADB on iOS

cagnulein:
https://github.com/zonghai-li/adb-ios/issues


cagnulein:
importing in the project https://github.com/zonghai-li/adb-ios/issues/11#issuecomment-1097900218

Issue #1805, opened 2023-11-22
Support for treadmill NordicTrack Z1300i (Bluetooth only model NTL99418.0)


JosePadron:
**Is your feature request related to a problem? Please describe.**
* No problem. Device is just not supported yet. 

**Describe the solution you'd like**
* Add support treadmill NordicTrack Z1300i

**Describe alternatives you've considered**
* None

**Additional context**
* Treadmill info:
   * Max incline: 12%
   * Max speed: 12 miles/h (19.3 Km/h)

* Steps during Bluetooth HCI snoop log:
   * Increase incline 0.5 percent up to 4.0%
   * Increase speed 0.1 miles/hour up to 3.0 miles/hour

* Log file:
   * `CE:DD:B5:2E:16:49 [  LE  ] I_TL`
[hci_snoop_2023_11_22_07_45_33.cfa.zip](https://github.com/cagnulein/qdomyos-zwift/files/13439940/hci_snoop_2023_11_22_07_45_33.cfa.zip)

cagnulein:
thanks! i will check this tomorrow!

Il giorno mer 22 nov 2023 alle 14:22 Jose Padron ***@***.***>
ha scritto:

JosePadron:
Data seems to be in byte 11 and 12 for the incline and velocity.  Values are multiplied by 100. 

Byte 10 is set to 2 for incline, and is set to 1 for speed. 

* Example 0.5% Incline (Raw data is 50):
0000   02 03 00 1b 00 17 00 04 00 12 0e 00 ff 0d 02 04   ................
0010   02 09 04 09 02 01 *02* **32 00** 00 44 00 00 00 00 00   .......2..D.....

* Example 4.0% incline (Raw data is 400):
0000   02 03 00 1b 00 17 00 04 00 12 0e 00 ff 0d 02 04   ................
0010   02 09 04 09 02 01 *02* **90 01** 00 a3 00 00 00 00 00   ................

* Example 3 mph speed (raw data is 482, 4.82 kmh = 3mph):
0000   02 03 00 1b 00 17 00 04 00 12 0e 00 ff 0d 02 04   ................
0010   02 09 04 09 02 01 *01* **e2 01** 00 f4 00 00 00 00 00   ................

JosePadron:
It looks like you need to update `proformtreadmill.cpp` to have one more condition in `forceIncline` and `forceSpeed`. 
Incline data and speed bytes doesn't need to be updated.
```c

[...]

PR #1817, opened 2023-11-24
Update label to include NordicTrack info

JosePadron:
Related to: https://github.com/cagnulein/qdomyos-zwift/issues/1805

- [x] Update label info
- [x] Bench test

`Proform z1300i` configuration works well for `NordicTrack Z1300i`. Different branding but the same software.


cagnulein:
i did it change it only that line. Thanks

JosePadron:
Thank you!

PR #1860, opened 2023-12-04
Python scripts update

victorypoint:
Update python scripts to support Windows DPI Aware and Zwift Windowed mode (Borderless Gaming install is now optional).

### DIFF
--- a/src/windows/zwift-incline-climb-portal.py
+++ b/src/windows/zwift-incline-climb-portal.py
@@ -1,6 +1,6 @@
 # iFit-Wolf3 - Autoincline control of treadmill via ADB and OCR
 # Author: Al Udell
-# Revised: August 16, 2023
+# Revised: November 23, 2023
 
 # zwift-incline-climb-portal.py - take Zwift screenshot, crop incline, OCR incline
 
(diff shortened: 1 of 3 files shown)


victorypoint:
@cagnulein, we should also update Python to 3.7 in this build as well with required components to keep aligned with AI Server build.

victorypoint:
Here are the Python 3.7 components and updates to install:

python -m pip install --upgrade pip
python -m pip install --upgrade setuptools
python -m pip install "protobuf<=3.20.2,>=3.1.0"
python -m pip install paddlepaddle==2.5.1
python -m pip install paddleocr
python -m pip install imutils
python -m pip install "Pillow<10.0.0"
python -m pip install opencv-python
python -m pip install numpy
python -m pip install pywin32

cagnulein:
done, let's see the result. Thanks!

cagnulein:
done @victorypoint https://github.com/cagnulein/qdomyos-zwift/suites/18802188985/artifacts/1096108496
are you able to test it? it's not so smaller anyway

victorypoint:
> done @victorypoint https://github.com/cagnulein/qdomyos-zwift/suites/18802188985/artifacts/1096108496 are you able to test it? it's not so smaller anyway

@cagnulein, I tested all OCR features and they work great. I guess with all the required components, the python package size is not that small after all. I would suggest embedding the path to the python exe in QZ rather than bothering with adding PATH entries to make the setup easier. So, calling python would use the path relative to QZ install folder - \python\x64\python.exe. There's no need to add python and \Scripts folders to PATH in this case.

cagnulein:
yes you're right, just did

```
  QProcessEnvironment env = QProcessEnvironment::systemEnvironment();
    QString currentPath = env.value("PATH");
    QString updatedPath = currentPath + ";" + QCoreApplication::applicationDirPath() + "\\python\\x64";
    env.insert("PATH", updatedPath);
    QProcess process;
    process.setProcessEnvironment(env);
    qDebug() << "run >> " << command;
```

let me know if it works correctly so I will merge this!
Thanks as always!

victorypoint:
> yes you're right, just did
> 
> ```
>   QProcessEnvironment env = QProcessEnvironment::systemEnvironment();
>     QString currentPath = env.value("PATH");
>     QString updatedPath = currentPath + ";" + QCoreApplication::applicationDirPath() + "\\python\\x64";
>     env.insert("PATH", updatedPath);
>     QProcess process;
>     process.setProcessEnvironment(env);
>     qDebug() << "run >> " << command;
> ```
> 
> let me know if it works correctly so I will merge this! Thanks as always!

@cagnulein, I just tested latest build with embedded python path. It works great. No problems at all. I made sure to remove all python entries from my Windows PATH. It even works great with the PaddleOCR CUDA install. Good job!

cagnulein:
amazing! let's merge this! Thanks!

PR #1947, opened 2024-01-04
Healthkit without apple watch

cagnulein:


cagnulein:
@OrislavK I created this for the iphone in order to don't kill the app while it's in background (hopefully). I will test on my own this evening and then I will send you it. It's a very alpha stage, this version it's a test to understand if even in this way ios will kill the app.

**Important:
at the very first boot, enable ALL the permissions to apple health for QZ and then restart the APP.**

cagnulein:
check this https://github.com/timfraedrich/OutRun

cagnulein:
https://github.com/cagnulein/qdomyos-zwift/issues/2608

cagnulein:
done in the #3594

Issue #1951, opened 2024-01-04
Adb.exe stays resident on QZ exit (Windows)


victorypoint:
I've been experiencing no auto-incline sometimes when I leave my Zwift/QZ PC on during the day and then turn on the treadmill after several hours to start a Zwift session. I narrowed it down to the ADB.exe process staying in memory on PC. If I kill it, auto-incline works.

I confirmed.. single ADB.exe stays resident on QZ exit - both via quit from menu and x on window bar.

This works to kill adb.exe process.. 
`cmd /c wmic process where name='adb.exe' delete`

cagnulein:
@victorypoint thanks for the feedback, could you check this build? https://github.com/cagnulein/qdomyos-zwift/actions/runs/7420467774

victorypoint:
@cagnulein, I tested windows-binary.zip and windows-msvc2019-ai-server-binary.zip builds. In each case, adb.exe was not killed when exiting QZ by hitting window x close or Quit from menu.

[windows-binary-debug-Fri_Jan_5_11_20_46_2024.log](https://github.com/cagnulein/qdomyos-zwift/files/13843730/windows-binary-debug-Fri_Jan_5_11_20_46_2024.log)
[windows-msvc2019-ai-server-binary1-debug-Fri_Jan_5_11_15_42_2024.log](https://github.com/cagnulein/qdomyos-zwift/files/13843731/windows-msvc2019-ai-server-binary1-debug-Fri_Jan_5_11_15_42_2024.log)
[windows-msvc2019-ai-server-binary2-debug-Fri_Jan_5_11_16_49_2024.log](https://github.com/cagnulein/qdomyos-zwift/files/13843732/windows-msvc2019-ai-server-binary2-debug-Fri_Jan_5_11_16_49_2024.log)
[windows-msvc2019-ai-server-binary3-debug-Fri_Jan_5_11_17_53_2024.log](https://github.com/cagnulein/qdomyos-zwift/files/13843733/windows-msvc2019-ai-server-binary3-debug-Fri_Jan_5_11_17_53_2024.log)

cagnulein:
thanks @victorypoint i added some debug log in the terminate function 
let's see if i can understand which is the issue
https://github.com/cagnulein/qdomyos-zwift/actions/runs/7423599840

unfortunately i don't have a windows build to try :(

cagnulein:
[...]

PR #1953, opened 2024-01-05
Adb.exe stays resident on QZ exit (Windows) (Issue #1951)

cagnulein:
#1951


cagnulein:
done https://github.com/cagnulein/qdomyos-zwift/actions/runs/7424793324

Issue #1976, opened 2024-01-09
[BUG] Pofiles screen not showing profiles on Raspberry PI OS


josetta:
**Describe the bug**
Hi, I am using app on RPI4, I am setting 2 profiles, one for me (I am using elliptical for run) and one for wife (she is using it as bike), but I can't see those profiles in Profiles part... on android, it is showing but on RPI4 it is not. I can see profile files exists in other folders, adding screenshots

**To Reproduce**
1. Go to Menu -> Profile
2. Fill profile name and save
3. Repeat point 1 and 2 with another name
4. Go to Menu ->Profile
5. Click on Profiles
6. You will see no profiles

**Expected behavior**
I should see saved profiles in Menu ->Profile screen

**Screenshots**
![obrázok](https://github.com/cagnulein/qdomyos-zwift/assets/40663601/0b5e1ac1-6562-4b2e-ad9b-08f55e147422)
![obrázok](https://github.com/cagnulein/qdomyos-zwift/assets/40663601/c3c0b879-ab4d-439c-bbf6-cf004fa1a75a)

**Desktop (please complete the following information):**
 - OS: Raspberry PI OS with Desktop
 - Version: Kernel 6.1, Debian version 12

**Append a debug log**

**Additional context**

cagnulein:
it shows something in the debug log about this? the code should point to the right folder

josetta:
I didn't notice anything in logs in console but I can add log :)

josetta:
added log after running, going to profiles and saving profile... there is some warning about using Layout and some undefined behavior, maybe it is connected to it?

[debug-ut_jan_9_13_15_27_2024.log](https://github.com/cagnulein/qdomyos-zwift/files/13873650/debug-ut_jan_9_13_15_27_2024.log)

cagnulein:
I guess this is the error

qrc:/profiles.qml  qrc:/profiles.qml:12: TypeError: Cannot read property
'top' of null

I don't know why it's happening only on the raspberry Pi, maybe a Qt issue
on ARM.
Also the training program and gpx list do the same?

Roberto Viola
Software engineer and open source enthusiast
http://robertoviola.cloud



Il giorno mar 9 gen 2024 alle ore 13:24 josetta ***@***.***>
ha scritto:

cagnulein:
give a try to this branch Pofiles-screen-not-showing-profiles-on-Raspberry-PI-OS-#1976

josetta:
[...]

PR #1977, opened 2024-01-09
Pofiles screen not showing profiles on Raspberry PI OS #1976

cagnulein:


cagnulein:
#1976

cagnulein:
i guess it's fixed in the qt6 branch #1508

Issue #2102, opened 2024-02-06
[SUGGESTION] Code organisation


drmason789:
@cagnulein 

Would you be interested in a reorganisation of the source code into folders, with a structure akin the test project.

e.g.

src\devices\bluetooth.*
src\devices\bluetoothdevice.*
src\devices\activiotreadmill\activiotreadmill.*
...
src\characteristics\characteristic*.*

src\virtualdevices\virtual*.*

I think this would make navigation easier, but on the downside, referencing the header files would need the foldername added.

cagnulein:
yes sure i guess the only changes needed are

include files
pri project file
and probably the xcode project that i can fix


Roberto Viola
Software engineer and open source enthusiast
http://robertoviola.cloud



Il giorno mar 6 feb 2024 alle 19:10 David Mason ***@***.***>
ha scritto:

drmason789:
@cagnulein I noticed there is a devices/kmlworkout folder now, with a couple of files that may not be used - it's not a "device" anyway so I think it's a typo/artifact of one of my bulk operations.  Perhaps this folder could be deleted?

cagnulein:
done, thanks

PR #2103, opened 2024-02-06
#2102 reorganisation of file system

drmason789:
#2102


cagnulein:
You are fast!

drmason789:
> You are fast!

I'm skilled in regular expression find and replace, and batch commands for the file system. For me this wasn't as big a task as many may initially think.

PR #2123, opened 2024-02-12
Added requirement that Debian 11 must be used with Raspberry Pi

Szepi:


cagnulein:
Actually I used Debian 10 also. which distro did you have before?

Szepi:
[Here](https://downloads.raspberrypi.com/raspios_full_armhf/images/raspios_full_armhf-2021-01-12/) is the link to the distro. It came with CanaKit.
This has Qt 5.11. It seems
This is also Debian 10, and [it appears](https://packages.debian.org/search?keywords=qt5) Debian 11 was the first distro that switched from Qt 5.11 to 5.15.
I can also see on the first link that some Raspbian based on Debian 10 already in 2021 had Qt 5.15. So some Raspbian based on Debian 10 may be fine, but at least one based on Debian 10 was not.

cagnulein:
ok

Issue #2127, opened 2024-02-13
[REQ] Test project updates


drmason789:
https://github.com/cagnulein/qdomyos-zwift/commit/f184b2b545044e035f4fc98809e9af9e469c994e#commitcomment-138579511

There's now a bluetooth::deviceHasService function being used to determine if a device has a service and this is being used to filter devices.

The test framework currently supports 1 invalid and 1 valid QBluetoothDeviceInfo object per test data class. 

This is not adequate to cover all cases, but will do most so far. The problem is where there are multiple rejection combinations (i.e. has 1 service AND does not have another).

Initially, update the test data classes to cover these scenarios. 

Eventually, update the test framework to cater for multiple QBluetoothDeviceInfo objects in a test data class.

PR #2134, opened 2024-02-16
2127 test project updates - support for multiple QBluetoothDeviceInfo configuration criteria

drmason789:
#2127


docs/50_writing_tests.md
```diff
@@ -0,0 +1,463 @@
...
+For example, if any of a list of names is enough to identify a device, or another group of names but with a certain service in the bluetooth device info, that will require multiple classes.
+
+### Recognition by Name
+
+Consider the detection code for the Domyos Bike:
+
+```
+ } else if (b.name().startsWith(QStringLiteral("Domyos-Bike")) &&
+                       !b.name().startsWith(QStringLiteral("DomyosBridge")) && !domyosBike && filter) {
+
```
drmason789, line 52: @cagnulein Perhaps this expression in bluetooth.cs requires an update. If the name starts with Domyos-Bike it will never start with DomyosBridge.

docs/50_writing_tests.md
```diff
@@ -0,0 +1,463 @@
...
+```
+QString powerSensorName =
+        settings.value(QZSettings::power_sensor_name, QZSettings::default_power_sensor_name).toString();
+...
+} else if ((b.name().toUpper().startsWith(QStringLiteral("STAGES ")) ||
+                        (b.name().toUpper().startsWith("TACX SATORI")) ||
+                        ((b.name().toUpper().startsWith("KICKR CORE")) && !deviceHasService(b, QBluetoothUuid((quint16)0x1826)) && deviceHasService(b, QBluetoothUuid((quint16)0x1818))) ||
+                        (b.name().toUpper()==QStringLiteral("QD")) ||
+                        (b.name().toUpper().startsWith(QStringLiteral("ASSIOMA")) &&
```
drmason789, line 341: https://github.com/cagnulein/qdomyos-zwift/blob/2834f6c5d816a93ac591f736363ebb233c07916f/src/devices/bluetooth.cpp#L1452

@cagnulein I have changed the code in this example. In bluetooth.cs it's checking that it starts with QD and is 2 characters long. I wonder if that should just compare the upper cased strings with == or a length greater than 2 is expected.

tst/Devices/StagesBike/stagesbiketestdata.h
```diff
@@ -42,3 +42,33 @@ class StagesBike2TestData : public StagesBikeTestData {
...
+        // !deviceHasService(b, QBluetoothUuid((quint16)0x1826)) && deviceHasService(b, QBluetoothUuid((quint16)0x1818)))
+
+        if(valid) {
+            result.setServiceUuids(QVector<QBluetoothUuid>({QBluetoothUuid((quint16)0x1818)}));
+            return result;
+        }
+
+        result.setServiceUuids(QVector<QBluetoothUuid>({QBluetoothUuid((quint16)0x1826)}));
+
+        // this doesn't check the invalid case where it has both services. Framework does not currently support multiple QBluetoothDeviceInfo scenarios
```
drmason789: Note to self: obsolete comment

drmason789:
@cagnulein I don't consider this ready for merging yet.

cagnulein:
ok thanks!

Roberto Viola
http://robertoviola.cloud



Il giorno sab 17 feb 2024 alle 15:52 David Mason ***@***.***>
ha scritto:

docs/50_writing_tests.md
```diff
@@ -0,0 +1,463 @@
...
+```
+
+This presents 3 scenarios for the current test framework. 
+1. Match names only (starts with:"STAGES ", starts with: "TACX SATORI", equals: "QD")
+2. Match the name "KICKR CORE", presence and absence of specific service ids
+3. Match the name "ASSIOMA" and the power sensor name setting starts with "Disabled"
+					   
+The framework is not currently capable of specifying all these scenarios in a single class. 
+The generated test data is approximately the combinations of these lists: names * settings * bluetoothdeviceInfo * exclusions.
+If a combination should not exist, a separate class should be used.
```
drmason789, line 353: @cagnulein I think this test framework would be easier to understand if instead of "here's a device class, here are the configs that should result in its detection, check that they do; here are configs that shouldn't, check that they don't" ;  "here's a set of configurations corresponding to a real world device, and an expected device class, check that these configurations cause the device class to be created; invalidate the configurations and check they don't". 

I think this would be longer, but easier to understand.

drmason789: I'm not looking to do this anytime soon.

drmason789:
@cagnulein I believe this is ready to merge.

docs/50_writing_tests.md
drmason789: For clarity, I changed the code in the documentation example, not  bluetooth.cs.

docs/50_writing_tests.md
cagnulein: yes that's right. Domyos-bike must use the domyosbike class, isn't it @drmason789  ?

docs/50_writing_tests.md
cagnulein: @drmason789 do you mean to scrictly checking the QD string? I have to blame that line to check how was the original device

cagnulein:
wow! you did an amazing work! even with the documentation! I really appreciate this!

docs/50_writing_tests.md
cagnulein: > For clarity, I changed the code in the documentation example, not bluetooth.cs.

yes  got it!

drmason789: > @drmason789 do you mean to scrictly checking the QD string? I have to blame that line to check how was the original device

I mean:

```
 (b.name().toUpper()==QStringLiteral("QD")) ||
```

cagnulein: sorry it's not clear to me @drmason789 
the code on master is

`(b.name().toUpper().startsWith(QStringLiteral("QD")) && b.name().length() == 2) ||`

what's wrong with this?

drmason789: It's not wrong, but it's doing a starts with check and a length check. Because the length of the text it is supposed to start with is equal to the length the entire string is supposed to be, it can only ever accept "QD" so why not just use the equality operator instead?

cagnulein:
hah sorry got it now :) yes you are right :)

Roberto Viola
http://robertoviola.cloud



Il giorno sab 24 feb 2024 alle 23:48 David Mason ***@***.***>
ha scritto:

Issue #2144, opened 2024-02-19
[REQ] Static Port for Browser app


wtafe:
**Is your feature request related to a problem? Please describe.**
Overlay app is great for single screen experience on desktop Zwift. When the app restarts, the port changes and while it isn't a huge hassle to identify the new port, having it statically assigned would reduce zwift setup time. 

**Describe the solution you'd like**
Static port option for overlay webpage.

**Describe alternatives you've considered**
N/A

cagnulein:
Thanks I will do asap!

Roberto Viola
Software engineer and open source enthusiast
http://robertoviola.cloud



Il giorno lun 19 feb 2024 alle 21:18 wtafe ***@***.***> ha
scritto:

cagnulein:
i did it here https://github.com/cagnulein/qdomyos-zwift/actions/runs/7971407074

I don't remember: are you using qz also on PC right?

wtafe:
Thanks!

Current iteration is zwift through wine on Linux and qz on Android.
Couldn't get zwift click accessory to pair on Linux and didn't feel like
switching dongles as I had a hard time finding one that supported my other
BT devices.

Thanks!

cagnulein:
hah ok so I did the patch only for windows, ok so i have to change what I did

PR #2145, opened 2024-02-20
Static Port for Browser app (Issue #2144)

cagnulein:
Static Port for Browser app (Issue #2144)


evan-kolberg:
[I developed a workaround](https://github.com/evan-kolberg/QZ-Python-WebSocket/blob/main/main_macOS.py) if someone would want to connect to the localhost websocket - Works on macOS, possibly Linux, and not Windows.  Thanks to @cagnulein for helping me via email. 

Screenshot 2025-02-20 at 8 03 18 PM

How I did this:

1. Launched QZ
2. Went to ```General UI Options``` in Settings
3. Scrolled down and clicked on ```OPEN FLOATING ON A BROWSER```  - whatever the heck that means
4. Inspected the page and went to Sources
5. Found the ```main_ws_manager.js``` and ```global.js``` files
6. Noticed that the localhost port changed each time the QZ app restarted
7. Since on macOS, ran ```lsof -iTCP -sTCP:LISTEN``` in Terminal.app
8. Saw that there were 4 options for ports and 1 of them was the socket port.
9. Translated the JavaScript code into Python
10. Implemented a function to find the websocket port from the parent process named ```qdomyoszw```
11. Was able to successfully read data from the open websocket

How to run the code (if on macOS, maybe Linux):

1. Launch QZ
2. Git pull the repo (or add it as a submodule) and pip install the websocket library
3. Run ```main_macOS.py```

Your output should look like this:
Screenshot 2025-02-20 at 8 19 27 PM


This doesn't directly give a static port for the browser app, but it offers a solution to those who want to connect to the localhost websocket. 

I still don't know what the ```Server Port: 36866``` option does under ```Experimental Features``` does. Is this for emitting Bluetooth signals with Mqtt or OSC? I don't know.

--
Evan

cagnulein:
The server port is, if I remember correctly, for the templates done by @p3g4asus

p3g4asus:
I had to review the code because I completely forgot it. However it is pretty simple. When a template is found, the settings are searched for the key "template_[templateId]_port". If it is found and it is not >0 and < 65535, that port number is used. If it is found and it is 0, the first time the template is started, the port is set randomly and saved in settings so that it will remain fixed for every next start. If it is not found, it is set to a default value that is 6666. So I think that if you delete that line of code for windows build, you will have QZ windows inner web server on port 6666.

cagnulein:
Superseded by #5078. The new implementation keeps the automatically selected inner web server port stable across restarts on all platforms, without the Android-specific `floating_port` setting, and handles an occupied saved port by falling back to a new dynamic port.

PR #2174, opened 2024-03-01
Added tests for bike power functions

drmason789:
To finish this PR, some thought will be needed about what it should test.

I have defined the erginterface abstract class, which acts as an interface for the powerFromResistance and resistanceFromPower functionality offered by the bike class. Currently the bikeergfunctions class implements this for bikes, by setting the bike's current cadence, using the corresponding function to get the resistance or power, then restoring the bike's original cadence.

The interface supports optionally defined (min, max) x (cadence, resistance) representing the domain the bike knows about so the tests adapt to the boundaries provided, where provided.

What is tested?
 - if there is a (minimum,maximum) (cadence,resistance) defined, if the (cadence, resistance) goes (below,above) the (minimum,maximum)  for a constant (resistance, cadence), does the power stop changing?
- if the bike::powerFromResistanceRequest function returns a power p for a specific cadence c and range of resistances rmin..rmax, does the bike::resistanceFromPowerRequest function return the rmin when called on p and c?

What is not tested?
- effects of watt gain and offset, resistance gain and offset
- wattsFromResistance function


drmason789:
@cagnulein is this something you want to continue with?

- I think it would be helpful to refactor the power/resistance functions so that are not dependant on bike state then put these functions into an implementation of erginterface (or whatever you'd like to rename it to), then make bike have a field referring to an erginterface object.
- The power tables you have been implementing could be passed as an array into an implementation of erginterface that generically takes care of storage and searching.

drmason789:
- I believe I can make many tests pass simply by returning 0 resistance or 0W for a negative cadence. 
- In all the cases I've looked at, the resistance from power doesn't invert the power from resistance. I wonder if you don't intend it to.
- I am using loops to check every cadence and resistance value on the domain boundary. I think this is excessive, and is producing long reports. If I continue with this I would define an increment to test with that's not 1.

cagnulein:
@drmason789 Actually I was looking to finalize the dynamic erg table for the bike where I don't have a static power table first. Do you agree?

Of course the dynamic power table will be applied only to the bike with automatic resistance and without the static power table.

cagnulein:
@drmason789 i mean this one https://github.com/cagnulein/qdomyos-zwift/pull/2175
if you want to review my code go on! (I didn't test myself yet)

drmason789:
> Actually I was looking to finalize the dynamic erg table for the bike where I don't have a static power table first. Do you agree?

Yes.

drmason789:
> @drmason789 i mean this one https://github.com/cagnulein/qdomyos-zwift/pull/2175
> if you want to review my code go on! (I didn't test myself yet)

Definately. I took a quick look and will have some comments about this later.

src/devices/computrainerbike/computrainerbike.cpp
```diff
@@ -183,10 +161,8 @@ void computrainerbike::innerWriteResistance() {
...
     if (requestResistance != -1) {
-        if (requestResistance > max_resistance) {
-            requestResistance = max_resistance;
-        } else if (requestResistance < min_resistance) {
-            requestResistance = min_resistance;
+        if(!this->resistanceLimits().contains(requestResistance)) {
+            requestResistance = resistanceLimits().clip(requestResistance);
         } else if (requestResistance == 0) {
             requestResistance = 1;
         }
```
drmason789, line 168: @cagnulein What's going on with these few bikes that have this form of clipping of the resistance request?

 i.e. here the min_resistance constant is -20 and the max_resistance constant is 100, and you clip the requestedResistance to that range, except if it's 0 then you make it 1.

drmason789: This is also in the proformwifi and proformtelnet bikes.

src/devices/sportstechbike/sportstechbike.cpp
```diff
@@ -99,12 +99,7 @@ void sportstechbike::update() {
 
         QSettings settings;
         uint8_t noOpData[] = {0xf2, 0xc3, 0x07, 0x04, 0x00, 0x00, 0x00, 0x00, 0xff, 0xff, 0xbe};
-        if (requestResistance < 0) {
-            requestResistance = 0;
-        }
-        if (requestResistance > 23) {
-            requestResistance = 23;
```
drmason789, line 106: Unusual, the max_resistance constant from the header is 24.

src/devices/computrainerbike/computrainerbike.cpp
cagnulein: min_resistance can't never be less than 0 in QZ. I mean I never encountered a bike with a negative resistance. So the current master can't handle this. That's why for inclination, instead, i'm using a -100 value as a "no action" value (because inclination can go lower than 0 (but not to -100 of course)).

did I answer to your question?

src/devices/sportstechbike/sportstechbike.cpp
cagnulein: @drmason789 because the resistance levels are 24 but the bike wants it starting from 0. we can replace that 23 with max_resistance - 1

src/devices/computrainerbike/computrainerbike.cpp
drmason789: I'm aware of the use of -1 and -100 to specify "no action" for resistance and inclination. But I don't know the purpose of the -20..100 range you're using here. I'll use 1..100 for this exercise unless you specify something else.

BTW, you might have noticed in some test code I've switched to C++17 and used the std::optional<T> template. This enables you to specify something like
```
std::optional<resistance_t> resistanceRequest;
```
and use resistanceRequest.hasValue() and resistanceRequest.reset() instead of have a "no action constant". If you are agreeable, I will do this for the resistance, inclination and power requests, in a different PR.

drmason789: (moved to a separate message so you can approve the whole message with a thumbs up)

BTW, you might have noticed in some test code I've switched to C++17 and used the std::optional<T> template. This enables you to specify something like
```
std::optional<resistance_t> resistanceRequest;
```
and use resistanceRequest.has_value() and resistanceRequest.reset() instead of having a "no action constant". If you are agreeable, I will do this for the resistance, inclination and power requests, in a different PR.

drmason789:
@cagnulein What is the intended difference between the functions wattsFromResistance and powerFromResistanceRequest?

drmason789:
@cagnulein Remember the resistance_t change? Would you be agreeable to me adding cadence_t, heartRate_t, power_t, pelotonResistance_t, inclination_t alongside it and propagating throughout the codebase, in a different PR?

src/devices/proformwifibike/proformwifibike.cpp
drmason789, line 1: @cagnulein This class defines (min_resistance, max_resistance) to be (-20, 100) but its pelotonToBikeResistance function maps from (0..100) into (1,16). I thought the pelotonToBikeResistance should map (0 (or 1?) ..100) to the whole range of the bike's resistance values. Am I wrong?

src/devices/computrainerbike/computrainerbike.cpp
cagnulein: sorry for the delay @drmason789 

the -20 in this case was copied by another project that I found so actually I didn't care about it. It could be that this bike can have also negative resistance (so it will make sense to use -100 for the resistance too). I didn't remember this.

> and use resistanceRequest.has_value() and resistanceRequest.reset() instead of having a "no action constant". If you are agreeable, I will do this for the resistance, inclination and power requests, in a different PR.

yes sure! thanks!

cagnulein:
> @cagnulein What is the intended difference between the functions wattsFromResistance and powerFromResistanceRequest?

@drmason789 it was intented to use for renpho bikes but at the end powerFromResistanceRequest is a dead code

cagnulein:
> @cagnulein Remember the resistance_t change? Would you be agreeable to me adding cadence_t, heartRate_t, power_t, pelotonResistance_t, inclination_t alongside it and propagating throughout the codebase, in a different PR?

@drmason789 yes sure!

cagnulein:
@drmason789 i added erg tables for all the bikes with the new ergtable module. Maybe it could impact also this

PR #2175, opened 2024-03-02
Dynamic Erg Table

cagnulein:


src/ergtable.h
```diff
@@ -0,0 +1,140 @@
...
+public:
+    ergTable(QObject *parent = nullptr) : QObject(parent) {
+        loadSettings();
+    }
+
+    ~ergTable() {
+        saveSettings();
+    }
+
+    void collectData(double cadence, double wattage, double resistance) {
```
drmason789, line 37: Something that stands out to me about this approach is that you're keying the data using floating point numbers. 
If I look at this code as in independent component that could be used by anything, rather than the 2 I see currently (fakebike and trxappgateusbbike) I see the potential for every sample collected to create a new record, causing the list to grow, and your estimateWattage function, which scans the whole table every time it's called, would take longer. Significantly longer on modern CPU? Maybe not, but I prefer to write code that doesn't stress an RP Pico.

A solution: bucketing. Key by integer cadence and resistance and round to nearest. Perhaps you're doing it implicitly already because the only values you're getting from the hardware for cadence and resistance are actually integer.

src/ergtable.h
```diff
@@ -0,0 +1,140 @@
...
+        }
+    }
+
+    double estimateWattage(double givenCadence, double givenResistance) {
+        // Placeholder for closest points. Initialize with max possible difference.
+        double lowerDiff = std::numeric_limits<double>::max();
+        double upperDiff = std::numeric_limits<double>::max();
+        DataPoint lowerPoint, upperPoint;
+
+        for (const DataPoint& point : dataTable) {
```
drmason789, line 54: Can you be sure this dataTable will be consulted and updated always on the same thread?

cagnulein: Actually nope, do you think about a lock mechanism?

src/ergtable.h
cagnulein: Actually I wrote without thinking about performance. For example they can be uint16.

I would like first to make a usable version and then optimize it :)

src/ergtable.h
```diff
@@ -0,0 +1,140 @@
+#ifndef ERGTABLE_H
+#define ERGTABLE_H
+
```
drmason789, line 3: Much as I love C#, I'm wondering why you have the implementation in the header. 

You probable already know, but if you add ergtable.cpp to the project, QtCreator in Windows will put the implementations in the cpp file for you from the context menu activated on the class name (ergtable):


image

src/ergtable.h
drmason789: e.g. (60.2, 200.1, 50.5), (60.1, 200.1, 50.7)...

src/ergtable.h
cagnulein: Not a fan too, I did because it was quicker to write. I'm usually write code from the phone with my daughter on my legs :)

src/ergtable.h
```diff
@@ -0,0 +1,140 @@
...
+            if (point.cadence == cadence && point.wattage == wattage && point.resistance == resistance) {
+                return true; // Found duplicate
+            }
+        }
+        return false; // No duplicate
+    }
+
+    void loadSettings() {
+        QSettings settings;
+        int size = settings.beginReadArray("DataPoints");
```
drmason789, line 102: Is this 1 dynamic ERG table for an unspecified device, per settings profile? Perhaps storage should include an identifier for the device it's for.

cagnulein: yes i thought about this, and the solution is to use a profile. so no need to save the device too

src/ergtable.h
```diff
@@ -0,0 +1,140 @@
...
+
+Q_DECLARE_METATYPE(DataPoint)
+
+class ergTable : public QObject {
+    Q_OBJECT
+
+public:
+    ergTable(QObject *parent = nullptr) : QObject(parent) {
+        loadSettings();
+    }
```
drmason789, line 31: I'm wondering why you've made this a QObject, with all that comes with it, but never use said functionality. Are you planning to have signals and slots?

trxappgateusbbike::_ergtable is never passed a parent in its construction.
If you put this code in this constructor and run the tests, it never throws because parent is always null.
```
        if(parent!=nullptr) {
            throw "Got a parent";
        }
```

cagnulein: yes i made this for future implementation. now it's not necessary of course :)

src/ergtable.h
drmason789: I'm using Qt's various mutexes and QMutexLocker in #899 .

drmason789: > I'm using Qt's various mutexes and QMutexLocker in #899 .

Have a classmember of type QMutex or QRecursiveMutex if you might need multiple locks. 
Use a QMutexLocker to lock it AND make sure it gets unlocked if there's an exception.

drmason789:
It would be simple to create a test for this. 

The TestSettingsTestSuite class would be a good starting point, I think. If someone were to copy that, and delete one of the test functions, rename and rework the other one to test the ergtable class, you'd be in business on that.

Initially it could just pass in a sequence of samples and check that it does what's expected, and it could be expanded from there, e.g. with different sample sets, and multithreaded access.

cagnulein:
> It would be simple to create a test for this.
> 
> The TestSettingsTestSuite class would be a good starting point, I think. If someone were to copy that, and delete one of the test functions, rename and rework the other one to test the ergtable class, you'd be in business on that.
> 
> Initially it could just pass in a sequence of samples and check that it does what's expected, and it could be expanded from there, e.g. with different sample sets, and multithreaded access.

i'm so outside of all of this :) if you can help you it will be great! otherwise #2138 it will be our guinea pig :)

drmason789:
> > It would be simple to create a test for this.
> > The TestSettingsTestSuite class would be a good starting point, I think. If someone were to copy that, and delete one of the test functions, rename and rework the other one to test the ergtable class, you'd be in business on that.
> > Initially it could just pass in a sequence of samples and check that it does what's expected, and it could be expanded from there, e.g. with different sample sets, and multithreaded access.
> 
> i'm so outside of all of this :) if you can help you it will be great! otherwise #2138 it will be our guinea pig :)

Are there other contributors interested in getting experience in Google test? I think it's a simple entry point. The type-parameterized device tests are relatively complicated.

cagnulein:
> Are there other contributors interested in getting experience in Google test? I think it's a simple entry point. The type-parameterized device tests are relatively complicated.

as far as I know, nope, there aren't :)

cagnulein:
what i'm usually doing in this situation is run the code on chatgpt @drmason789

drmason789:
> i'm so outside of all of this :) if you can help you it will be great! otherwise #2138 it will be our guinea pig :)

I've committed some help to this branch in my fork.

cagnulein:
that's great! is it possible to create a sort of a PR of a PR? I'm new to this

drmason789:
> that's great! is it possible to create a sort of a PR of a PR? I'm new to this

PR onto PR initiated.

tst/Erg/ergtabletestsuite.cpp
```diff
@@ -0,0 +1,62 @@
...
+
+
+
+void ErgTableTestSuite::test_wattageEstimation(const QList<DataPoint> &inputs, const QList<DataPoint>& expectedOutputs) {
+
+    TestSettings testSettings("Roberto Viola", "QDomyos-Zwift Testing");
+    testSettings.activate(); // previous settings should be restored automatically when this object goes out of scope via its destructor
+
+    // "just in case", clear the DataPoints settings so that the ergtable is blank
+    testSettings.qsettings.remove("DataPoints");
```
drmason789, line 21: This bit here is not ideal IMO. 
The TestSettings class is something I made to manage a QSettings profile for a test case. There's invisible cleanup happening here. I think it would be better if the ergTable's automatic saving and loading could be optional (and turned off in these tests).

cagnulein: sure it makes sense.

tst/Erg/ergtabletestsuite.cpp
```diff
@@ -0,0 +1,62 @@
...
+
+    // test that the input points are mapped directly
+    expected.append(inputs);
+
+    this->test_wattageEstimation(inputs, expected);
+
+    // Now try some interpolated points
+    expected.clear();
+
+    expected.append(DataPoint(65,100, 75)); // closer to (60,100,50) or (60,200,100)?
```
drmason789, line 58: Note the comment. I don't consider the expected value of 100 to be correct, it's just what the ergTable does currently.

src/ergtable.h
```diff
@@ -75,6 +75,13 @@ class ergTable : public QObject {
 
             qDebug() << "interpolation" << lowerPoint.wattage + (upperPoint.wattage - lowerPoint.wattage) * (cadenceRatio + resistanceRatio) / 2;
 
+            if(qIsNaN(cadenceRatio) || qIsInf(cadenceRatio)) {
```
drmason789, line 78: Can this approach be trusted on every platform? I'd prefer to just not divide by 0 in the first place.

cagnulein:
yes these macro are safe in Qt

Roberto Viola
http://robertoviola.cloud



Il giorno dom 3 mar 2024 alle 18:16 David Mason ***@***.***>
ha scritto:

src/ergtable.h
cagnulein: now are uint16

tst/Erg/ergtabletestsuite.cpp
cagnulein: i changed this in order to reuse our standard qzsettings

src/ergtable.h
```diff
@@ -0,0 +1,188 @@
...
+        saveSettings();
+    }
+
+    void collectData(uint16_t cadence, uint16_t wattage, uint16_t resistance, bool ignoreResistanceTiming = false) {
+        if(resistance != lastResistanceValue) {
+            qDebug() << "resistance changed";
+            lastResistanceTime = QDateTime::currentDateTime();
+            lastResistanceValue = resistance;
+        }
+        if(lastResistanceTime.msecsTo(QDateTime::currentDateTime()) < 1000 && ignoreResistanceTiming == false) {
```
drmason789, line 41: In defense of the RP Pico, you could do the boolean check first to determine if the (admittedly tiny) effort to get the time do the calculation is actually necessary.

src/ergtable.h
```diff
@@ -0,0 +1,188 @@
...
+            qDebug() << "skipping collecting data due to resistance changing too fast";
+            return;
+        }
```
drmason789, line 44: What if the cadence has changed but the resistance hasn't. This would throw away samples.

Also, I wonder if the wattage for a specific (cadence, resistance) would ever be different, e.g. if you were reading cadence from a cadence meter strapped to a pedal and power from a power meter inside the pedal.

src/ergtable.h
cagnulein: that boolean is only for gtests so it's useless on runtime

src/ergtable.h
cagnulein: > What if the cadence has changed but the resistance hasn't. This would throw away samples.

the purpose of this is to thrown away every samples after a resistance change for 1 second. what issue do you see in this?

> Also, I wonder if the wattage for a specific (cadence, resistance) would ever be different, e.g. if you were reading cadence from a cadence meter strapped to a pedal and power from a power meter inside the pedal, rather than the bike's computer calculating it from its own measurement of cadence and what its resistance level achieves.


here i'm evaluating the power from the bike or from the power pedal based on the user configuration. so it's transparent

drmason789: > > What if the cadence has changed but the resistance hasn't. This would throw away samples.
> 
> the purpose of this is to thrown away every samples after a resistance change for 1 second. what issue do you see in this?

Based on the qDebug comment, I think you're looking to avoid collecting samples while the resistance is changing quickly.

It's currently waiting for 1000ms of stable resistance before storing the samples but could ignore changing cadence and wattage while the resistance remains stable for 1000ms. 

Also, if the user gradually decreased cadence, causing the resistance to increase at a rate of >1/s to match a target power, you could lose seconds of samples along the (C,R) line. This might actually be quite hard to do in practice though, with differing difficulty based on how many resistance levels the bike offers (i.e. you might have more "stability" as defined by this metric, from a bike that offers 16 resistance levels, over one that offers 250).

cagnulein: > Based on the qDebug comment, I think you're looking to avoid collecting samples while the resistance is changing quickly.
> 
> It's currently waiting for 1000ms of stable resistance before storing the samples but could ignore changing cadence and wattage while the resistance remains stable for 1000ms.

Yes that's the purpose because some bikes could send resistance level changed without recalculate the wattage 


> Also, if the user gradually decreased cadence, causing the resistance to increase at a rate of >1/s to match a target power, you could lose seconds of samples along the (C,R) line. This might actually be quite hard to do in practice though, with differing difficulty based on how many resistance levels the bike offers (i.e. you might have more "stability" as defined by this metric, from a bike that offers 16 resistance levels, over one that offers 250).

Yes I considered this but I guess it's more important to don't add some spurious samples than adding more samples  :)

drmason789: > here i'm evaluating the power from the bike or from the power pedal based on the user configuration. so it's transparent

If the bike provides the cadence and power level all itself, it could be capable of working it out from its own internal table of (C,R,W), and always give you the same W for the (C,R). However, if the C and W are coming from independent sources that don't know the R, their individual margins of error may lead to a range of W for any given (C,R). So if the first (C,R,W) you collect has a sporadically bad W value, the current implementation is stuck with it.

cagnulein: this scenario is impossible because i'm resetting the table if the user change the power sensor setting. so the table is or only filled with bike metrics or only filled with power pedal one + resistance

src/devices/fakebike/fakebike.cpp
```diff
@@ -163,4 +165,35 @@ void fakebike::changeInclinationRequested(double grade, double percentage) {
...
+    //QSettings settings;
+    //bool toorx_srx_3500 = settings.value(QZSettings::toorx_srx_3500, QZSettings::default_toorx_srx_3500).toBool();
+    /*if(toorx_srx_3500)*/ {
+        qDebug() << QStringLiteral("resistanceFromPowerRequest") << Cadence.value();
+
+        if (Cadence.value() == 0)
+            return 1;
+
+        for (resistance_t i = 1; i < maxResistance(); i++) {
+            if (wattsFromResistance(i) <= power && wattsFromResistance(i + 1) >= power) {
```
drmason789, line 182: This function could be offered by the ergTable class. 

Also, this is doing the ergTable search twice for almost every resistance level from 1 to the result. You could save the computation by caching the values.
e.g.:
```
uint16_t power1 = wattsFromResistance(1), lastPower = power1;
for(resistance_t i = 1; i<maxResistance(); i++) {
   uint16_t thisPower = wattsFromResistance(i+1);
   if(lastPower <= power && thisPower>=power) {
        qDebug() << QStringLiteral("resistanceFromPowerRequest") << lastPower;
        return i;
   }
   lastPower = thisPower;  
}
return power < power1 ? 1:maxResistance();
```

cagnulein:
yes i saw this yesterday too (the qdebug part). i will fix it today

for the

resistanceFromPowerRequest

i would like to put it in the bike class

thanks

Roberto Viola
http://robertoviola.cloud



Il giorno gio 7 mar 2024 alle 23:25 David Mason ***@***.***>
ha scritto:

PR #2234, opened 2024-03-15
Android 5 Fix

cagnulein:


cagnulein:
on the android 5.1 simulator on my mac m2 i have this issue https://bugreports.qt.io/browse/QTBUG-84849
so let's try with qt 5.15.2

PR #2336, opened 2024-05-16
let's see if the tacx app gets QZ

cagnulein:


cagnulein:
#2327

cagnulein:
Maybe use the service data as peloton?

Issue #2566, opened 2024-09-03
[REQ] Skandika Morpheus Support


dadude2607:
The Skandika Morpheus is not recognized on start (bluetooth name HT2011015332). I have attached a wireshark capture and the datasheet.

it has power, resistance, rotation speed, heart rate. So actually a perfect candidate.

for the log I connected the bike, pedaled a few times and touched the heart rate sensor.

it would be really great if the bike is added.

Thanks and regards
Andre

[Skandika Morpheus.zip](https://github.com/user-attachments/files/16855249/Skandika.Morpheus.zip)

dadude2607:
[Skandika Morpheus.zip](https://github.com/user-attachments/files/16855242/Skandika.Morpheus.zip)

cagnulein:
thanks i will add it asap

Roberto Viola
Software engineer and open source enthusiast
http://robertoviola.cloud



Il giorno mar 3 set 2024 alle 22:27 dadude2607 ***@***.***>
ha scritto:

cagnulein:
done https://github.com/cagnulein/qdomyos-zwift/actions/runs/10696139694

@dadude2607 
Google is very slow in these days, reviewing the app on the store, even the beta ones. For this reason, in order to check right now if the patch that I did is right, you can use the android build in the link (check at the bottom of the page in about 1 hour). In order to test it you need first to remove the current QZ version on your phone, install the one that you downloaded.
This version is a 15 minutes trial (you can restart it as many times as you want) and I can unlock it if you want, send me an email to roberto.viola83@gmail.com mentioning this ticket and that you are on android.
Let me know if you have any questions

dadude2607:
Perfect, thank you very much. I will test it tomorrow evening and will give you a feedback

dadude2607:
Although I'm not at home to test, I reviewed your code. I see you expanded the size a HT* bluetooth name can be (from 11 to 12). So it should be discovered.

But in skandikawiribike.cpp the heart rate is not read from the packet. But the Morpheus transmits this data (Position 8, after calories high and low). What I see is:

`double skandikawiribike::GetSpeedFromPacket(const QByteArray &packet) {
    uint16_t convertedData = (convertHexToDec(packet.at(2)) * 100) + convertHexToDec(packet.at(3));
    double data = (double)convertedData / 10.0;
    return data;
}

double skandikawiribike::GetWattFromPacket(const QByteArray &packet) {
[...]

PR #2567, opened 2024-09-04
Added HR for Skandika Morpheus issue #2566

dadude2607:
Because the morpheus behaves like a mixture of the x-2000 and the wiry, I have introduced a new variable. Speed, watts and rpm are read out like the wiry, but the heart rate is read out like the x-2000. With this code the morpheus works for me. I have attached a screenshot in the issue #2566. to distinguish an x-2000 from a Morpheus it seems to be enough to pay attention to the number of characters of the bluetooth name


cagnulein:
thanks!

PR #2570, opened 2024-09-05
Patch 1

bkeown64:
[debugfiles.zip](https://github.com/user-attachments/files/16899462/debugfiles.zip)


cagnulein:
thanks i will check it asap

cagnulein:
hi log checked but i need a hcisnoof following this from ifit app while you are changing speed and inclination there from the app directly https://github.com/cagnulein/qdomyos-zwift/blob/master/docs/20_supported_devices_and_applications.md#android-device

bkeown64:
Sorry, I probably am doing this wrong.  I am not using the iFit app at
all.  I was trying to use this just with Peloton app.

PR #2572, opened 2024-09-05
Skandika Morpheus: calc speed based on watts fixed

dadude2607:
The speed could not be calculated based on watt, because the code was in an unsolvable if query. I have changed the if query to "cadence", like on the Schwinn bike (and others). Now the speed is calculated correctly as soon as you pedal. Before this fix the option "calculate speed based on watt" had be turned off to get speed values.

This is a screenshot taken with "calculate speed based on watt" turned on. And as expected, the speed on bike display and the speed on qdomyos-zwift ​​differ greatly.

speed_on_watt


dadude2607:
Fix is of course also for other skandika bikes. This bikes should also not have been able to calc speed from watts

cagnulein:
Thanks!

Issue #2630, opened 2024-10-03
Speed Adjustment Issue on FS 465D50 Walking Pad


travsosborne:
Description: I am experiencing an issue with my FS 465D50 Walking Pad. When I manually adjust the speed using the + buttons, the speed changes for approximately 3 seconds before reverting back to 1 km/h. This issue also occurs when using the preset speed buttons. However, when I manually adjust the speed directly on the walking pad, the speed adjustment is reflected correctly in the QZ app.

Steps to Reproduce:

Connect the FS 465D50 Walking Pad to the QZ app on an iPad Air 3rd Gen.
Open the QZ app (Version 2.16.70).
Manually adjust the speed using the + buttons on the walking pad.
Observe the speed change for approximately 3 seconds before it reverts to 1 km/h.
Repeat the process using the preset speed buttons.
Expected Behavior: The speed should remain at the manually adjusted value and not revert to 1 km/h.

Actual Behavior: The speed reverts to 1 km/h after approximately 3 seconds of adjustment using the + buttons or preset speed buttons.

Sorry if I've gone into too much information, i do work with developers all day and know what they expect for Bug reports 
Device Information:

Walking Pad Model: FS 465D50
QZ App Version: 2.16.70
Device: iPad Air 3rd Gen
Additional Information:

When manually adjusting the speed directly on the walking pad, the speed adjustment is correctly reflected in the QZ app.
Attachments:

Debug log (attached)
[debug-Thu_Oct_3_16_47_23_2024.log](https://github.com/user-attachments/files/17240471/debug-Thu_Oct_3_16_47_23_2024.log)

cagnulein:
hi @travsosborne i checked the log but the speed was always 0 and i don't see any events on qz to change the speed. Can you create debug log showing the issue?

travsosborne:
[debug-Thu_Oct_3_17_10_01_2024.log](https://github.com/user-attachments/files/17240614/debug-Thu_Oct_3_17_10_01_2024.log)
 take two 
I ran manually adjust and with presents

cagnulein:
i did a quick patch @travsosborne 

Do you have by chance a android device? if so you can test it right now, otherwise i have to create a new ios build. Let me know

https://github.com/cagnulein/qdomyos-zwift/actions/runs/11157639734

[...]

PR #2631, opened 2024-10-03
Speed Adjustment Issue on FS 465D50 Walking Pad (Issue #2630)

cagnulein:


cagnulein:
Speed Adjustment Issue on FS 465D50 Walking Pad (Issue #2630)

Issue #1705, opened 2023-10-14
Adidas Treadmill


cagnulein:
![adidas1340_nfrconnect-2](https://github.com/cagnulein/qdomyos-zwift/assets/48216/8d5a051f-e2fb-4af3-950e-84a92c90d68f)

maba80:
# Steps Taken in Kinomap app
Started Video Simulation where the incline increased and decreased by the track profile.

# Bluetooth Debug Logs
[btsnoop_hci.log](https://github.com/cagnulein/qdomyos-zwift/files/12903092/btsnoop_hci.log)

cagnulein:
great I will add this in the next days! @maba80

maba80:
#  Steps taken in Kinomap app
free ride with max elevation

1. connected treadmill via Kinomap on Samsung tablet 
2. Started free ride
3. Started via play button on treadmill
4. Raised speed via treadmill from 1 to 2 kph 
5. Increased incline from 0 to 15 every 2 seconds step by step
6. Decreased incline from 15 to 0 every 2 seconds step by step
7. Paused workout on Kinomap 
8. Stopped treadmill via stop button


I hope this is a better log for analyzing the behavior via Bluetooth on kinomap




# Bluetooth Debug Logs 2
[btsnoop_hci-2.log](https://github.com/cagnulein/qdomyos-zwift/files/12903430/btsnoop_hci-2.log)

cagnulein:
@maba80 good news! I finished the first raw implementation!
I will send you the new beta later today
Remember to put the FTMS treadmill setting to disable and press ok when you will receive it
Let me know!

cagnulein:
hi @maba80 did you try the beta?

maba80:
I've just downloaded it and will try it in the next hour. I will let you know hier it works!

maba80:
Here are my results:
Inclination works perfectly!
Speed adjustments doesn't work anymore.

cagnulein:
@maba80 yes because this protocol doesn't support speed changing from bluetooth. Could you change the speed on kinomap?
Just to give you some more informations:
your treadmill supports 2 protocols: a custom one and a FTMS one. Zwift use the FTMS (and qz also when you select the ftms treadmills setting), Kinomap instead uses the custom one.

maba80:
Thank you for the information.
From my side I'm happy with Your Implementation 😃
Many thanks for your Support!

cagnulein:
[...]

PR #2705, opened 2024-10-28
Adidas Treadmill #1705 on android

cagnulein:


cagnulein:
https://github.com/cagnulein/qdomyos-zwift/actions/runs/11549624991

Issue #2719, opened 2024-10-31
[REQ] Improvements made for Proform 705 CST support


Novamusix:
I've been using QZ on both Android and a Raspberry Pi to control the Proform 705 CST treadmill. You kindly updated QZ to accommodate the latest firmware update https://github.com/cagnulein/qdomyos-zwift/issues/2430

One major issue was that I needed to manually start the treadmill before connecting QZ; otherwise, the treadmill would lock out and not respond to the start button on the console or set speed requests from apps.

After lots of trial and error and monitoring bluetooth sequences sent from iFit app, I've made some updates to the `proformtreadmill.cpp` source code which looks to help address this:

- **App-Controlled Start:** If the treadmill is not running and ready to start, when an app sends a request to set the speed, an initiate start command sequence is sent. This defaults to 2 kph with an incline of 1, followed by setting the requested speed and incline. This works fantastically with Runna. `(Line changes on 338-339, 382, 838-930)`
- **Console-Controlled Start:** If the treadmill is not running and ready to start, pressing the Start button on the console will trigger QZ to send an initiate start command sequence of 2 kph with an incline of 1. This is perfect for other family members who are not using an app.` (Line changes on 2159-2169)`

It may be that this will work with other Proform models with updated firmware, but I cannot check this.

I've attached my updates using the latest `proformtreadmill.cpp` , 
[proformtreadmill.zip](https://github.com/user-attachments/files/17591802/proformtreadmill.zip)
which successfully compiled and is running on the RPi. Unfortunately I have no experience with pull requests so I hope attaching is ok.

cagnulein:
Thank you I will check the code and I will do a pull request for you

cagnulein:
i created it and i added some comments https://github.com/cagnulein/qdomyos-zwift/pull/2723 
Let me know

PR #2723, opened 2024-11-01
Improvements made for Proform 705 CST support (Issue #2719)

cagnulein:


cagnulein:
@Novamusix

src/devices/proformtreadmill/proformtreadmill.cpp
```diff
@@ -378,7 +378,7 @@ void proformtreadmill::update() {
                 writeCharacteristic(noOpData4, sizeof(noOpData4), QStringLiteral("noOp"), true);
                 break;
             case 4:
-                writeCharacteristic(noOpData2, sizeof(noOpData2), QStringLiteral("noOp"));
```
cagnulein, line 381: @Novamusix why change this if you created a new profile for your treadmill?

src/devices/proformtreadmill/proformtreadmill.cpp
```diff
@@ -2062,6 +2155,18 @@ void proformtreadmill::characteristicChanged(const QLowEnergyCharacteristic &cha
 
     lastPacket = newValue;
 
+    if (proform_treadmill_705_cst && currentSpeed().value() == 0)
+    {
+        QString hexString = newValue.toHex(' ');
+        if (hexString.startsWith("fe 02 25 04 00 00") && newValue.at(17) == 0x01)
+        {
+            emit debug(QStringLiteral("start button pressed on console!"));
+            Distance = 0;
```
cagnulein, line 2164: @Novamusix i don't like that Distance = 0, the distance is resetted on qz when you press the commands on the qz top bar.

src/devices/proformtreadmill/proformtreadmill.cpp
Novamusix: You are correct. This was a left over from my initial investigation before splitting the profile. I noticed that noOpData5 was not being referenced where noOpData2 was referenced in both case 2 and 4. Unlike other profiles such as norditrack_s25i which referenced all the noOpData fields in order. This change can be ignored.

src/devices/proformtreadmill/proformtreadmill.cpp
Novamusix: When running via an Android/iPhone, this may be the case. However, when running headless, the following scenario is encountered:

**QZ (RPi) connected to treadmill:**
- Start a new session via a connected app (e.g., Runna).
- Treadmill metrics progress on the console.
**Pause:**
- Press stop on the console once to pause the session
- Treadmill speed is zero.
- Press start again to resume.
- No start Bluetooth sequence is sent.
**Finish:**
- Press stop once to pause and stop again to finish the session.
- Pressing start will not do anything
- No start Bluetooth sequence is sent
**Reset:**
- Press stop once to pause, a second time to finish, and a third time to reset the treadmill session. 
- The console now displays zero values. 
- Pressing start will now send a Bluetooth sequence.
- However, because the distance is calculated in QZ and not retrieved from the treadmill, it will keep adding to the previous sessions distance.

Knowing that the start Bluetooth sequence is only sent when first powered on or when having been reset with the console showing zero values, meant resetting the distance value ensured it matches what it is shown on the treadmill.

And that when sending a speed request from an app, we can initiate start sequence which is needed to unlock the treadmill. The risk being without checking distance is zero the initiate sequence could be sent when paused.

There may be better ways of handling this, but this approach seem to work for the above scenarios.

cagnulein: ok it makes sense. i conditioned this only if the homeform is not available, this means a headless session. Could you try this branch and check if it's still working fine on your end?

Novamusix:
I downloaded the updated proformtreadmill.cpp and compiled successfully. Qz is running headless and working as expected. Thanks!

cagnulein:
Hah i just saw another case for the distance in the start request. I will
review it tomorrow


Roberto Viola
http://robertoviola.cloud



Il giorno lun 4 nov 2024 alle 18:17 cagnulein ***@***.***> ha
scritto:

src/devices/proformtreadmill/proformtreadmill.cpp
```diff
@@ -834,6 +835,99 @@ void proformtreadmill::update() {
...
+                        emit debug(QStringLiteral("writing incline = ") + QString::number(requestInclination));
+                        forceIncline(requestInclination);
+                    }
+                    requestInclination = -100;
+                }
+
+//              Request start does not seem to get initiated automatically.
+//              Instead, trigger it when the treadmill is stopped and a new speed is set via the application.
+//              Note: Setting a speed directly on the treadmill will not work until the initiate commands are sent.
+                if (Distance.value() == 0 && currentSpeed().value() == 0 && requestSpeed > 0) {
```
cagnulein, line 879: @Novamusix why also adding here distance.value() == 0 ?

it couldn't be useful without distance == 0 so you can resume it after a pause for example?

Novamusix: To validate my assumption, I compiled a version that removes distance.value() == 0.

**The following behaviour is observed:**
1. Start a treadmill session, set the inclination to 5, and set the speed to 6 km/h.
2. Pressing the Stop button on the console once will pause the treadmill session.
3. On a connected app, send a new speed request.
4. QZ will detect the current treadmill speed as zero and the requested speed as greater than zero.
5. QZ will trigger send the initiate command, which resets the treadmill timer, sets the inclination to 1, and sets the requested speed.

Previously with the check in place, steps 4 and 5 would be ignored, requiring me to intentionally press the start button to resume the session. With out having a check in place, this could potentially be dangerous if pausing to adjust clothing or stepping half off etc, only for the treadmill to start up again uncontrolled.

Ideally, we would monitor the stop button to determine the treadmill status, but I didn't have any luck with this. It's something I can explore further when I can dedicate some quality time, probably next month.

cagnulein: ok thanks, I understand the safety reason. Let's leave like this! Thanks for the tests

cagnulein:
hi @Bassai-Sho just a question: are you able to stop the treadmill from QZ directly with this code?

Bassai-Sho:
Hi @cagnulein ,

No unfortunately not.

I carried out test using android app v2.18.4, the QML implementation using Rpi02w keeps dropping the Bluetooth connection so could not check this as it is not powerful enough.

- I can begin the treadmill session by pressing the console start button, but not the start button on the QZ app or + on speed tile
- I can stop the treadmill by pressing the console stop button, but not the stop button on the QZ app.
- Pressing stop button on QZ, takes me to the charts screen but the treadmill continues to run

Hope this helps

cagnulein:
Yes you have to use use a raspberry pi 4 for also doing the virtual bridge
to zwift

Ok for the stop, thanks

Roberto Viola
http://robertoviola.cloud



Il giorno gio 14 nov 2024 alle 14:05 Adam Sharpe ***@***.***>
ha scritto:

PR #2757, opened 2024-11-09
Zwift hub gear custom

cagnulein:


GraVvVel:
@cagnulein So it's a bit of a mixed bag I must say. I did two tests. First log is from mywhoosh and it works in the beginning, but then stops working after a very short time. The initial gear shifts feel good.

In the second log I'm not connected to an app at first and shifts work great and seem to be correct - then I connect to mywhoosh and it stops working after a couple of shifts.

One idea that I have: Are you using the Zwift BLE Protocol consistently - i.e. also for the gradient change (see proto file under SimulationParameters)? Because maybe if you mix standard BLE and Zwift BLE it stops working - so e.g. on the first gradient change via standard BLE gear changes stop working... Just an idea.

[logs.zip](https://github.com/user-attachments/files/17688449/logs.zip)

cagnulein:
Yes it's absolutely the issue, I'm using mixed protocols right now. So it's
for sure an easy fix!
I will build a new one ASAP!

Roberto Viola
http://robertoviola.cloud



Il giorno sab 9 nov 2024 alle 16:34 GraVvVel ***@***.***> ha
scritto:

cagnulein:
sent a new one!

Roberto Viola
http://robertoviola.cloud



Il giorno sab 9 nov 2024 alle ore 17:01 cagnulein ***@***.***> ha
scritto:

GraVvVel:
@cagnulein Unfortunately same behavior (I did  the same test as above with two log files) 
[newlogs.zip](https://github.com/user-attachments/files/17689271/newlogs.zip)

GraVvVel:
Are you sending other commands apart from inclination? Maybe even the reading part needs to be changed - maybe ask Makinolo if he switched all communication (including power readings etc.) to Zwift BLE Protocol?

cagnulein:
Yes there are some errors, I left by mistake some commands. I will remove them in the next build (probably Monday)

Thanks for trying

cagnulein:
i checked the log and from what you said it seems the problem is the new frame about the inclination, maybe is not a valid one?

@Makinolo a frame like this for sending a target inclination to the device is ok?

04 22 0b 08 00 10 14 18 ec 27 20 90 03

In case if not, could you please show a right frame just to understand the error? I have the flu so I could did stupid mistakes :D 

Thanks!

GraVvVel:
@cagnulein I think for test purposes you should leave all of the wind etc. fields out as they are optional and only send inclination. Down the road it would be best if you used generated protobuf classes as that makes it easier for you to serialize correctly and you have the .proto from Makinolo which serves as input.

I think the command should just contain -> HubCommand -> SimulationParam (4: ) -> InclineX100 (2: ) -> 800 (for 8 percent).

Using something like Protobufpal.com with the .proto file and prefixing 0x04 I get: 0x04220310c00c (for 8%)

I whish you a speedy recovery!

cagnulein:
I did in this way because I used a buffer sniffed from a zwift hub so I was sure about the formatting 

About the protobuf yes it would be the right way to do, but if I have only to send a value it's easier bevause otherwise I have to use 4 different source code for all the platforms (ios, android, pc, Linux...) and it's a pain.

I will try to use only the inclination as you suggested, let's see if it's better.
Thanks

GraVvVel:
Ok understood! Currently there are two problems though: If you encode the entire protobuf it would need to be '18 ec 4f' instead of '18 ec 27' for CWa. And the actual inclination value seems wrong. e.g. in your example above you have '14' as the actual inclination value. For 8% it should be 'c0 0c', for 10% 'd0 0f'. I don't fully understand your conversion logic in the code, but it needs to be a signed int - maybe you are using unsigned like for the gearing?

cagnulein:
Yes it could be for sure the issue. Thanks for pointing me out on this!

Roberto Viola
http://robertoviola.cloud



Il giorno dom 10 nov 2024 alle 11:06 GraVvVel ***@***.***> ha
scritto:

Makinolo:
Protocol buffers uses varints not regular ints. That's why you can't just write the Hex value to the data frame.  Doing the conversion is not extremely difficult, but it's much easier if the protobuf library handles it. I was doing it manually at first because I wanted to avoid having another dependency in the project, but I ended up conceding.
You have the encoding explanation at the beginning of this page https://protobuf.dev/programming-guides/encoding/
The fact that the control messages are nested doesn't help either. It's really messy doing it manually.
https://www.protobufpal.com/  and https://protobuf-decoder.netlify.app/ are very helpful tools to help you verify the correctness of your encoded messages

cagnulein:
thanks @Makinolo I'm following your advice :)
@GraVvVel i'm sending you a new one with the protobuf implementation, let me know! i did this only on ios, but then i can add this on android too for the other guys.
Thanks

GraVvVel:
@cagnulein That was quick - I did a quick test in Rouvy. Gear shifting somehow did not work at all (both just in QZ without Rouvy or with Rouvy), but inclination changes were fine - not sure if I did the test correctly though. I had FTMS active  and otherwise no other changes. 

[debug-Mo__Nov__11_10_19_08_2024.log](https://github.com/user-attachments/files/17700209/debug-Mo__Nov__11_10_19_08_2024.log)

cagnulein:
got the issue! let me fix it! i will use also the protobuf for the gears one too. I'm sending you a new one shortly

cagnulein:
@GraVvVel sent! i hope it will be the good one!

GraVvVel:
Here's the log: 
[debug-Mo__Nov__11_12_42_57_2024.log](https://github.com/user-attachments/files/17701984/debug-Mo__Nov__11_12_42_57_2024.log)

Again first without connection to Rouvy and then with connection. It's hard  to say, but I think the gears worked initially but stopped working before connecting to Rouvy. Or they might be working, but with very little change.

cagnulein:
Inclination is still working OK after adding the protobuf file in the previous build?

GraVvVel:
Yes -  inclination is working

cagnulein:
I'm still seeing the same issue as before, so there is hope. That's strange bevause I did a static test before sending. Let me check again

cagnulein:
i guess i got it, i'm sending you another one!
thanks

Roberto Viola
http://robertoviola.cloud



Il giorno lun 11 nov 2024 alle ore 12:59 GraVvVel ***@***.***>
ha scritto:

cagnulein:
Sent a new one @GraVvVel

GraVvVel:
Thanks! I will test later today

GraVvVel:
@cagnulein Ok - so: Great progress! Shifting and inclination are working very quick and reliably!! Good job. Attached are logs of two short rides in Rouvy.

One thing that did feel a bit strange is when shifting up (i.e. harder) the gearing shortly gets very hard before it adjusts to the target gear - I never noticed that in Zwift, but I have not used gearing a lot in Zwift, so I will do a short test later to determine how it behaves there.

And it's currently possible to go below 1 and above the maximum gear number.

Apart from that I can't really test our physical crankset/cog adjustment formula, because my gearing (40/14) is too close to the Zwift Ride and I can't change it. Someone else would have to provide feedback on that.

But great stuff! A big step forward.

cagnulein:
> @cagnulein Ok - so: Great progress! Shifting and inclination are working very quick and reliably!! Good job. Attached are logs of two short rides in Rouvy.
> 
> 

Amazing!

> 
> One thing that did feel a bit strange is when shifting up (i.e. harder) the gearing shortly gets very hard before it adjusts to the target gear - I never noticed that in Zwift, but I have not used gearing a lot in Zwift, so I will do a short test later to determine how it behaves there.


Ok let me know!
> 
> 
> 
> And it's currently possible to go below 1 and above the maximum gear number.
> 

Ops i will fix it tomorrow!

> 
> 
> Apart from that I can't really test our physical crankset/cog adjustment formula, because my gearing (40/14) is too close to the Zwift Ride and I can't change it. Someone else would have to provide feedback on that.


I will invite here some guys to try this!
> 
> 
> 
> But great stuff! A big step forward.

Thanks to you and @Makinolo!
> 
> [logs (1).zip](https://github.com/user-attachments/files/17707758/logs.1.zip)
> 
> 
> 
>

GraVvVel:
Just checked in Zwift - it also has the behavior of getting harder for 1 - 2s and then settling into the gear. So is consistent and probably the intended behavior (it seemed stronger in Rouvy, but might be imagination and nothing to worry about for now) 👍

cagnulein:
Ok perfect, tomorrow I will also implement erg mode and gears limits and I will port it too android too in order to merge it!

Makinolo:
> One thing that did feel a bit strange is when shifting up (i.e. harder) the gearing shortly gets very hard before it adjusts to the target gear - I never noticed that in Zwift, but I have not used gearing a lot in Zwift, so I will do a short test later to determine how it behaves there.

This is the expected behavior. It works the same in Zwift and in GT Bike V. It tries to emulate the real world surge in power when you gear up and then progressively decreases when you get momentum

GraVvVel:
@cagnulein I don't know if you are doing it already, but maybe then also implement sending rider weight and bike weight in the simulation params via Zwift BLE

cagnulein:
nope, i'm not doing it, what do you like to add them as well? there is already a bike weight setting in qz and of course a player weight so i can use them to send them to the trainer. i don't see which effect could the bike weight could have on the trainer

cagnulein:
i would like to stabilize this also for android, and then i can also add these if you agree

cagnulein:
@GraVvVel i sent a version with the gear < 0 and > of the max should be fixed. I also merged the code with the master so i can deploy this version also to other beta testers.
Let me know!

GraVvVel:
Just tested it on a 45 min ride on Rouvy and it's working great - no issues, super quick shifts. But I must say that the circumference feature is also very good on the Kickr v6. It's great that you have the option to use either and both use the custom gearing table!

cagnulein:
i'm merging this and then i will add the weight on another PR. thanks

VictorSharov:
In my case this gear changing method works terrible with Wahoo Kick Core. When I'm trying to change gear up, for example, from 4 (around 60w at 80 cadence) to 5 (80w at 80 cadence) it generate huge spike for 5 seconds at 170w. So it's almost impossible to change gears. Previous (release) version of the app works good.
Also this problem not happens when changing gears natively in Zwift, it works smooth.

cagnulein:
@GraVvVel @Makinolo Do you have any idea about the behavior of @VictorSharov ?

If you want I can create a setting to restore the previous behavior based on inclination only but I would like to understand this since this method should work better than the previous one

cagnulein:
@GraVvVel I don't understand why but it seems that this only works for you. Did you do something on the bike to make it work? It's really a mistery without having the trainer unfortunately

cagnulein:
I guess I understood the issue: i was thinking that all the devices that has the custom bluetooth service 000001-... would be a zwift trainer with this protocol, BUT instead, the wahoo ones, has this service but they are talking a different protocol over this service. So I simply hanlded them using the wahoo standard protocol.
Everything should be fine now

PR #2794, opened 2024-11-18
Fix port on csafe rower

marcelrv:
Fixes 2 issues in the csafe rower..  

* the deviceFilename variable is not updated, hence a blank port is being opened
* metatype issue: because of void onStatus(uint16_t status);
```
(Make sure 'uint16_t' is registered using qRegisterMetaType().)
QObject::connect: Cannot queue arguments of type 'uint16_t'
(Make sure 'uint16_t' is registered using qRegisterMetaType().)
```

### DIFF
--- a/src/devices/csaferower/csaferower.cpp
+++ b/src/devices/csaferower/csaferower.cpp
@@ -90,7 +90,7 @@ void csaferower::onDistance(double distance) {
 
 }
 
-void csaferower::onStatus(uint16_t status) {
+void csaferower::onStatus(char status) {
     qDebug() << "Current Status received:" << status;
 }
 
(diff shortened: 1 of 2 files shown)


src/devices/csaferower/csaferower.cpp
```diff
@@ -90,16 +90,15 @@ void csaferower::onDistance(double distance) {
...
     qDebug() << "Current Status received:" << status;
 }
 
 csaferowerThread::csaferowerThread() {}
 
 void csaferowerThread::run() {
     QSettings settings;
-    /*devicePort =
-        settings.value(QZSettings::computrainer_serialport, QZSettings::default_computrainer_serialport).toString();*/
+    deviceFilename = settings.value(QZSettings::csafe_rower, QZSettings::default_csafe_rower).toString();
```
cagnulein, line 101: @marcelrv on android this is not needed. i guess you need to add a #ifdef here for linux only

PR #2811, opened 2024-11-22
Move csafe lib to separate folder and add commands

marcelrv:
In preparation to make a clean PR for the LifeFitness CSAFE elliptical. 

Changes:
* This PR moves the csafe lib to a separate folder (serveral csafe related files will be added later to the same folder)
* Adds few csafe commands
* Makes the CSAFE non-standard wrapping optional
* Apply the wrapping to the csaferower


cagnulein:
thanks!

PR #2863, opened 2024-12-03
Add Life Fitness 95 CSAFE Ellipical

marcelrv:
This is implementation of the CSAFE protocol for the  Life Fitness 95  Ellipical

It separates some of the device specific portion of the code from the more generic csafe code.
Hope this becomes useful when I get my treadmill working again. (as I expect it should also support CSAFE)
It also allows to connect the device to the serial over network instead of a physical serial port.


src/settings.qml
```diff
@@ -1005,6 +1005,10 @@ import QtQuick.Dialogs 1.0
             // from version 2.18.3
             property bool proform_performance_400i: false
 
+            // from version tbd
+            property string csafe_elliptical_port: ""
```
cagnulein, line 1009: they must be at the bottom of the others, otherwise the settings file will be corrupted @marcelrv

cagnulein (changes requested):

cagnulein:
wow thanks for your incredible work! i just leave you a note in the settings.qml before merging it!

src/settings.qml
marcelrv: ah.. yes, I had merge conflict each time, so I left it a bit above so not impacted by each version you made...
fixed now

PR #2868, opened 2024-12-03
Added files for building and running in Docker with GUI

nix155:
Added files for building and running in Docker with GUI /docker/linux_gui_vnc


cagnulein:
thanks!

PR #2879, opened 2024-12-05
Added build with qtwebglplugin

nix155:
Added build with qtwebglplugin. The web interface is available by default on port 8080. Using docker-compose, you can change the port or run without GUI and with custom parameters.


cagnulein:
amazing!

PR #2880, opened 2024-12-05
Powermeter pédale and ERG #2818

cagnulein:


cagnulein:
[powertraining.txt](https://github.com/user-attachments/files/18024368/powertraining.txt)

Issue #2954, opened 2024-12-29
[BUG] Raspberry PI - log-flooding / intermittent failures


magicdude4eva:
**Describe the bug**
This is a follow-up from https://github.com/cagnulein/qdomyos-zwift/discussions/2939 and the Pi installation as described in https://github.com/cagnulein/qdomyos-zwift/blob/master/docs/10_Installation.md

I noticed that the QZ service randomly stops this morning and I do not see anything in the log-files. Restarting the service seems to work again. When the treadmill is off, QZ in debug mode logs a lot of verbose data, which makes the journal unreadable and seems to log batches of messages like this:
[qz-log-debug-redundant.txt.zip](https://github.com/user-attachments/files/18268943/qz-log-debug-redundant.txt.zip)

For context: I automatically restart QZ via crontab at 5am in the morning: `0 5 * * * systemctl restart qz`. This morning I did a treadmill run at 7:46 and Zwift recorded fine via QZ. At around 9:30 I wanted to try something else and noticed that Zwift did not show QZ. Checking the log via `journalctl -u qz --since "2024-12-29 07:45:00" --until "2024-12-29 09:30:00"` did not show anything meaningful other than flooded logs. Restarting the service via `systemctl restart qz` made the threadmill discoverable again.

How would I be able to identify the cause of this "disconnect", especially since QZ worked some 2 hours before?

**To Reproduce**
1. Install the service as described in the Wiki
2. Run the service and then disconnect the treadmill
3. Observe the flooding of logs

**Expected behavior**
If a previously connected device is turned off, QZ should not flood debug logs or provide an alternative info/error/warn logging mechanism to make it easier to find issues such as service failures.

**Screenshots**
The service is started with logging enabled:
<img width="1914" alt="image" src="https://github.com/user-attachments/assets/d9302148-fede-4813-9f11-8f5e687bc61e" />

Service file:
<img width="519" alt="image" src="https://github.com/user-attachments/assets/17a4d295-6a8e-464a-b867-7342f0538b61" />

**Desktop (please complete the following information):**
 - Linux KICKR-RUN 6.6.67-v8+ 1833 SMP PREEMPT Fri Dec 20 14:14:15 GMT 2024 aarch64 GNU/Linux
 - Raspberry Pi 4 Model B Rev 1.5

**Smartphone (please complete the following information):**

**Append a debug log**
[...]

PR #2960, opened 2024-12-30
Included QZ service monitoring

magicdude4eva:
Added section to have your treadmill detected and the QZ service automatically started/stopped based on the treadmill state. This is based on the reported bug #2954


cagnulein:
Thanks!

Issue #3007, opened 2025-01-08
[REQ] Power transmission to garmin watch (via ANT+) when running on treadmill


dragilla:
I'm using QZ with a fitshow treadmill.
I want to register my training on my garmin watch (fenix8). I'm using ANT+ dongle connected to phone.
I want garmin to register speed, distance, hr and power. ATM I can only get speed, distance and HR (hr directly from HR sensor).
Power is not transmitted (it seems).

cagnulein:
Thanks I will check it ASAP

Roberto Viola
Software engineer and open source enthusiast
http://robertoviola.cloud



Il giorno mer 8 gen 2025 alle 14:46 dragilla ***@***.***> ha
scritto:

cagnulein:
@dragilla I checked: garmin doesn't accept power over ant+ or Bluetooth for running : for example, even with the stryd, you have to install its data field to see it on garmin, but you will not see in the garmin connect in the power field but just in a custom one


You can search also on internet about this, it's a voluntary garmin limitation to protect its ecosystem about running power 

Let me know your thoughts

dragilla:
Ok. I understand. I will ask for a different solution then. But I will create a separate feature request for that. Thanks!

sirfergy:
Could you emulate a running dynamics pod? That sends power across right?

cagnulein:
Yes it's already available using ant+ on android
Just enable the ant cadence setting in qz

Roberto Viola
Software engineer and open source enthusiast
http://robertoviola.cloud



Il giorno ven 17 gen 2025 alle 06:04 sirfergy ***@***.***> ha
scritto:

PR #3014, opened 2025-01-09
Power transmission to garmin watch (via ANT+) when running on treadmill (Issue #3007)

cagnulein:
…ll (Issue #3007)


cagnulein:
this is not possible because garmin doesn't handle

PR #3026, opened 2025-01-10
Add two console options to set horizon treadmill settings

sirfergy:
I was trying to use qdomyos-zwift with my Horizon 7.8 treadmill and Raspberry Pi and noticed that it wasn't relaying the data correctly, where on my Android it was. I learned that it was due to the two settings, `horizon_treadmill_7_8` and `horizon_treadmill_force_ftms` which can be set from the UX but not the console.

This PR makes them settable via the console!


cagnulein:
Thanks!

PR #3048, opened 2025-01-14
Test timeout exploration

drmason789:
This PR is host to code changes made to explore recent timeouts of the test suite.


drmason789:
It seems github can't pull from github right now. Will need to check up on this later.

cagnulein:
Thanks @drmason789

drmason789:
It is taking a long time between these:

```
qt.bluetooth: Using BlueZ kernel ATT interface
qt.bluetooth.bluez: Cannot open HCI socket
```

cagnulein:
Yeah I don't understand what changed

drmason789:
These tests run okay on my Linux development machine. Perhaps there's a system setting in the Github environment that is a timeout for trying to open the HCI socket that's changed.

This may take a while to solve. Perhaps you could comment them out in the Github Actions workflow until we figure this out.

cagnulein:
Ok I will disable them, thanks for your time!

drmason789:
@cagnulein I updated the linux environment to ubuntu 24.04 and the tests are running as they did before

drmason789:
@cagnulein An update of the linux environment to ubuntu 24.04 appears to have fixed it.

cagnulein:
> @cagnulein An update of the linux environment to ubuntu 24.04 appears to have fixed it.

Wow that's great! So we would had just to wait? I can simply restore back the tests?

drmason789:
Yes, restore the tests, but also update the Linux environment. See the code change in this PR.

cagnulein:
Thanks as usual!

PR #3084, opened 2025-01-24
 fixing-raspberry-64bit-segfault-build 

cagnulein:


cagnulein:
probably is about this https://github.com/docker/setup-buildx-action/issues/403

PR #3095, opened 2025-01-25
Fixing crash on windows about fit file writing

cagnulein:


cagnulein:
image
qfitsave

cagnulein:
fit files seems anyway unreadable but at least it doesn't crash anymore

Issue #3130, opened 2025-02-01
[BUG] Cannot set Virtufit Etappe 2 auto resistance


kminkov:
[debug-Sat_Feb_1_17_25_38_2025 (1).log.tar.gz](https://github.com/user-attachments/files/18629411/debug-Sat_Feb_1_17_25_38_2025.1.log.tar.gz)

**Describe the bug**
I am trying to run Virtufitt Etappe bike 2 with the Trainerday app in ERG mode but cannot make it to work properly. It connect properly but does not change resistance at all. The only time when resistance changes is when I change from ERG mode to slope mode in the app, than the resistance changed to a pretty hard gear once.

Target power in QZ app tiles is displayed properly. 

**To Reproduce**
1. Paired the Trainer Day app with QZ
2. Enabled the Virtufit Etappe setting under bike plus

**Desktop (please complete the following information):**
 - OS: Android
 - Version 13

**Smartphone (please complete the following information):**
 - OS: Android
 - Version 13

cagnulein:
Which version of qz do you have?

Roberto Viola
Software engineer and open source enthusiast
http://robertoviola.cloud



Il giorno sab 1 feb 2025 alle 16:52 kminkov ***@***.***> ha
scritto:

kminkov:
Version 2.16.66

> Which version of qz do you have?
> 
> Roberto Viola
> Software engineer and open

cagnulein:
Try to upgrade to the beta
Join it here https://play.google.com/apps/testing/org.cagnulen.qdomyoszwift
and update it from the store then

Let me know

Roberto Viola
Software engineer and open source enthusiast
http://robertoviola.cloud



Il giorno sab 1 feb 2025 alle 17:42 kminkov ***@***.***> ha
scritto:

kminkov:
Not any real difference. 

[debug-Sat_Feb_1_18_47_13_2025.log](https://github.com/user-attachments/files/18629572/debug-Sat_Feb_1_18_47_13_2025.log)

kminkov:
Actually managed to make it run. I think I have selected the bike in TMS Bike. After setting it to disabled it seems to work. But the power it is trying to hold is about 50w higher than the suggested. Do you have idea why is that and how to fix it? 
One more side question - Is it possible to use a separate power meter as the power source and still have the ERG mode working properly?

cagnulein:
For erg mode qz has to be calibrate : in order to do so spin for about 20
[...]

PR #3163, opened 2025-02-09
Improve Fitplus bike target resistance calculation #3130

kminkov:
Testing thoroughly with the Virtufit bike, I noticed that the ERG mode is working pretty well as for changing and holding the resistance. But the wattage is consistently 10 watts lower than the target. I think that it is related on how the target resistance is being calculated. This improvement should fix this I think.

### DIFF
--- a/src/devices/fitplusbike/fitplusbike.cpp
+++ b/src/devices/fitplusbike/fitplusbike.cpp
@@ -1133,10 +1133,19 @@ resistance_t fitplusbike::resistanceFromPowerRequest(uint16_t power) {
         return 1;
 
     for (resistance_t i = 1; i < max_resistance; i++) {
-        if (wattsFromResistance(i) <= power && wattsFromResistance(i + 1) >= power) {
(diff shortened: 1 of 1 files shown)


cagnulein:
ok if you are on android it will build automatically from here to test the patch. let me know

kminkov:
This change does not include the one from - https://github.com/cagnulein/qdomyos-zwift/issues/3130 . Should I add it this PR as well in order to be able to test properly ?

cagnulein:
Yes you can pull it from the master

kminkov:
Is there a way to reinstall the app without removing the already stored cadance/power/resistance table ?

cagnulein:
If you save the settings in qz and then load it back it will save them in
the Settings

Roberto Viola
http://robertoviola.cloud



Il giorno dom 9 feb 2025 alle 13:40 kminkov ***@***.***> ha
scritto:

cagnulein:
hi @kminkov any news?

cagnulein:
@kminkov any news?

PR #3165, opened 2025-02-10
skillbike: allow 3-digit bikes

liskin:
My TechnoGym Skillbike is called "BIKE 861" and QZ wouldn't detect it previously because it assumed all such bikes have 4-digit names.

The fix is to relax requirement and detect any /BIKE \d+/ as technogymBike.


cagnulein:
thanks, did you already try it?

liskin:
Yes with this fix it works fine with the TechnoGym Skillbike I have here

PR #3292, opened 2025-03-11
Cadence calculated from trainer speed

### LINKED ISSUE #3282: [REQ] Cadence Missing on FMTS on Decathlon D100
Unfortunatelly, by specs, the Decathlon D100 doesn't send cadence measurements. It looks like to be simply disabled by software. The D100 is a just a rebranded Magene T110, which does broadcast cadence data.

Maybe is possible that the data are still broadcasted. Maybe a check in the non-standard FMTS data transmitted.
_EDIT: Cadence is not broadcasted, at least not in the 0.106 firmware version_

Another solution for me could be to add a cadence sensor by a hall sensor connected to a GPIO on a Raspberry Pi.
_EDIT: Another solution could be to calculate cadence based on speed using virtual shifting. Coasting will give wrong cadence but could be acceptable._

**Additional context**
[btsnoop_hci_20250306_164047.zip](https://github.com/user-attachments/files/19117580/btsnoop_hci_20250306_164047.zip)
[debug-Thu_Mar_6_17_15_26_2025.log](https://github.com/user-attachments/files/19117581/debug-Thu_Mar_6_17_15_26_2025.log)

geel97:
This will give the possibility to activate the automatic calculation of the cadence by ratio from the speed transmitted by the trainer. Obviously it can only work if you use a fixed ratio with virtual shifting or ERG.
For now is still a draft. I'm working on a Decathlon D100 (it does not have the cadence sensor).
This will not close #3282, but will help.

### DIFF
--- a/src/devices/ftmsbike/ftmsbike.cpp
+++ b/src/devices/ftmsbike/ftmsbike.cpp
@@ -534,6 +534,14 @@ void ftmsbike::characteristicChanged(const QLowEnergyCharacteristic &characteris
             }
             index += 2;
             emit debug(QStringLiteral("Current Cadence: ") + QString::number(Cadence.value()));
+        } else if(VANRYSEL_HT && newValue.length() > 7) {
+            if (settings.value(QZSettings::cadence_sensor_name, QZSettings::default_cadence_sensor_name)
+                    .toString()
+                    .startsWith(QStringLiteral("Disabled"))) {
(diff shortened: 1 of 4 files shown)


geel97:
@cagnulein Can you enable Actions so Github can compile the APK file?

cagnulein:
I guess you have to use actions on your repo because I can't see any way to allow this.

Issue #3285, opened 2025-03-08
[BUG] Proform Performance 300i


octera:
**Describe the bug**
I have a performance 300i, with the 400i preset information are getting. displayed in app, but I'm not able to change speed,  inclination, start/pause/stop, ...
So look like I read value are OK, but write are not

**To Reproduce**
1. Go to app
2. Connnect
3. Click on start, stop, "+" on speed
4. Nothing happens
**Expected behavior**
Interaction with treadmill are OK

**Smartphone (please complete the following information):**
 - Device: Xiaomi 15T pro
 - OS: Android 15
 - Version 2.18.21

**Append a debug log**

[debug-Fri_Mar_7_22_54_24_2025.log](https://github.com/user-attachments/files/19142973/debug-Fri_Mar_7_22_54_24_2025.log)

cagnulein:
Are you able to collect a ifit logs using this guide
https://github.com/cagnulein/qdomyos-zwift/blob/master/docs/20_supported_devices_and_applications.md#android-device
?

Roberto Viola
Software engineer and open source enthusiast
http://robertoviola.cloud



Il giorno sab 8 mar 2025 alle 11:12 octera ***@***.***> ha
scritto:

octera:
I have used an other phone to make the BT capture (don't like enable dev mode on my main phone :) )
Here is the log.
What I have done : 
- Start a training : 2km/h, then around 3
- Go to 10km/h
- Set incline on 4%
- Set incline on 10% (max)
- Set incline on 0%. (min)
- Set speed to 16km/h. (max)
- Set speed to 3.1km/h
- Stop training

[btsnoop_hci.log](https://github.com/user-attachments/files/19144858/btsnoop_hci.log)

cagnulein:
Perfect, I will check it ASAP, I'm on Germany for 2 weeks with limited
access to my pc

I will let you know ASAP

Roberto Viola
Software engineer and open source enthusiast
http://robertoviola.cloud



Il giorno sab 8 mar 2025 alle 18:04 octera ***@***.***> ha
scritto:

octera:
I may try to make the fix with some help

I've tried to understand your code and the bluetooth protocol
From log I have extracted frame setting the 

- incline : 
0% : ff0d020402090409020102000000120000000000
4% : ff0d020402090409020102900100a30000000000
10% : ff0d020402090409020102e80300fd0000000000

- speed
3.1km/h : ff0d020402090409020101360100480000000000
[...]

PR #3398, opened 2025-04-18
feat: add Proform performance 300i

octera:
NB : not able to make start and stop action work, all the rest work well (data gathering, speed and inclination set, ...)

Linked to #3285


cagnulein:
perfect, thanks!

PR #3538, opened 2025-07-11
Hide QZ files from Android gallery and update UI flags

cagnulein:
Added createNoMediaFiles() to generate .nomedia files in the QZ directory and its subdirectories, preventing media scanning on Android. Updated CustomQtActivity to keep the status bar visible for notification access while hiding only the navigation bar.


cagnulein:
Screenshot_20250711_101708

cagnulein:
let's continue on #3540

PR #3620, opened 2025-08-16
Add cadence support from power sensor via crank torque data

cagnulein:
Introduces subscription to raw crank torque data to extract cadence from dedicated power sensors (ANT+ Page 16). Updates cadence reconciliation logic to prioritize power sensor cadence over speed/cadence sensor, and adds a getter for raw power sensor cadence.


cagnulein:
mail "QZ reading ANT+ Power Sensor"
14/08/2025

PR #3741, opened 2025-10-08
typo

giuseppe-macario:


cagnulein:
Actually tiles is right, isn't?

giuseppe-macario:
tiles -> plural
doesn't -> singular

cagnulein:
So the fix should be "don't" :)

giuseppe-macario:
If it's more than one tile, yes
(I didn't see them yet)

cagnulein:
Yes they are :)

Roberto Viola
http://robertoviola.cloud



Il giorno mer 8 ott 2025 alle 18:00 Giuseppe Macario <
***@***.***> ha scritto:

Issue #3765, opened 2025-10-16
[BUG] Gearing off


povlhp:
**Describe the bug**
Gearing is way off, feels like 3x larger steps between gears than it needs to be.

**To Reproduce**
I run with zwift Cog behind, and 22T chainring.  Kickr Core 1.
One change in gear does like 20% or more extra effort. Thus gear changes are too big.
I use onscreen gear buttons or Zift Play.

**Expected behavior**
With inifinity gears I expect one gear change to be around 10% change in watt output 

**Smartphone (please complete the following information):**
- iPhone 14 Pro max with iOS 26  
  
**Additional context**
Scaling the gears with gears gain 0.25 helped (was too much). Now set at 0.4 - Gives funny outputs on screen. Gear 3.8 etc. Either this should be fixed, or the calculation.

I suspect the cause is in  ftmsbike.cpp - see lines 356..358 below. (and just above original_ratio is user input values)
gear_value  is is calculated to current/original * 42/14. I would expect current_ratio/original to ALWAYS be 1.0.
Thus the calculation does not make much sense.   

I would expect it to be something dividing 42/10 and current_ratio * some constant. Now it is scaled by 42/14 only. If it is gear steps, It would likely have to be current_ration / (42/14), to get a scale factor of 2

            double current_ratio = ((double)g.crankset / (double)g.rearCog);
            
            uint32_t gear_value = static_cast<uint32_t>(10000.0 * (current_ratio/original_ratio) * (42.0/14.0));
            
            qDebug() << "zwift hub gear current ratio" << current_ratio << g.crankset << g.rearCog << "gear_value" << gear_value << "original_ratio" << original_ratio;

cagnulein:
Could you please attach a debug log with the current setup?
https://github.com/cagnulein/qdomyos-zwift/wiki/How-do-i-get-the-debug-log-in-case-something-doesn't-work%3F

Roberto Viola
Software engineer and open source enthusiast
http://robertoviola.cloud



Il giorno gio 16 ott 2025 alle 21:34 Povl H. Pedersen <
***@***.***> ha scritto:

povlhp:
[debug-fre__okt__17_10_50_58_2025.log](https://github.com/user-attachments/files/22967759/debug-fre__okt__17_10_50_58_2025.log)

Debug log included. With gears gain 0.4 - But I assume the correct value should be (22/14)/(42/14) = 0.5238

cagnulein:
[...]

PR #3770, opened 2025-10-17
[BUGFIX] Gearing off #3765

povlhp:
Behavior change for existing users.

Fixed gear ratio for non-standard ratios. Before it was reversed, so small front gear = larger value.  So a 22 chainring (like MTB) would have very steep gearing, like 20% per step.

This changes behavior for users who already used gear gain to compensate for non-standard ratio. They will likely have to reset gear gain to 1.0 and fine tune from there.

### DIFF
--- a/src/devices/ftmsbike/ftmsbike.cpp
+++ b/src/devices/ftmsbike/ftmsbike.cpp
@@ -354,8 +354,9 @@ void ftmsbike::update() {
             ((double)settings.value(QZSettings::gear_cog_size, QZSettings::default_gear_cog_size).toDouble());
             
             double current_ratio = ((double)g.crankset / (double)g.rearCog);
-            
-            uint32_t gear_value = static_cast<uint32_t>(10000.0 * (current_ratio/original_ratio) * (42.0/14.0));
+
+            // Scale gear ratio based on users chainring:cog ratio (original_ratio) relative to default Zwift 42 teet front, 14 rear.
+            uint32_t gear_value = static_cast<uint32_t>(10000.0 * 10000 * current_ratio * ( original_ratio / (42.0/14.0)));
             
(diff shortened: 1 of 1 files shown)


cagnulein:
Thanks it's building here https://github.com/cagnulein/qdomyos-zwift/actions/runs/18597169117

povlhp:
Submitted #3770

povlhp:
Rouvy on AppleTV will not see Windows version. 
Windows version crashes after less than 2 minutes. 

Android version would not read config uploaded/downloaded from iPhone. (Windows would). 

I can connect to Android version but disconnects after short time. Then I can not find it again and reconnect. 

So no test possible with these builds for me.

cagnulein:
> Rouvy on AppleTV will not see Windows version. 
> Windows version crashes after less than 2 minutes. 

Did you use the msvc 2022 build?

> 
> Android version would not read config uploaded/downloaded from iPhone. (Windows would). 

Share the config and which setting it doesn't load and I will check it thanks.

> 
> I can connect to Android version but disconnects after short time. Then I can not find it again and reconnect. 

Test the version even with android standalone, rouvy is not mandatory to check the gear ratio.

Thanks

povlhp:
Tested on Android. Found out I have 2x 10000 in there. So bad. Discard this. Will resubmit when fixed

cagnulein:
Ok let me know!

Roberto Viola
http://robertoviola.cloud



Il giorno dom 19 ott 2025 alle 10:06 Povl H. Pedersen <
***@***.***> ha scritto:

cagnulein:
Do you mean the line code is wrong?

Roberto Viola
http://robertoviola.cloud



Il giorno dom 19 ott 2025 alle 10:08 cagnulein ***@***.***> ha
scritto:

PR #2941, opened 2024-12-26
Raspberry kickr run

cagnulein:

PR #3780, opened 2025-10-19
feat/ant_linux_footpod

Bassai-Sho:
# Add ANT+ Virtual Footpod Broadcasting for Linux

## What This Feature Does

This PR adds ANT+ footpod broadcasting to QDomyos-Zwift on Linux. The application can now send treadmill speed and cadence data to Garmin watches and other ANT+ devices using an ANT+ footpod payload format compatible with Garmin devices.

**User benefit**: Connect your treadmill directly to your Garmin watch without needing a physical footpod sensor.

### Architecture Overview

The feature uses a **hybrid C++/Python design**:

- **C++ Layer** (`AntManager`, `AntWorker`): Manages the ANT+ worker thread and bridges data from the treadmill to Python
- **Python Layer** (`ant_broadcaster.py`): Handles low-level USB communication with the ANT+ dongle using the `openant` library
- **Integration**: Python script is embedded in the binary using `pybind11` (no external files needed)

### Key Design Decisions

**Non-blocking delayed startup**: The ANT+ feature starts 10 seconds after device connection using `QTimer::singleShot`. This delay ensures the treadmill's blocking initialization (`btinit()`) completes before ANT+ broadcasting begins.

**No core file changes**: All new code is isolated in the `antlinux` directory and `main.cpp`. Existing device drivers remain unchanged for stability.

**Thread safety**: ANT+ broadcasting runs on a dedicated `QThread` to avoid blocking the main application.

### Initialization Flow

```
1. Application starts → creates bluetooth object
2. Bluetooth discovers treadmill → creates device object
3. Device starts blocking initialization (btinit, ~8 seconds)
4. After connection, main.cpp arms 10-second timer
5. Timer fires → starts ANT+ worker thread
6. Worker thread broadcasts treadmill data via ANT+ dongle
```

### System Dependencies
```bash
sudo apt-get install -y libusb-1.0-0-dev libudev-dev python3-pip python3-venv
```

### Python Virtual Environment
Create a Python virtual environment (required for ANT+ libraries):
```bash
python3 -m venv ~/ant_venv
~/ant_venv/bin/python3 -m pip install pybind11 pyusb openant
```

**Note**: The application automatically finds this virtual environment at `~/ant_venv`.

### Hardware
- ANT+ USB dongle (tested with Garmin ANT+ USB-m, model 0fcf:1008/1009)
- Compatible ANT+ receiver (tested with Garmin Forerunner 245 Music)

## Usage

```bash
# Enable ANT+ footpod broadcasting
sudo ./qdomyos-zwift -no-gui -ant-footpod

# Use custom ANT+ device ID
sudo ./qdomyos-zwift -no-gui -ant-footpod -ant-device 12345

# Enable verbose ANT+ debug logging
sudo ./qdomyos-zwift -no-gui -ant-footpod -ant-verbose
```

### New Files

**Core ANT+ Implementation:**
- `src/devices/antlinux/AntManager.cpp/h` - Singleton managing ANT+ feature lifecycle and worker thread creation
- `src/devices/antlinux/AntWorker.cpp/h` - Worker thread managing embedded Python interpreter and treadmill data feeding
- `src/devices/antlinux/ant_broadcaster.py` - Python script implementing ANT+ footpod protocol using openant library
- `src/devices/antlinux/ant_test_broadcaster.py` - Standalone test utility for verifying ANT+ dongle connectivity independently

**Build System:**
- `src/devices/antlinux/py_to_header.py` - Converts Python scripts to C++ header string literals for embedding
- `src/devices/antlinux/antlinux.pri` - Qt project configuration defining build flags and Python integration settings
- `src/devices/antlinux/build_check_list.sh` - Validates development environment setup including Python venv and USB permissions

**Documentation:**
- `src/devices/antlinux/README.md` - Complete installation guide
### Modified Files
- `src/main.cpp` - Added ANT+ command-line parsing and delayed startup trigger logic
- `src/devices/bluetooth.cpp` - Added device connection signal emission for ANT+ initialization
- `src/devices/bluetoothdevice.h` - Extended header with ANT+ integration declarations

## Testing

Tested configuration:
- **Platform**: Raspberry Pi Zero 2 W (Raspberry Pi OS Bookworm 64-bit)
- **Hardware**: Proform treadmill + Garmin ANT+ USB-m dongle
- **Receiver**: Garmin Forerunner 245 Music

## Known Limitations

1. **Linux only** - Currently supported only on Linux systems
2. **Python dependency** - Requires Python virtual environment setup with openant library
3. **Startup delay** - ANT+ broadcasting begins ~2 seconds after treadmill is fully ready (10-second total delay accounts for the ~8-second `btinit()` blocking initialization)

## Backward Compatibility

This feature is **completely optional** and does not affect existing functionality:

- Works alongside all existing Bluetooth features
- Disabled automatically if dependencies are missing
- All existing command-line arguments unchanged
- No modifications to core device driver code

If the ANT+ feature is not used or dependencies are unavailable, the application behaves exactly as before.


cagnulein:
Wow it looks amazing, thanks! I suggest to enable ANT_LINUX_ENABLED automatically in the pro file when building for Linux. What do you think?

Bassai-Sho:
That's a great suggestion. I completely agree that making the feature as easy as possible to enable is the ideal goal.

My only concern with enabling ANT_LINUX_ENABLED automatically for all Linux builds is that the ANT+ feature has a hard dependency on a specific Python virtual environment (ant_venv) and several development packages (python3-dev, pybind11, etc.). 

This hybrid C++/Python approach was chosen to leverage the mature and well-tested openant library, as a full native C++ implementation would have been a massive undertaking.

Because of these dependencies, if we define ANT_LINUX_ENABLED by default, the build would unfortunately fail for any user who hasn't manually performed this specific Python setup. This could also break the standard Linux CI build, as the runner won't have these dependencies installed.

To solve this, the current approach in antlinux.pri acts as an "opt-in" feature detector. It automatically checks for the existence of the .ant_venv_path file, which our setup scripts create as a signal that the environment is ready.

- If the file exists, the feature is seamlessly enabled.
- If it doesn't exist, it's safely skipped.

What do you think of this approach? It seems to provide the best balance, it offers "just-works" automation for users who follow the setup, while guaranteeing that the build remains safe and clean for everyone else and for the CI.

cagnulein:
Yes I like it! Also maybe we can try to adapt the raspberry images in the
CI to have the full setup ready? What do you think?

Roberto Viola
http://robertoviola.cloud



Il giorno dom 19 ott 2025 alle 18:14 Adam Sharpe ***@***.***>
ha scritto:

Bassai-Sho:
That would be fantastic and a much better experience 🙂 , adding ANT+ to the CI builds would help ensure it's tested automatically and stays stable.

**CI Build - Steps the user would need to take**
When users download a CI-built binary with ANT+ included, they will need to set up their environment on their machine. They would follow these steps from the installation guide:
- Step 1: Install System Dependencies (libusb-1.0-0-dev, python3-venv, etc.). no need to clone
- Step 2: Create Python Virtual Environment (python3 -m venv ~/ant_venv and install packages)
- Step 3: Configure Build Environment (.ant_venv_path file)
- Step 4: Configure USB Permissions (udev rules and plugdev group)
- Steps 5-9: All remaining setup steps (verify setup, test, configure systemd, etc.)
The only real difference is users skip Step 7 (Build QZ) since they already have the compiled binary, they just run through the environment setup to ensure everything is ready on their machine.

**Python Version Match**
Users must ensure they have the same Python version installed on their machine as was used in the CI build. For example, if the CI built the binary with Python 3.11, users need to install Python 3.11 on their system. Using a different Python version will cause library loading errors.

**Testing**
We should test the CI build on a vanilla PI installation (without the python dependencies followed) just to ensure graceful failure, allowing the user to still use the build without ant+, rather than crashing.

**Documentation**
The documentation can be tweaked to factor this in to help guide the user, given they will still need to follow the manual setup steps.

cagnulein:
You can try to patch the github workflow as well. Chatgpt or claude should
do quite easily.
Let me know if you have any other questions

Roberto Viola
http://robertoviola.cloud



Il giorno dom 19 ott 2025 alle 19:05 Adam Sharpe ***@***.***>
ha scritto:

Bassai-Sho:
Hi Roberto,

I believe the ANT+ feature is now stable that I can start following the TESTPLAN.md. Before I do, I want to confirm the build plan with you.

For now, I have created two separate Docker builds for ANT+

- raspberry-pi-build-64bit-ant
- linux-x86-build-ant

This seemed like the safest approach. The original builds do not change, so there is zero risk for current users, and the new Python dependencies are kept separate.

My suggestion is to use these separate builds for testing and the first release. Later, once the feature is proven to be stable, we can make these the main builds for all Linux and Pi users.

Do you agree with this plan, or do you have a different preference?

Also, where is the best place for users to download these new builds? I could not find a "nightly builds" area. Right now, I can only find the download links deep inside the GitHub Actions workflow archives. Is there a better, more direct link?

Many Thanks,
Adam

cagnulein:
Wow that's great!
Yes i think it's a safe approach and I like it
Does the current github workflow in this branch builds the images? or it
builds the the docker images?

Roberto Viola

please support my time:
http://paypal.com/cagnulein

http://robertoviola.cloud
http://qzfitness.com/


Il giorno mar 28 ott 2025 alle ore 09:40 Adam Sharpe <
***@***.***> ha scritto:

Bassai-Sho:
Hi Roberto,

Great, Thanks for confirming.

The GitHub workflow builds the binary file (the artifact for users to run on their machine). These are found in the Archive section

image

It uses a Docker image as a temporary tool to compile the code. The Docker image is deleted after the build finishes.

I shall start following the test plan this week , using the builds created here, and will let you know when I have completed.

For reference I am using this feature on my treadmill, it has been working well, but I want to make sure we have covered everything before others start using and raising questions.

Many Thanks,
Adam

cagnulein:
That's great so we can add the 2 task on the github workflow instead of
changing the actual ones so we will have 2 new artifacts in the end. Does
it sound good to you?

Roberto Viola

please support my time:
http://paypal.com/cagnulein

http://robertoviola.cloud
http://qzfitness.com/


Il giorno mar 28 ott 2025 alle 10:04 Adam Sharpe ***@***.***>
ha scritto:

Bassai-Sho:
Hi Roberto,

Apologies for the delay, things got busy.

After extensive testing, it was discovered a serious conflict between the Python library (needed for ANT+) and the Qt GUI library.
 When the ANT+ feature is compiled in then:
- The headless (-no-gui) mode works.
- The GUI mode (-gui) the application fails to start. It crashes or freezes with 100% cpu.
- The problem appears to be a low-level library conflict that is very difficult to solve.

I spent many hours trying to solve with AppImage and trying to dynamically load the python library, but each time there were more race conditions being uncovered, resulting in 100% cpu with a frozen and a broken application.

Perhaps for now, ANT+ Footpod build is only available for the headless (-no-gui) mode and is unable to be ran in gui mode?
Or would this be a broken user experience that is difficult to support and so instead we continue in the background to see if there is a solution, understanding that this may take a while?

Many thanks
Adam

cagnulein:
even with the -qml or the webgl implementation has the same?

Bassai-Sho:
Hi Roberto,

The ANT+ feature is now working on both the Raspberry Pi (no screen) and the Desktop (Ubuntu 24.04). I am just fixing a few things in the setup-dashboard.sh script so the user has the tools to setup ANT+ on Linux. I will then submit the changes for you to check.

**The "Runna" App Fix**

I also fixed a problem so that the Runna app now works with QZ on Pi.
-     Someone on Facebook asked if they worked using two device
-     It might work on iPhones, but I can't test that.
-     On the Raspberry Pi, it was not working

**What was the problem?**
See 391db60a (Raspberry kickr run, #2941
In December 2024, some code was added to use a "Wahoo" style connection. This was causing the Runna app to crash. I thought it was Runna's fault, but I found a fix for 'virtualtreadmill.cpp'. When I told the code to use the new "standard" connection path (standard-ftms) based on Horizon 7.0 AT, it started working perfectly.

Right now, I have set the code to use the Wahoo path by default. If a user wants the Standard FTMS, they have to use a switch in the command line.

How would you like to handle this? Is this approach ok?

cagnulein:
Thanks @Bassai-Sho for your massive work!
I will start a review soon!

For the wahoo tread part: if you simply change the name to something different from the wahoo name, was it still crashing? was the runna app crashes also when you run qz from your mobile?

Bassai-Sho:
here is the response from Claude via Vscode

The root cause was that Runna detected the Wahoo proprietary GATT service (A026EE0E-...) advertised by the virtual treadmill and switched to the Wahoo BLE protocol instead of standard FTMS. The Wahoo protocol skips subscribing to the FTMS Control Point CCCD, which caused a 6-second timeout and immediate disconnect (Crash).

The fix was removing that Wahoo service from the BLE peripheral entirely when running in standard FTMS mode (via -no-wahoo-service). Without it, Runna falls back to standard FTMS, correctly subscribes to the Control Point, and the connection holds. Everything else — the device name, DIS strings, feature flags — was supporting work to make the device look like a known standard FTMS treadmill so Runna trusted it.

I did not check with two android devices as i only have the one, but I can check over the weekend (heading into work now)

cagnulein:
so i guess the best way to approach this is to create a new setting to enable/disable the wahoo tread (maybe called Runna or similar). What do you think? I would like to keep the ANT_ defines at less as possible

Bassai-Sho:
Agree, that was my thought , the guards were made to be easily removed, given this feels not just limited to PI. The setting would need to be in the QML gui, but not sure how to go about this, can certainly attempt, however this Runna issue seems more of a higher priority. I can raise this as a new branch to just get Runna working then look for some help to get the GUI settings made? Unless you are happy to keep as is and when I fix the ant+ setup-dashboard issue we do this as one update?

cagnulein:
if you want i can create the setting and the UI for you so you can link in the virtualtreadmill.cpp, what do you think?

Bassai-Sho:
That would be awesome, thank you 😊. Then I just point to that setting and remove the no-wahoo switch.

cagnulein:
done, it's under the treadmill options, enabled by default

Bassai-Sho:
Hi Roberto, so the Runna crashing was actually more simpler than first thought, it was resolved by adding the serviceFTMS and serviceWahoo signals to the reconnect(). Without the reconnect they were pointing to dead objects. No need for the new flags or additional workaround, which is better overall. Have tested and works.

There are some known issues but nothing that prevents the core functionality from working:
- Virtual Bluetooth devices (e.g. Just Fit) do not work on BlueZ 5.70+ / Ubuntu 24.04 — ANT+ footpod unaffected
- SD card protection (Overlay FS) is not currently working and needs investigation — do not rely on it
- ANT+ footpod flag will default to false  this needs to be enabled via QZ Runtime Flags and service regenerated
- Device online/offline status does not always reflect the true current state
- Guided setup wizard does not offer Smart Monitor — must be set up manually after wizard completes
- Occasional screen artefacts in the dashboard — display only, does not affect functionality

I can look at these later, but thought it is ready enough especially with the Runna fix. :+1:

cagnulein:
> Hi Roberto, so the Runna crashing was actually more simpler than first thought, it was resolved by adding the serviceFTMS and serviceWahoo signals to the reconnect(). Without the reconnect they were pointing to dead objects. No need for the new flags or additional workaround, which is better overall. Have tested and works.

That's an awesome news! Let me review it and I will merge it then! thanks!

cagnulein:
hi @Bassai-Sho i was reviewing it, were you able to test this also with zwift? it's still working? i was worried about https://github.com/cagnulein/qdomyos-zwift/blob/3e3149ce311edc51c4320d6f683f0bd91bf3f24f/src/virtualdevices/virtualtreadmill.cpp#L147-L155

Thanks!

Bassai-Sho:
HI @cagnulein,

I have not used Zwift for while now. however here is my reasoning and steps followed so far.

- I just did a test using Zwift on my Android phone, selecting "Kicker Run" for both [Run Speed] and [Controllable] tiles in Zwift.
- The assumption is, Incline will automatically adjust.
- I choose Hill + Tempo # 3, and 12 minutes into the run it was asking to change inclination to ~4%. The  treadmill did not automatically adjust the incline.
- The speed and incline from treadmill was succesfully getting pushed to Zwift
- The ANT+ metrics were succesfully being sent to the watch.

So your concern around incline might be valid , I will look to do the following additional steps

- Create a custom .TWO workout, which will quickly adjust the inclination
- Use two Android devices , Android 1 for  QZ and Android 2 for Zwift, to validate inclination adjusts automatically prior to this change.

My initial thinking is the code segment might require gating if Zwift is being used, still allowing 3rd party apps such as Runna to work. But I can experiement and let you know how I get on.

Bassai-Sho:
Hi @cagnulein,

Ok, the rasperry pi **worked**!!  The two Android setup also worked as a baseline validation.

I followed https://github.com/cagnulein/qdomyos-zwift/wiki/Can-Zwift-control-speed-and-or-inclination-of-my-treadmill%3F

I needed to have been in a **Free Run** in Zwift  (running on android). I selected "Mountain Mash" and travelled ~1km into the run. The incline on the treadmill started to self adjust. Seems that Workouts do not send gradient information due to ERG/pace mode, where Free runs use the actual terrian and gradient data.

I also learnt that custom run workouts using .ZWO files cannot include gradient/incline information so stopped that idea. https://www.zwiftworkout.com/

I attach the debug file as evidence
[debug-Thu_May_28_16_02_24_2026.log](https://github.com/user-attachments/files/28352716/debug-Thu_May_28_16_02_24_2026.log)

Hope this helps :+1:

Issue #2176, opened 2024-03-02
[BUG]App showing unexpected resistance changes when clicking the + or - in the resistance tile and then stopped connecting to the bike


Mrschley:
Qz app connected to echelon ex3 bike but when i clicked on the + or - in the resistance tile it would jump up or down to a different resistance that didn't make any sense.  For example I could be at a resitance of 20, click the + and it would change to 13.  I tried restarting the app and powering down the phone running Qz.  At that point it would no longer connect to the bike via bluetooth.  It would either find something else like our washer or dryer, or not find anything at all.  I also tried unplugging and replugging the bike.

Can you tell what is going on from the debug preports?  Not sure which ones to send so sending them all from that day.

[drive-download-20240302T202100Z-001.zip](https://github.com/cagnulein/qdomyos-zwift/files/14470678/drive-download-20240302T202100Z-001.zip)

cagnulein:
for sure your issue is about the gear setting under echelon settings marked
as experimental . disable it

Roberto Viola
Software engineer and open source enthusiast
http://robertoviola.cloud



Il giorno sab 2 mar 2024 alle 21:24 Mrschley ***@***.***> ha
scritto:

Mrschley:
so i can't use the gears tile?

cagnulein:
i'm not talking about the gear tile. i'm talking about this 

![image](https://github.com/cagnulein/qdomyos-zwift/assets/48216/b7cf2d38-8b3f-4a7d-84b8-23e69abb5160)

disable it

cagnulein:
also during a workout with peloton or zwift you should never use the plus and minus resistance tile. only the gear one

Mrschley:
Thanks

Mrschley:
I disabled it and restarted the app.  Still having issues connecting.  It
says it found the bike but it doesn't register any movement.  There is also
a "stroke rate" tile showing up so it seems to think the bike is a rower?

Mrschley:
Looks like i fixed it.  Somehow fake rower got turned on.  I can't imagine
why I would turn that on.  Maybe I accidentally changed it.

cagnulein:
Set your bike on the manual device setting in the advanced setting and
restart Qz. You are not connected to the bike

Roberto Viola
Software engineer and open source enthusiast
http://robertoviola.cloud



Il giorno dom 3 mar 2024 alle 16:04 Mrschley ***@***.***> ha
scritto:

[...]

PR #3932, opened 2025-11-30
Fix Echelon watt tables and gear feedback loop

colbond:
## Summary

- Make `wattTable` and `wattTable_mgarcea` **static const** to avoid repeated stack allocation (~5.8KB) on every `wattsFromResistance()` call. This also prevents potential issues with values drifting during long sessions.

- Fix gear feedback loop by **saving and restoring** `lastRawRequestedResistanceValue` around `setGears()` call. Previously the value was set to -1 but never restored, which could affect future resistance checks.

- Change `watts_of_level` pointer to `const` to match the now-const tables.

## Related Issues

- Discussion #483 (watt table / resistance falling off during long rides)
- Issue #2176 (unexpected resistance changes with gear settings)

## Testing

Tested for 1 hour on Echelon Connect Sport with Zwift on macOS.

### DIFF
--- a/src/devices/echelonconnectsport/echelonconnectsport.cpp
+++ b/src/devices/echelonconnectsport/echelonconnectsport.cpp
@@ -272,8 +272,10 @@ void echelonconnectsport::characteristicChanged(const QLowEnergyCharacteristic &
                 int8_t g = gears();
                 g += (res - qRound(Resistance.value()));
                 qDebug() << QStringLiteral("gears_from_bike APPLIED") << gears() << g;
-                lastRawRequestedResistanceValue = -1; // in order to avoid to change resistance with the setGears
+                resistance_t savedRawValue = lastRawRequestedResistanceValue;
+                lastRawRequestedResistanceValue = -1; // temporarily prevent setGears from re-applying resistance
                 setGears(g);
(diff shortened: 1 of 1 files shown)


cagnulein:
So changing the real resistance wheel on the bike now works better with this patch?

Does your bike have auto resistance? Or it has the wheel on the front?

colbond:
Bike is an EX-5s with auto resistance - wheel in the rear. I very occasionally had issues that seemed like a race condition when changing the virtual gears using the knob. If the resistance value was also dropping from Zwift, it could cause the bike resistance to drop to the minimum and stay there. Also sometimes the gears value would change when Zwift changed resistance but I didn't change it.  I think those are now fixed with the setGears change. 

The static const change on the watt tables ensures the lookup tables stay in fixed memory rather than being stack-allocated on every call.

cagnulein:
Yeah for sure the const thing it's perfect

Let me know how it goes with the gear thing in the next sessions so I can definitely merge it

Thanks!

cagnulein:
@colbond Any news?

colbond:
So far so good. I’ve done a number of tests and haven’t reproduced the race condition that was causing the resistance to drop to the minimum and stay there.

cagnulein:
Ok let's merge! Thanks!

Issue #4000, opened 2025-12-19
[BUG] Treadmill converting to KPH instead of recognizing MPH


tivo177:
**Describe the bug**
ZWO for example is supposed to be 6mph (which it is) but QZ makes the treadmill go 9.6mph.

**To Reproduce**
1. Connect QZ to treadmill (TM4800)
2. Open Train Program
3. Other Folders
4. Select ZWO
5. Start program

**Expected behavior**
QZ to control treadmill per workout plan in ZWO. For example, plan calls for 6mph, treadmill then goes 6mph

**Smartphone (please complete the following information):**
 - Device: iPhone 16
 - OS: iOS26.1

**Append a debug log**

 Follow this guide 

[0QZ-backup-Thu Dec 18 17_02_57 2025.zip](https://github.com/user-attachments/files/24247616/0QZ-backup-Thu.Dec.18.17_02_57.2025.zip)

 
Related to this reddit thread:
https://www.reddit.com/r/qdomyos_zwift/comments/1pph11m/customs_workouts/

cagnulein:
Hi thanks unfortunately this is a fit file but I need a debug log one.
Could you please upload it?
Thanks

Roberto Viola

I sometimes trade time with my daughter to help. If this was valuable,
please support my time:
https://buymeacoffee.com/cagnulein
https://www.patreon.com/cagnulein/join
https://paypal.me/cagnulein

Software engineer and open source enthusiast, QZ Creator
http://robertoviola.cloud
http://qzfitness.com/


Il giorno ven 19 dic 2025 alle 01:12 tivo177 ***@***.***> ha
scritto:

tivo177:
Not sure which one it is, sorry about that.

[debug-Thu_Dec_18_17_02_57_2025.log](https://github.com/user-attachments/files/24260555/debug-Thu_Dec_18_17_02_57_2025.log)
[debug-Thu_Dec_18_17_01_37_2025.log](https://github.com/user-attachments/files/24260556/debug-Thu_Dec_18_17_01_37_2025.log)

cagnulein:
Thanks could you please attach also the zwo file?

tivo177:
[vqwoxpmdior.zip](https://github.com/user-attachments/files/24260846/vqwoxpmdior.zip)

cagnulein:
This zwo uses the 1 mile total time setting under the training program options in qz to set the speed accordingnly. Are you sure that setting is correct for you?

[...]

PR #4002, opened 2025-12-19
Add FTMS treadmill support for TM4800- device

cagnulein:
- Added TM4800- device detection in bluetooth.cpp
- Added TM4800 boolean flag to horizontreadmill.h
- Added device name detection in horizontreadmill.cpp
- Implemented miles_unit special case handling in forceSpeed()
  When miles_unit setting is enabled, the TM4800 treadmill
  receives speed commands in miles instead of km/h


cagnulein:
Treadmill converting to KPH instead of recognizing MPH (Issue #4000)

PR #4009, opened 2025-12-20
Fix Echelon Connect Sport to use power pedal sensor value

cagnulein:
When power sensor/pedal is enabled, the Echelon Connect Sport was
incorrectly reading from the erg table instead of using the actual
power value from the power pedal sensor. This fix ensures that when
a power sensor is enabled, the watts() method returns the actual
power value from m_watt (populated by the power pedal) instead of
calculating from the erg table.

### DIFF
--- a/src/devices/echelonconnectsport/echelonconnectsport.cpp
+++ b/src/devices/echelonconnectsport/echelonconnectsport.cpp
@@ -209,6 +209,16 @@ resistance_t echelonconnectsport::pelotonToBikeResistance(int pelotonResistance)
 resistance_t echelonconnectsport::resistanceFromPowerRequest(uint16_t power) {
     qDebug() << QStringLiteral("resistanceFromPowerRequest") << Cadence.value();
 
+    QSettings settings;
+    bool powerSensorEnabled =
+        !settings.value(QZSettings::power_sensor_name, QZSettings::default_power_sensor_name)
+             .toString()
+             .startsWith(QStringLiteral("Disabled"));
+
+    if (powerSensorEnabled) {
+        return _ergTable.resistanceFromPowerRequest(power, Cadence.value(), maxResistance());
+    }
+
     if (Cadence.value() == 0)
         return 1;
 


retroromantic:
[debug-Sun_Mar_22_18_58_23_2026.log.zip](https://github.com/user-attachments/files/26249732/debug-Sun_Mar_22_18_58_23_2026.log.zip)
[debug-Tue_Mar_24_20_11_02_2026.log.zip](https://github.com/user-attachments/files/26249740/debug-Tue_Mar_24_20_11_02_2026.log.zip)
Hello

As discussed via email I'm attaching my most recent debug logs documenting a "bouncing" behavior in the auto resistance on my echelon ex5 using favero assioma duos. 

One update, I realized that I don't know the appropriate crank length to use for the ex5 in the favero app. I updated the length to 170mm as an experiment and it seems the wattage readings were more inline with the 'mgarcia' watt table, mitigating the issue somewhat. 

Please let me know if I should provide any additional email. 

Thanks

cagnulein:
Thanks I will check it ASAP

Roberto Viola

please support my time:
https://paypal.me/cagnulein

http://robertoviola.cloud
http://qzfitness.com/


Il giorno mer 25 mar 2026 alle 17:56 retroromantic ***@***.***>
ha scritto:

cagnulein:
@retroromantic done! https://github.com/cagnulein/qdomyos-zwift/actions/runs/23588340288

- Save the current settigns of qz from the settings page top icon on the right
- Move the settings file outside qz folder 
- remove the current qz version
- Download this version and install it.
- load the settings 
- try

What I did is adding a new way to qz to learn from your power sensor, so qz while you are riding will always learn the power curve from it and it will improve over time. 

--

retroromantic:
Hi i gave it a try this morning and the auto resistance was different for sure. It didnt seem to recognize any changes in wattage during spinups and the initial build up in class. However, as class went on it almost seemed like the app had "learned" and made improvements in the auto resistence process and did a better job recognizing my zones. I even tried to increase my cadence to bike above the prescribed zone and the app decreased the resistance as necessary. 

I am unsure about what happened in the first half of the ride but assume that the modiciations you made may learn more about my bike as i continue to ride. 

I attached my debug log here for your review. 
[debug-Thu_Mar_26_07_54_14_2026.log.zip](https://github.com/user-attachments/files/26286118/debug-Thu_Mar_26_07_54_14_2026.log.zip)

cagnulein:
@retroromantic Did you load the settings?

The reason why you are seeing this it's because the power table is empty so qz was still learning

In order to make it easier for qz do a free ride staying on each resistance for at least 30 seconds. This should create a perfect power table

Let me know!

retroromantic:
Ah I see, I did attempt to reload the settings as directed but the only thing that seemed to change was the default age.  I then input all my preferred settings manually.  I'll attempt to do the free ride as directed and then do a power zone and report back.

Thanks again.

retroromantic:
A quick update, i did the free ride and hopefully built a better power table.  Will do a pz ride in the next few days and report back. 

[debug-Fri_Mar_27_07_43_37_2026.log.zip](https://github.com/user-attachments/files/26312754/debug-Fri_Mar_27_07_43_37_2026.log.zip)

cagnulein:
Yes I see the table populated correctly

retroromantic:
Hello,

I rode two power zone classes and the auto resistance seemed to be much more rrsponsive and kept me in line with the call outs for the majority of the ride.

I think this is a great update for individuals with a power meter on bikes with a pre-existing watt table. Thanks for letting me try it out.  Please let me know if i can provide any additional info. 

[debug-Fri_Mar_27_18_24_47_2026.log.zip](https://github.com/user-attachments/files/26336902/debug-Fri_Mar_27_18_24_47_2026.log.zip)
[debug-Sun_Mar_29_14_56_35_2026.log.zip](https://github.com/user-attachments/files/26336903/debug-Sun_Mar_29_14_56_35_2026.log.zip)

cagnulein:
Glad to hear it!
Consider that qz is still learning the table while you are doing the workout so it will improve.
Do you have any future suggestion or do you think it's OK to merge?

retroromantic:
Hello,

I think the core functionality seems like its good to go.

My only suggestions would be;

1) an option in the settings somewhere to allow a user with a similar setup to return to the standard power table if they need to for some reason. 

2) perhaps instructions on how to export/import the power table for back up purposes in case a user needs to do a reinstall.

Thanks

cagnulein:
Thanks.

1. You can do this simply setting the power sensor setting to disabled (or simply pressing Ok there)

2. You can do saving and loading the profiles or the settings already

retroromantic:
Ok thats all i have. Thanks again.

PR #4026, opened 2025-12-26
Fix workout timer stopping 1 second early

cagnulein:
Fixed bug in trainprogram::totalElapsedTime() where it was using
QTime(0, 0, ticks) which creates an invalid QTime for any workout
longer than 59 seconds. Changed to QTime(0, 0, 0).addSecs(ticks)
to properly handle elapsed time for workouts of any duration.

This fixes the issue where workouts would appear to stop at 00:39:59
instead of 00:40:00 (for a 40-minute workout).

The fix makes totalElapsedTime() consistent with other time functions
in the file like remainingTime() and currentRowElapsedTime().


cagnulein:
@PugMamma do you also have this issue?

PR #4050, opened 2025-12-31
Removing loop.exec() from bluetooth write mechanism to avoid ios crashes

cagnulein:
This commit eliminates QEventLoop::exec() calls in Bluetooth write operations for domyostreadmill and wahookickrsnapbike devices, which were causing crashes on iOS due to nested event loop incompatibilities.

Changes:
- Replaced synchronous loop.exec() with async queue-based system
- Implemented WriteRequest queue with processWriteQueue() method
- Added timeout management with QTimer for async operations
- Updated characteristicWritten/packetReceived handlers for queue processing
- Removed all #ifdef Q_OS_IOS conditionals in wahookickrsnapbike
- Unified Bluetooth handling code across all platforms using pure Qt

The new architecture:
1. writeCharacteristic() enqueues write requests instead of blocking
2. processWriteQueue() handles requests sequentially
3. Completion signals (characteristicWritten/packetReceived) trigger next item
4. Timeout handling prevents queue stalls without blocking main thread

This eliminates the problematic pattern of calling loop.exec() from within the main event loop, which was causing watchdog timeouts and crashes on iOS. The solution is cross-platform compatible and improves responsiveness on all platforms.

Fixes iOS crash issues related to nested event loops in Bluetooth operations.


cagnulein:
reference: ssssgibbs

PR #4059, opened 2026-01-01
[iOS] Fix Bluetooth device search timeout issue when a device is cached

cagnulein:
…nnection

When connecting directly to a cached device (e.g., IC BIKE on iOS), the Bluetooth discovery was stopped immediately, preventing accessory devices (heart rate monitors, power sensors, cadence sensors) from being discovered within the standard 10-second timeout window.

This fix removes the immediate stopDiscovery() call in the direct connection path, allowing the discovery process to continue for the full 10-second timeout. This gives accessories enough time to be found while still maintaining the fast direct connection to the cached primary device.

The discovery will now stop naturally after:
- 10 seconds (standard timeout), OR
- When all configured accessories are discovered

Fixes issue where accessories were not found when using direct connection.


cagnulein:
mail from 
Carlos G. on 1/1/2026

PR #4108, opened 2026-01-06
Add lap tracking for Peloton workouts and TrainProgram rows

cagnulein:
Implement automatic lap triggering for each completed workout row.
The lap is emitted when a row finishes, but intelligently skips
intermediate steps in ramps to avoid creating a lap every second.

Key features:
- Emit lap() signal when completing a workout row
- Detect ramp segments using rampDuration field
- Only trigger lap at the END of a ramp (when rampDuration == 0)
- Avoid lap spam during 1-second ramp intermediate steps
- Maintains compatibility with existing lap tracking system

This enables better workout interval tracking for Peloton workouts
and other structured training programs.


cagnulein:
@pugmamma

Issue #4125, opened 2026-01-09
[BUG] Weird treadmill pace on FR955 (via QZ/ANT+)


andtwele:
My aim is to use QZ to feed speed/pace and distance from my Fitshow treadmill (Cardiostrong TX30) to my Garmin watch. 
I am using an ANT+ dongle (same manufacturer as the Garmin one) on my Android tablet for this purpose. My watch is a Forerunner 955, I have also access to a HRM Pro and I am using the latest nightly of QZ.

After activating the ANT+ cadence option, I can connect my watch to a new run sensor ("19353") and when doing a workout, a pace and distance are being shown on my watch. 

The problem is that the pace (and accordingly also the distance) that are being shown on the Forerunner are completely weird, in most cases much too high and the pace decreases the further the treadmill speed increases (see screenshot from Garmin Connect).

**To Reproduce**
1. Start a treadmill run activity with activated run sensor (ANT+ from QZ)
2. Start workout on QZ
3. Change speed
4. Observe the speed/displayed on Garmin watch and QZ
5. See that speed/pace on Garmin does not match the one on treadmill/QZ

**Expected behavior**
The speed transferred via ANT+ matches that of the treadmill and what is displayed in QZ.

To test the behavior, I did a small workout where I went from 6 to 12 km/h, changing the speed approx. every 30sec.

**Screenshots**

<img width="2524" height="1683" alt="Image" src="https://github.com/user-attachments/assets/34c2fc6c-3128-4f35-9643-7a73994eeb49" />

<img width="2301" height="787" alt="Image" src="https://github.com/user-attachments/assets/bc578b11-c127-40b4-bf3b-8a84b839b661" />

<img width="1080" height="1064" alt="Image" src="https://github.com/user-attachments/assets/8a7fbcca-2f29-4e09-b9fd-3ffe82cac922" />

**Smartphone (please complete the following information):**
 - Device: Honor 10
 - OS: Android
 - Version 15

[Fr. Jan. 9 14_45_19 2026_session.xml](https://github.com/user-attachments/files/24529681/Fr.Jan.9.14_45_19.2026_session.xml)

[debug-Fri_Jan_9_14_42_40_2026_030510.log.txt](https://github.com/user-attachments/files/24529774/debug-Fri_Jan_9_14_42_40_2026_030510.log.txt)

cagnulein:
mmm from the log the speed sent by ant to the watch seems fine.
[...]

PR #4129, opened 2026-01-10
Revert SDMChannelController speed TX calculation changes

cagnulein:
Reverts the speed encoding changes from commits 6f166d2 and eb4b73b.
Restores the original speed calculation logic:
- payload[4] = (byte) speedM_s
- payload[5] = (byte) ((speedM_s - (double)((int)speedM_s)) / (1.0/256.0))
- payload[6] = (byte) stride_count (without increment)

Reverted in both TimerTask and TX event handler.


cagnulein:
#4125

Issue #4131, opened 2026-01-10
Rouvy/dumb bike - cadence to 0, watts still transmitting


BeaSchirmer:
Stopping pedalling on magnetic bike now transmits cadence to zero ok,
BUT wattage/HR is still transmitting-avatar cycles on.

<img width="1170" height="2532" alt="Image" src="https://github.com/user-attachments/assets/569a2de3-697b-4321-976b-58d883724a4b" />

<img width="1170" height="2532" alt="Image" src="https://github.com/user-attachments/assets/b2f4fb2e-ae02-478c-a60e-00ec77fd8e7c" />

*Describe the bug**

**To Reproduce**

**Expected behavior**

**Screenshots**

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Append a debug log**

 Follow this guide https://github.com/cagnulein/qdomyos-zwift/wiki/How-do-i-get-the-debug-log-in-case-something-doesn't-work%3F
  
  
**Additional context**

cagnulein:
Adding debug log [debug-Sat_Jan_10_14_48_24_2026.log](https://github.com/user-attachments/files/24544050/debug-Sat_Jan_10_14_48_24_2026.log)

cagnulein:
Sent an update. Let me know

Roberto Viola

I sometimes trade time with my daughter to help. If this was valuable,
please support my time:
https://buymeacoffee.com/cagnulein
https://www.patreon.com/cagnulein/join
https://paypal.me/cagnulein

Software engineer and open source enthusiast, QZ Creator
http://robertoviola.cloud
http://qzfitness.com/


Il giorno sab 10 gen 2026 alle 16:16 BeaSchirmer ***@***.***>
ha scritto:

cagnulein:
@BeaSchirmer Are we OK with this?

BeaSchirmer:
Yes okSent from my iPhoneOn 3 Feb 2026, at 05:15, Roberto Viola ***@***.***> wrote:﻿cagnulein left a comment (cagnulein/qdomyos-zwift#4131)
@BeaSchirmer Are we OK with this?

PR #4134, opened 2026-01-10
Fix watts not resetting to zero when cadence stops in cscbike

cagnulein:
When using wattsFromHR on cscbike, watts were not going to zero when
cadence reached zero. Added explicit cadence check before calculating
watts from HR to ensure watts reset to zero when user stops pedaling,
while avoiding circular dependency with speed_power_based setting.


cagnulein:
#4131

Issue #4018, opened 2025-12-23
Delayed Resistance Changes when Virtual Shifting


kcknuckey:
**Issue Description**

When shifting to a higher gear in Zwift, trainer resistance takes approx 10 seconds or so before the resistance changes take effect. When changing to a lower gear, resistance changes are somewhat quicker, but still delayed.

The attached debug file should illustrate the situation:
-  I selected 'Tick Tock' as the route in Zwift in order to use a flat route where the only resistance changes should be from shifting into a higher or lower gear (assuming a consistent cadence)
- I start riding at gear 1, which obviously is very low resistance, and choose a cadence that I try to maintain through the test
- I then rapidly change all the way up to gear 24.  Both qdomyos and Zwift instantly show the gear change.
- Resistance from the trainer does not change for 8-10 seconds
- Eventually the resistance increases to the appropriate level
- I then take the gearing all the way back down to 1. Both qdomyos and Zwift instantly show the gear change.
- Resistance from the trainer again takes a long time to decrease to low resistance, though a little quicker to reduce than to increase

...so the setup is functional in that gear changes show on both my phone and Zwift as soon as I shift gear, but the changes in resistance from the trainer take a long time to take effect. This makes it very difficult to time gear changes for sprints or when hitting a steep hill.

**Expected Behaviour**

The trainer is clearly capable of changing resistance a lot quicker, because if I'm pedaling on the flat and suddenly hit a hill, the trainer resistance changes very rapidly.  I'm not sure why gear changes should take so long to take effect..

**My Setup**
 
Hardware
- iPhone 17 Pro
- iPad Mini A17 Pro (current 2024 model)
- Wahoo Kickr 2018

OS
 - Phone OS: iOS 26.2
 - iPad Mini OS: iOS 26.2
 - Wahoo Kickr Firmware: 3.5.2

**Append a debug log**

[debug-Tue_Dec_23_11_20_55_2025.log](https://github.com/user-attachments/files/24315682/debug-Tue_Dec_23_11_20_55_2025.log)

  
**Additional context**

cagnulein:
![image](https://github.com/user-attachments/assets/1f058f4d-d0d4-458e-98ba-de3c5823618d)

kcknuckey:
By the way, I am running version 2.20.17 of qdomyos

cagnulein:
In the above screenshot you can see that the whole process takes 200 milliseconds from the input request to Qz to the answers of the trainer.

[...]

PR #4156, opened 2026-01-14
Fix: Initialize lastGearValue to prevent unwanted setResistance on init (#4018)

cagnulein:
Problem:
- lastGearValue was initialized to -1 (default)
- During init, it was not set to the current gears() value
- On first update after init, the condition "lastGearValue != gears()"
  was always true (-1 != 0)
- This caused an unwanted setResistance command (0x40 0x5b 0x3f) to be
  sent immediately after setSimMode

Solution:
- Initialize lastGearValue = gears() during init (line 264)
- This prevents false gear change detection on first update
- setResistance is now only sent when there's an actual gear change

Fixes issue #4018 - Resistance Changes when Virtual Shifting


cagnulein:
#4881

PR #4169, opened 2026-01-16
Add Garmin device training fit file Session

cagnulein:


cagnulein:
not needed

cagnulein:
Required for the endurance score

Mail "QZ app - Endurance Score in Garmin Connect"

PR #4173, opened 2026-01-17
Fix settings tiles switch behavior

cagnulein:
Replace SwitchDelegate with IndicatorOnlySwitch in settings-tiles.qml to ensure switches only toggle when clicking on the indicator itself, not the entire area. This brings the behavior in line with settings.qml.

Changes:
- Replace SwitchDelegate with IndicatorOnlySwitch for cadenceColorEnabled
- Replace SwitchDelegate with IndicatorOnlySwitch for pelotonResistanceColorEnabled
- Replace SwitchDelegate with IndicatorOnlySwitch for tile_biggears_swap

### DIFF
--- a/src/AccordionCheckElement.qml
+++ b/src/AccordionCheckElement.qml
@@ -53,7 +53,7 @@ ColumnLayout {
         }
 
     }*/
-    SwitchDelegate  {
+    IndicatorOnlySwitch {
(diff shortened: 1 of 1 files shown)


cagnulein:
better to don't do this because it also affects the virtual device setting in the main qz settings

Issue #4171, opened 2026-01-17
Proform 325CSX resistance not changing when changed in the QZ app


GilbertLittle:
**Describe the bug**
Proform 325 CSX - model PFEX439210INT.0 connects to the app fine. When using Bike option specific for this model the app reads all values, including resistance changes from 1 to 22.

But increasing/decreasing the resistance on the QZ app via the +/- buttons has no effect on the bike.

**To Reproduce**
Open QZ app on android tablet. Connect via bluetooth to I_ED device. Change bike options to Proform 325csx. Start riding. Try to change resistance in app. Bike does not change.

**Expected behavior**
Resistance on bike should change.

**Device - Samsung galaxy tab A (2016). First tried when running android 8.1, then tried when running lineage OS 20 (needed this to get the ifit app to run.)

**Append a debug log**

[btsnoop_hci.log](https://github.com/user-attachments/files/24685602/btsnoop_hci.log)
  
  
**Additional context**
When using the ifit app in manual workout can change the resistance via the bluetooth connection. I have attached the btsnoop_hci.log from when I was using the ifit app to change the resistance.

When using ifit app without a paid membership you can only set the resistance from 4 to 22, in increments of 2 - e.g 4,6,8 ... 16,18.22.
So the log is of me connecting to the bike, then changing the resistance to 4, then 6, then 8 all the way to 22, then all the way back down to 4.

I notice you got it working for a version of this bike in this bug: 
https://github.com/cagnulein/qdomyos-zwift/issues/2359

cagnulein:
thanks I will check it asap

Roberto Viola

I sometimes trade time with my daughter to help. If this was valuable,
please support my time:
https://buymeacoffee.com/cagnulein
https://www.patreon.com/cagnulein/join
https://paypal.me/cagnulein

Software engineer and open source enthusiast, QZ Creator
http://robertoviola.cloud
http://qzfitness.com/


Il giorno sab 17 gen 2026 alle ore 07:37 Gil Little <
***@***.***> ha scritto:

cagnulein:
did the first very raw implementation here https://github.com/cagnulein/qdomyos-zwift/actions/runs/21095835322

enable the new profile under the qz settings, bike options, nordictrack options and restart qz

[...]

PR #4174, opened 2026-01-17
Proform 325CSX resistance not changing when changed in the QZ app (Issue #4171)

cagnulein:


cagnulein:
Proform 325CSX resistance not changing when changed in the QZ app (Issue #4171)

PR #4180, opened 2026-01-18
Add iOS volume-to-gears sync and range mapping features

cagnulein:
Implements two new iOS-specific features for volume button gear control:

1. Volume-to-Gear Sync on Startup (ios_volume_gears_sync):
   - Aligns gear to current iOS volume level (0-16) when device connects
   - Example: if volume is at level 10, gear is set to 10
   - Helps maintain useful gear range throughout the session

2. Volume Range Mapping (ios_volume_gears_range):
   - Maps 17 iOS volume levels (0-16) to custom gear range
   - Configurable min/max gears (default: 1-24, Zwift reality-bender range)
   - Volume buttons jump directly to mapped gear instead of incrementing by 1
   - Formula: gear = min + (volumeLevel / 16.0) * (max - min)

Settings added:
- ios_volume_gears_sync (bool, default: false)
- ios_volume_gears_range (bool, default: false)
- ios_volume_gears_min (double, default: 1.0)
- ios_volume_gears_max (double, default: 24.0)

All features are iOS-only and clearly marked in UI.


cagnulein:
michael casper

PR #4214, opened 2026-01-25
Improve Garmin Connect device lifecycle management on iOS

cagnulein:
- Add device UUID tracking to properly maintain device references
- Implement proper cleanup by unregistering app message delegates when devices become invalid or disconnected
- Fix indentation in urlParser method

This ensures proper resource cleanup and prevents potential memory leaks when Garmin devices are disconnected or become unavailable.


cagnulein:
the original user fixed with updating its fenix 8 to the beta firmware version

Issue #4224, opened 2026-01-27
[BUG] Workout Editor: Editing Speed


andtwele:
**Describe the bug**
When trying to type in a number for speed (km/h) in the workout editor, the keyboard automatically closes. I am not able to type in a comma or any position after the decimal point, only by using the + and - buttons.

**To Reproduce**
1) Open workout editor
2) Try to type in a speed such as "14,5"
3) The on-screen keyboard automatically closes 

**Expected behavior**
That I am able to directly type a speed such as "14,5" km/h without using the +/- buttons.

HW/SW: Android 15 Tablet, Honor 10
QZ: Latest Nightly (27-01-2026)

cagnulein:
yeah i noticed that too, let me try to fix it!

Roberto Viola

I sometimes trade time with my daughter to help. If this was valuable,
please support my time:
https://buymeacoffee.com/cagnulein
https://www.patreon.com/cagnulein/join
https://paypal.me/cagnulein

Software engineer and open source enthusiast, QZ Creator
http://robertoviola.cloud
http://qzfitness.com/


Il giorno mar 27 gen 2026 alle ore 10:39 andtwele ***@***.***>
ha scritto:

cagnulein:
fixed
https://github.com/cagnulein/qdomyos-zwift/actions/runs/21393084101/artifacts/5270618854
could you please check it?
thanks!

Roberto Viola

I sometimes trade time with my daughter to help. If this was valuable,
please support my time:
https://buymeacoffee.com/cagnulein
https://www.patreon.com/cagnulein/join
https://paypal.me/cagnulein

Software engineer and open source enthusiast, QZ Creator
http://robertoviola.cloud
http://qzfitness.com/


Il giorno mar 27 gen 2026 alle ore 10:41 cagnulein ***@***.***> ha
scritto:

andtwele:
Editing speed works fine now. 
The side-effect is however, that speed does not get updated anymore when pace is edited. 
As most runners, I am usually defining my workouts using pace. But what will QZ use for the workout when pace (min/km) is not equivalent to speed (km/h) anymore?

cagnulein:
argh i tried on my mac and it was working correctly. let me fix this as well. Sorry for the issue anyway the workout will be ok (this is only an issue in the UI).

andtwele:
Thanks. So will QZ take pace or speed then? Ideally of course, both should be linked.

cagnulein:
[...]

PR #4225, opened 2026-01-27
Prevent re-render on interval field updates to maintain focus

cagnulein:
## Summary
This PR improves the user experience when editing workout intervals by preventing full re-renders when updating synced fields (speed/pace). This keeps the input focus and mobile keyboard open while the user is actively editing.

## Key Changes
- Removed `renderIntervals()` calls when updating speed and synced pace fields
- Added direct DOM updates to synced input fields instead of re-rendering
- When pace field changes and syncs to speed: directly update the speed input value
- When speed field changes and syncs to pace: directly update the pace input value using `speedToPace()` conversion

## Implementation Details
- Uses `document.querySelector()` with `data-index` and `data-key` attributes to target specific input fields
- Maintains state updates while only updating the DOM for the affected field
- Preserves existing chart and status updates via `updateChart()` and `updateStatus()`
- Handles empty values appropriately (converts empty strings to undefined in state, empty strings in DOM)

This approach provides a smoother editing experience, especially on mobile devices where re-renders would dismiss the keyboard.

https://claude.ai/code/session_01PX2BfeXuZgwAnfHckfS3Gw


cagnulein:
#4224

PR #4236, opened 2026-01-29
Add configurable speed correction gain for Stryd sensor integration

### LINKED ISSUE #4167: Runn sensor control treadmill speed
[debug-Thu_Jan_15_18_12_51_2026.log](https://github.com/user-attachments/files/24652263/debug-Thu_Jan_15_18_12_51_2026.log)

Hello,

I have a JTX sprint 9 treadmill and have am able to control speed to my treadmill using the Runna app and QZ.

I’ve been using a Runn sensor paired with my watch to track Runna treadmill sessions. I’ve noticed that when the treadmill is set to 14 kph, the Runn sensor reports a speed of about 13.8 kph on the watch. Over time, this causes the interval distance to drift, which means I can’t reliably use Runna to control the treadmill.

I was wondering whether it would be possible to use the Runn sensor purely for data reading, while allowing QZ to scale and control the treadmill speed so that the reported speed matches the target speed.

Example:
Target speed: 14 kph
Treadmill set speed: 14 kph
Runn sensor reading: 13.8 kph
QZ applies scaling
Treadmill controlled speed: 14.2 kph
Runn sensor reading: 14 kph

Then send completed workout data to Garmin.

I have attached a debug showing my treadmill being controlled by Runna via QZ

cagnulein:
## Summary
This PR adds a new configurable setting to control how aggressively QZ corrects treadmill speed based on Stryd sensor readings. This allows users to fine-tune the speed correction behavior to avoid oscillations or achieve faster response times.

## Key Changes
- **New Setting**: Added `stryd_speed_correction_gain` setting with a default value of 1.0 (full correction)
  - Range: 0.0 - 1.0
  - Lower values (e.g., 0.3) apply gradual corrections to avoid speed fluctuations
  - Higher values apply more aggressive corrections

- **Core Logic Update** (`treadmill.cpp`):
  - Modified speed correction calculation to multiply the delta by the correction gain factor
  - Enhanced debug logging to show the corrected delta, raw delta, and gain value for troubleshooting

- **Settings Infrastructure**:
  - Added setting declaration in `qzsettings.h` with default value of 1.0
  - Registered setting in `qzsettings.cpp` and incremented `allSettingsCount` to 856
  - Added QML property binding in `settings.qml`

- **UI Component** (`settings.qml`):
  - Added input field for users to adjust the correction gain (0.0 - 1.0 range)
  - Included validation to ensure values stay within acceptable range
  - Added descriptive help text explaining the purpose and recommended values
  - UI element only visible when Stryd speed override is enabled

## Implementation Details
- The correction gain is applied as a multiplier to the speed delta before subtracting from the requested speed
- Input validation prevents values outside the 0.0 - 1.0 range
- The setting persists across application restarts
- Debug output now includes all three values (corrected delta, raw delta, gain) for easier tuning

https://claude.ai/code/session_01A3c7YGsoALtmFszVZ8rLxm


cagnulein:
#4167

Issue #4288, opened 2026-02-08
[BUG] Sportstech Rower RSX500


xmnL94:
**Describe the bug**
Title:
[Device Support] Sportstech RSX500 Rower - Connected but 0 Watts / Data not flowing
​Body:
​Describe the bug
​I am trying to connect my Sportstech RSX500 Rower to QZ to bridge it to MyWhoosh (as a Concept2 emulator). While QZ successfully connects to the rower (green status), the Watts, Cadence (SPM), and other metrics remain at 0 during rowing.
​To Reproduce
​Set Device Type to "iConsole Rower" or "Sportstech SX600/RSX500".
​Connect the rower via Bluetooth.
​Start rowing.
​Observe that all values remain at 0.
​Technical Details
​Device: Sportstech RSX500 (Rower)
​Protocol: iConsole / Toorx
​Status: Roberto suggested disabling "Fake Rower" settings, but this did not solve the issue. Watts are still 0.
​Goal: Use the RSX500 with MyWhoosh. MyWhoosh requires the data to be broadcasted as a Concept2 (PM5) device.
​Bluetooth Diagnostics (via nRF Connect)
​Connection is possible.
​Shows Descriptor UUID: 0x2902 (CCCD).
​However, no data flow is visible without a proper handshake.
​Additional Context
​The rower works perfectly with the original "Sportstech Live" app, so the hardware is sending data. It seems like a specific handshake or characteristic mapping for the RSX500 is missing.
​Attached Debug Log:
(Hier solltest du erwähnen, dass du das Log bereits per E-Mail gesendet hast oder es hier hochladen, falls du die Datei hast.)
​I have already emailed the debug log to Roberto, or I will attach it to this issue if required.

**Smartphone (please complete the following information):**
 - Device: Xiaomi P14
 - OS: Android
 - Version 14

[debug-Sun_Feb_8_18_41_48_2026.log](https://github.com/user-attachments/files/25163357/debug-Sun_Feb_8_18_41_48_2026.log)

cagnulein:
Got it and patched 

https://github.com/cagnulein/qdomyos-zwift/actions/runs/21803121800


Google is very slow in these days, reviewing the app on the store, even the beta ones. For this reason, in order to check right now if the patch that I did is right, you can use the android build in the link (check at the bottom of the page in about 1 hour). In order to test it you need first to remove the current QZ version on your phone, install the one that you downloaded.
[...]

PR #4289, opened 2026-02-08
Add EW-ST- Bluetooth name prefix for iconsole rower detection

cagnulein:
Adds "EW-ST-" as a recognized device name prefix in bluetooth.cpp so
that EW-ST devices are detected as iconsole rowers when the
iconsole_rower setting is enabled, and as generic FTMS rowers otherwise.

https://claude.ai/code/session_01SMg787bSRaEy2uFf3mMgj3


cagnulein:
#4288

PR #4292, opened 2026-02-09
Mywhoosh compatibility: Add Zwift inclination helper and gear init

cagnulein:
Introduce sendZwiftPlayInclination to centralize platform-specific Zwift hub protobuf creation (iOS/Android) and send the inclination command via writeCharacteristicZwiftPlay. Use this helper in FTMS characteristic handling and add a workaround in update(): send an initial 0.0 inclination before the first gear command because gear commands don't work until an inclination command has been sent. Add gearInclinationSent flag to track this and reset it on disconnect. Also replace duplicated platform-specific code with the new helper and add debug logging for failures.


cagnulein:
"Qz zwift ride mywoosh" email from ninours

PR #4317, opened 2026-02-15
Fix #2144: Persist web server port across restarts

### LINKED ISSUE #2144: [REQ] Static Port for Browser app

eenterwebz:
## Summary
- The inner web server port was reset to 0 (random) on every app startup in `homeform.cpp`, causing a different port each time
- Root cause: `settings.setValue(sKey + "port", 0)` was called unconditionally on every launch, overwriting any previously saved port
- Now only sets port to 0 on first run (when the key doesn't exist yet)
- Always persists the assigned port after successful bind
- Adds graceful fallback to a dynamic port if the saved port is already in use

## Changes
- **`src/homeform.cpp`**: Guard port reset with `settings.contains()` so saved port isn't overwritten
- **`src/webserverinfosender.cpp`**: Always persist port after bind; try saved port first, fall back to dynamic if in use

## Test plan
- [x] Launched QZ on Windows, noted port (61611) in overlay URL
- [x] Closed QZ completely, relaunched — same port (61611) persisted
- [x] CI passes on all platforms (Linux, Windows MinGW, Windows MSVC, iOS, Android)

Closes #2144

### DIFF
--- a/src/homeform.cpp
+++ b/src/homeform.cpp
@@ -180,7 +180,8 @@ homeform::homeform(QQmlApplicationEngine *engine, bluetooth *bl) {
 
     settings.setValue(sKey + QStringLiteral("enabled"), true);
     settings.setValue(sKey + QStringLiteral("type"), TEMPLATE_TYPE_WEBSERVER);
-    settings.setValue(sKey + QStringLiteral("port"), 0);
+    if (!settings.contains(sKey + QStringLiteral("port")))
+        settings.setValue(sKey + QStringLiteral("port"), 0);
     this->innerTemplateManager =
         TemplateInfoSenderBuilder::getInstance(innerId, QStringList({QStringLiteral(":/inner_templates/")}), this);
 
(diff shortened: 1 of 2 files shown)


cagnulein:
Thanks! Did you already try it?

eenterwebz:
@cagnulein yes, tested it. After restart got the same port 61611 as stated in the PR description.
image

cagnulein:
Perfect! Thanks! I will check it too and I will merge it!

Roberto Viola

please support my time:
https://paypal.me/cagnulein

http://robertoviola.cloud
http://qzfitness.com/


Il giorno dom 15 feb 2026 alle 18:46 eenterwebz ***@***.***>
ha scritto:

cagnulein:
mmm but what happen if the port is busy?

cagnulein:
probably the best safest approach would be adding a new setting by default to 0. if a user wants a static port it can choose one. what do you think? because in the way that it's done now, the port is not written back if it's closed, or am i wrong?

cagnulein:
@eenterwebz What do you think?

cagnulein:
@eenterwebz Any news?

cagnulein:
Thanks again for the implementation and testing. I picked this back up because the same request came up again. I moved the final version to #5078 so I could also fix the busy-port fallback: after falling back to port `0`, we need to persist `serverPort()` rather than the old occupied port. Closing this PR as superseded by #5078.

PR #4322, opened 2026-02-16
ci: fix silent build failure in Android build steps

eenterwebz:
## Summary

Split `qmake && make -j4 && make install` chains into separate commands in 4 Android build jobs.

**Bug:** In bash with `set -e` (GitHub Actions default), commands in `&&` chains are [exempt from errexit](https://www.gnu.org/software/bash/manual/bash.html#The-Set-Builtin). If `make -j4` fails, the `&&` chain returns non-zero but `set -e` does NOT exit the script. Subsequent `sed`/`cat` commands on following lines run and succeed, making the step report **success** even though the build failed. The actual failure only surfaces later as a confusing "Cannot find application binary" error in the APK step.

**Fix:** Replace `&&` chains with separate commands on separate lines so `set -e` catches each failure immediately.

### Jobs fixed
- `android-build` — "Build QZ for Android (4 ABIs)"
- `nordictrack-build` — "Set Android NDK 21 && build"
- `peloton-bike-plus-build` — "Set Android NDK 21 && build"
- `peloton-bike-build` — "Set Android NDK 21 && build"

### Before (buggy)
```yaml
run: |
  qmake ... && make -j4 && make INSTALL_ROOT=... install
  sed -i '...' deployment-settings.json   # runs even if make fails!
  cat deployment-settings.json             # exits 0, masking the failure
```

### After (correct)
```yaml
run: |
  qmake ...
  make -j4        # set -e catches failure here immediately
  make INSTALL_ROOT=... install
  sed -i '...' deployment-settings.json
  cat deployment-settings.json
```

## Test plan
- [x] Verified our previous CI run (22036799631) passes — code compiles fine, the split doesn't change behavior on success
- [x] Verified PR #4321's build failure would now correctly fail at "Build QZ for Android" step instead of silently passing

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -682,7 +682,9 @@ jobs:
 
       - name: Build QZ for Android (4 ABIs)
         run: |
-          qmake -spec android-clang 'ANDROID_ABIS=armeabi-v7a arm64-v8a x86 x86_64' 'ANDROID_NDK_ROOT=/usr/local/lib/android/sdk/ndk/21.4.7075529' && make -j4 && make INSTALL_ROOT=${{ github.workspace }}/output/android/ install
+          qmake -spec android-clang 'ANDROID_ABIS=armeabi-v7a arm64-v8a x86 x86_64' 'ANDROID_NDK_ROOT=/usr/local/lib/android/sdk/ndk/21.4.7075529'
+          make -j4
+          make INSTALL_ROOT=${{ github.workspace }}/output/android/ install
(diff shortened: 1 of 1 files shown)


eenterwebz:
@cagnulein https://github.com/cagnulein/qdomyos-zwift/pull/4321 seems to be NOT able to have build the APK but error was swallowed. This PR should make it visible next time it happens. But I could not reproduce your error with my test https://github.com/cagnulein/qdomyos-zwift/pull/4323  my current assumptions it was a race condition as I was not able reproduce it.

cagnulein:
Let me try to run it again
Thanks

Roberto Viola

please support my time:
https://paypal.me/cagnulein

http://robertoviola.cloud
http://qzfitness.com/


Il giorno lun 16 feb 2026 alle 21:53 eenterwebz ***@***.***>
ha scritto:

cagnulein:
seems fine now, thanks for the analysis!

PR #4315, opened 2026-02-15
ci: optimize CI pipeline — 2h to 21min (5.9x faster)

eenterwebz:
## Summary

Comprehensive CI pipeline optimization reducing wall-clock time from **~2 hours → ~21 minutes** (5.9x faster).

### Build Optimizations
- **Concurrency controls** — cancel superseded CI runs on rapid pushes
- **Fix redundant checkouts** — 5x repeated `actions/checkout` → 1x in 5 jobs
- **Cache vcpkg** on Windows MSVC builds (~13 min/job saved on cache hit)
- **Cache Python/PaddleOCR** with pip cache
- **Cache qthttpserver** build artifacts across 4 jobs
- **RPi 64-bit → native ARM runner** (`ubuntu-22.04-arm`): 1h50m → 8min (13.5x faster)
- **RPi 32-bit → ARM runner with QEMU**: 2h5m → 10min (12.5x faster)

### Android Build Optimizations
- **ccache for 4-ABI C++ compilation** — wraps NDK clang with ccache, caches compiled objects across runs
- **Cache Android NDK** — skip 1.1 GB download on cache hit
- **Cache Gradle** dependencies for APK build
- **Break monolithic build step** into discrete steps for better CI visibility

### iOS Build Optimizations
- **ccache with Xcode wrapper scripts** — Xcode uses absolute compiler paths (ignoring PATH), so wrapper scripts at `/tmp/ccache-wrappers/clang{,++}` call `exec ccache clang "$@"`. Passed via `QMAKE_CC`/`QMAKE_CXX`
- **Cache ccache artifacts** across runs

### Emulator Test Cleanup
- **Permission cleanup** — removed 12 non-grantable permissions (normal/special) that always threw SecurityException silently. Kept 11 runtime + 3 appops grants
- **`continue-on-error` for flaky API 33** — Android 13 emulator test is intermittently flaky (app crashes within 90s, happens on master too). Marked as allowed-to-fail using `matrix.flaky` variable so only API 33 is non-blocking

### Cache Scoping Note

For caches to benefit all PRs, this needs to be merged and run on `master` first. GitHub Actions caches are branch-scoped — PR branches can only restore caches from their own branch or the base branch (`master`). After merging, the first CI run on `master` populates all caches, and every subsequent PR automatically inherits them.

### Results — Per-Job Breakdown

| Job | Before (master) | After (PR) | Speedup |
|-----|-----------------|------------|---------|
[...]

PR #4326, opened 2026-02-17
ci: fix RPi build compatibility (ARMv6 + glibc) and add smoke tests

### LINKED ISSUE #4325: [BUG] Raspberry Pi 32 bit builds not executing
**Describe the bug**
QZ binaries fail to launch. On [2.20.24](https://github.com/cagnulein/qdomyos-zwift/actions/runs/21662694328/artifacts/5371684677), generates an illegal instruction error. On [2.20.17](https://github.com/cagnulein/qdomyos-zwift/actions/runs/20128433607), the oldest binary I could find, a segmentation fault error is generated. Both dump back out to the command prompt with little output.

`andrew@raspberrypi:~/qz $ sudo ./qdomyos-zwift-32bit -no-gui -heart-service`
`Segmentation fault`

**To Reproduce**
This was tested from a Raspberry Pi Zero W, clean installed onto a 32gb card with Raspbian Lite Trixie, updated and dependencies installed per the [installation guide](https://github.com/cagnulein/qdomyos-zwift/blob/master/docs/10_Installation.md). Chatted with Robert and `dmesg` nor `gdb` revealed anything obvious.

eenterwebz:
### Raspberry Pi build fixes

- **Fix 32-bit RPi for Pi Zero W** (fixes #4325): switch to `balenalib/raspberry-pi-debian:bullseye-build` container which uses Raspbian repos where GCC, CRT, Qt5, and glibc all target ARMv6 natively. The previous `arm32v7/debian` container produced ARMv7 binaries incompatible with the Pi Zero W's ARMv6 CPU (ARM1176JZF-S). ARMv6 binaries remain forward-compatible with all newer Pi models (2/3/4/5).
- **Fix 64-bit RPi glibc regression** from #4315: restore `arm64v8/debian:bullseye` Docker container so the binary links against glibc ≤2.31. The CI optimization PR switched to native `ubuntu-22.04-arm` (glibc 2.35), breaking Raspberry Pi OS Bullseye compatibility. The ARM64 runner is kept for speed (runs arm64 Docker natively, no QEMU).
- **Rename artifacts** from `32bit`/`64bit` to `armv6hf`/`aarch64` to clearly reflect target architectures.

### CI verification improvements

- **Add `readelf` verification steps** after both RPi builds to catch regressions:
  - 32-bit: assert `Tag_CPU_arch: v6`
  - 64-bit: assert max GLIBC ≤ 2.31
- **Add `raspberry-pi-smoke-test` job**: downloads both binaries and runs them in Docker containers matching target hardware. Checks ELF metadata (CPU arch, hard-float ABI, interpreter path) and runs the binary with `-smoke-test` to verify Qt loads without SIGILL/SIGSEGV. 32-bit smoke test is `continue-on-error` due to QEMU arm/v6 slowness.
- **Gate `upload_to_release`** on smoke test passing.

### Application change

- **Add `-smoke-test` CLI flag** (`src/main.cpp`): initializes `QCoreApplication`, prints `SMOKE_OK`, and exits cleanly. Bypasses the root check on Linux, requires no Bluetooth hardware — designed for deterministic CI verification.

### vcpkg cache fix

- **Extract vcpkg deps to `vcpkg-deps.json`** and use `hashFiles('vcpkg-deps.json')` as cache key, replacing the static `protobuf-abseil-v1` key. Cache now invalidates automatically when dependencies or baseline change.

### Root cause analysis for #4325

The `arm32v7/debian:bullseye` Docker image has always produced ARMv7 binaries (`Tag_CPU_arch: v7` confirmed via `readelf`), but the Pi Zero W has an ARMv6 CPU. The "illegal instruction" vs "segfault" difference between versions is because some 32-bit Thumb-2 instructions decode differently on ARMv6 — some trap as undefined (SIGILL), others decode as two 16-bit instructions with wrong operands (SIGSEGV). This predates the CI optimization PR #4315.

However, #4315 **did** introduce a separate 64-bit regression: switching from Docker to native `ubuntu-22.04-arm` raised the glibc requirement from 2.31 to 2.35, breaking Bullseye.

## Test plan

- [x] RPi 32-bit build succeeds
- [x] `readelf -A` shows `Tag_CPU_arch: v6` (not v7)
- [x] RPi 64-bit build succeeds
- [x] `readelf -V` shows max GLIBC ≤ 2.31
- [x] Smoke test job passes (ELF metadata + runtime checks)
- [x] vcpkg cache uses content-based key
- [ ] End-to-end: run 32-bit binary on Pi Zero W without crash

Fixes #4325

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1085,7 +1085,7 @@ jobs:
         uses: actions/cache@v4
         with:
           path: ${{ runner.workspace }}\vcpkg\installed
-          key: vcpkg-msvc2019-x64-windows-${{ hashFiles('.github/workflows/main.yml') }}
+          key: vcpkg-msvc2019-x64-windows-${{ hashFiles('vcpkg-deps.json') }}
 
       - name: Clone vcpkg
         if: steps.cache-vcpkg-msvc2019.outputs.cache-hit != 'true'
@@ -1100,17 +1100,7 @@ jobs:
       - name: Create vcpkg.json
         if: steps.cache-vcpkg-msvc2019.outputs.cache-hit != 'true'
         working-directory: ${{ runner.workspace }}
-        run: |
-          echo '{
-            "name": "qdomyos-zwift",
-            "$schema": "https://raw.githubusercontent.com/microsoft/vcpkg-tool/main/docs/vcpkg.schema.json",
-            "dependencies": [
-              "protobuf",
-              "protobuf-c",
-              "abseil"
-            ],
-            "builtin-baseline": "8c2fcacefba009d63672f9d137f192765e632c9f"
-          }' > vcpkg.json
+        run: cp ${{ github.workspace }}/vcpkg-deps.json vcpkg.json
 
       - name: Install dependencies
         if: steps.cache-vcpkg-msvc2019.outputs.cache-hit != 'true'
@@ -1288,7 +1278,7 @@ jobs:
         uses: actions/cache@v4
         with:
           path: ${{ runner.workspace }}\vcpkg\installed
-          key: vcpkg-msvc2019-x64-windows-${{ hashFiles('.github/workflows/main.yml') }}
+          key: vcpkg-msvc2019-x64-windows-${{ hashFiles('vcpkg-deps.json') }}
 
       - name: Clone vcpkg
         if: steps.cache-vcpkg-aiserver.outputs.cache-hit != 'true'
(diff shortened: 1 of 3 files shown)


eenterwebz:
@cagnulein This fixes the Pi Zero build. The ARMv6 binary also runs fine on newer Pi models (2/3/4/5), so one 32-bit build covers everything. We could also do two separate 32-bit builds if you prefer, one for the Zero and one targeting ARMv7 like before. Let me know what you think.

cagnulein:
Perfect! Thanks! @uberlegit can you test it?

cagnulein:
@eenterwebz Yes probably a 32bit separate one could be great. 
Also could you please remove the changes from the windows workflow if they are not necessary?

PR #3869, opened 2025-11-14
Add Wahoo DirCon client support for network-based device connections

cagnulein:
Implements client-side support for the Wahoo DirCon protocol, allowing QZ to connect to fitness devices (bikes and treadmills) over TCP/IP via mDNS discovery, in addition to the existing server-side broadcasting functionality.

Key Features:
- mDNS service discovery for _wahoo-fitness-tnp._tcp.local. devices
- DirconDiscovery class for automatic network device detection
- WahooDirconBike and WahooDirconTreadmill device implementations
- Full FTMS protocol support over DirCon (Indoor Bike Data, Treadmill Data)
- Support for commands (resistance, power, speed, inclination)
- Automatic device type detection (bike vs treadmill based on name)
- Integration with manual device selection in settings UI

Technical Details:
- Created DirconDiscovery class using QMdnsEngine for service discovery
- Implemented wahoodirconbike inheriting from bike base class
- Implemented wahoodircontreadmill inheriting from treadmill base class
- Added TCP socket communication using DirconPacket protocol
- Integrated with existing bluetooth discovery mechanism
- Devices appear in manual device list with "(DirCon)" suffix

Files Added:
- src/devices/dircon/dircondiscovery.h/cpp
- src/devices/dircon/wahoodirconbike.h/cpp
- src/devices/dircon/wahoodircontreadmill.h/cpp

Files Modified:
- src/devices/bluetooth.h/cpp (discovery integration)
- src/homeform.cpp (device list integration)
- src/qdomyos-zwift.pri (build system)

PR #4331, opened 2026-02-18
ANT+ control inclination as virtualbike does

cagnulein:


cagnulein:
merged with https://github.com/cagnulein/qdomyos-zwift/pull/3869 for Gaël Le G.

Issue #3897, opened 2025-11-19
[BUG] HR program stuck in one gear


DavidNr2:
**Describe the bug**
When running my own HR program it jumps from gear 1 to gear 12(of 17) on my bike and stays there for the whole session. Target resistance is changig between 35-40

**To Reproduce**
Calibrate program by running freemode 1 minute per gear on bike.
Start hr program. Warm-up -> medium hr-load -> -> more hr-load -> cool down.
I use the yesoul program on  my Odin s900 - it works when changing target resistance in the app.

**Expected behavior**
Low to medium resistance in start and progressive load

**Screenshots**

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**
Android 14

[debug-Tue_Nov_18_20_09_07_2025.log](https://github.com/user-attachments/files/23631896/debug-Tue_Nov_18_20_09_07_2025.log)
[debug-Tue_Nov_18_20_03_14_2025.log](https://github.com/user-attachments/files/23631897/debug-Tue_Nov_18_20_03_14_2025.log)

**Append a debug log**

 Follow this guide https://github.com/cagnulein/qdomyos-zwift/wiki/How-do-i-get-the-debug-log-in-case-something-doesn't-work%3F
  
  
**Additional context**
Zwift Reistance Offset -20
Zwift Resistance gain 0.36
Works very well manually but can maybe explain the jump up to gear 12?

cagnulein:
I guess I already fixed something like this with https://github.com/cagnulein/qdomyos-zwift/commit/da3b5b168eab0ddf4d5c021c5ecc06e257d1f70f
could you please try the nightly build https://github.com/cagnulein/qdomyos-zwift/releases/download/nightly-2025-11-19/android-debug.apk

Google is very slow in these days, reviewing the app on the store, even the beta ones. For this reason, in order to check right now if the patch that I did is right, you can use the android build in the link (check at the bottom of the page in about 1 hour). In order to test it you need first to remove the current QZ version on your phone, install the one that you downloaded.
This version is a 15 minutes trial (you can restart it as many times as you want) and I can unlock it if you want, send me an email to roberto.viola83@gmail.com mentioning this ticket and that you are on android.
Let me know if you have any questions

If it still doesn't work, could you please attach a fresh debug log and the workout program that you are using?

Thanks

DavidNr2:
Yes, I believe you did when I reported for the first time. I have version
[...]

PR #4352, opened 2026-02-23
Reduce bike inclination step from 0.5 to 0.2

cagnulein:
Lower the inclination adjustment increment in homeform::update from 0.5 to 0.2 for bikes without erg mode (e.g., FTMS bikes). This makes PID/HR-based inclination control apply finer, smoother changes (updated in two code paths).


cagnulein:
#3897

PR #4406, opened 2026-03-12
Joroto X2Pro appears to have wrong power output (Issue #4405)

### LINKED ISSUE #4405: [BUG] Joroto X2Pro appears to have wrong power output
**Describe the bug**
Using both Zwift and MyWoosh using Joroto X2Pro the speed and power output seems to be extremely low as to what could be expected, healthy experienced cyclist cannot keep up with anyone / complete any challenges.

Speed seems to hover around 8kmph, watts when exercising hard seem to stick around 70-90

**To Reproduce**

**Expected behavior**

**Screenshots**

**Append a debug log**

Here is my debug log

https://filebin.net/9k7irfhhn028dudb

In this log I cycled relaxed as if I was on a flat for 1 minute, I then turned resistance on my bike up to full and cycled as hard as possible for 30 seconds, then did another relaxed / flat 30 seconds.

cagnulein:


cagnulein:
Joroto X2Pro appears to have wrong power output (Issue #4405)

PR #4445, opened 2026-03-20
Virtual Yesoul Bike #4359

### LINKED ISSUE #4359: [BUG] Connecting Strava With Google Sign In Fails
I'm trying to connect to Strava using Sign in With Google.

I get this:

Image

According to Gemini, this might be because the app is doing this in a WebView instead of Chrome.

cagnulein:
Virtual Yesoul Bike #4359


cagnulein:
this can work only witrh qz on an android because the yesoul app wants to see the manufacturer code.

PR #4454, opened 2026-03-23
Fix FTMS bike +/- resistance tiles stuck when device reports physics-computed resistance

cagnulein:
## Problem

On FTMS bikes running in Indoor Bike Simulation mode (`FTMS_SET_INDOOR_BIKE_SIMULATION_PARAMS`), pressing **+** on the resistance tile and then **-** had no effect — both buttons effectively increased resistance.

Root cause: the app sends a grade/simulation command (e.g. resistance level 6 → grade 0.6%), but the device computes and reports back its own resistance from the physics simulation (e.g. reports 7, not 6). When the user pressed **-**, the code read the device-reported value (7) as the baseline, computed `7 - 1 = 6`, and sent the same grade command again — resulting in the device reporting 7 again indefinitely.

## Fix

In `homeform::Plus()` and `homeform::Minus()`, always use `lastRequestedResistance` as the baseline for the next step whenever it has been set (> 0), instead of gating it on `currentResistanceValueAvailable()`.

- First press still anchors to `currentResistance()` (device-reported), so the starting point is correct.
- Every subsequent press steps relative to the last value *we sent*, not what the device echoed back.
- This is correct for all device types: devices with 1:1 resistance mapping are unaffected (lastRequested == reported), devices that don't report resistance at all work as before, and FTMS simulation bikes now step correctly.

`currentResistanceValueAvailable()` on `ftmsbike` is preserved as `resistance_received` (semantically correct — it reflects whether the device sends resistance notifications, not whether those values match our command scale).


cagnulein:
mail from 
daan p.

date 22/3/2026

cagnulein:
[debug-Mon_Jun_15_08_47_01_2026.log](https://github.com/user-attachments/files/28955887/debug-Mon_Jun_15_08_47_01_2026.log)

Issue #4495, opened 2026-04-03
[BUG] Airpod Pro 3 Heart Rate data are not showing on QZ App


ccchiou:
**Describe the bug**
Airpod Pro 3 Heart Rate data are not showing on QZ App

**To Reproduce**
1. Spin up my echelon EX-15
2. Open QZ App on iPhone(With Debug log enabled already)
3. Take out My Airpod Pro 3 from case and wear it
4. Heart Rate  is 0 on QZ App
5. Open "Health" App on Iphone
6. Make sure my heart rate is show in in app
7. Wear the Airpod Pro 3 for 3 min
8. QZ App Heart Rate is still 0

Heart Belt Name is "Disabled"
other HR settings are Default(All off except for Disable HRM from Machinery)
QZ App version 2.20.28
QZ App allow access to Health HR Data

**Expected behavior**
Expect QZ Heart rate is not 0 when wearing Airpod Pro3

**Screenshots**

![Image](https://github.com/user-attachments/assets/85916a2d-812c-49f8-b9ca-57ec8087f910)

![Image](https://github.com/user-attachments/assets/16d4f743-4e20-44ee-bbcc-24dc69c28031)

![Image](https://github.com/user-attachments/assets/29021426-d668-42b4-a788-8933abea7579)

![Image](https://github.com/user-attachments/assets/a1df3bb1-f8ae-4d13-8a78-f1b05afdcdb6)

**Smartphone (please complete the following information):**
 - Device: IPhone14Pro
 - OS: iOS 26.4
 - Version 26.4

**Append a debug log**

[debug-Thu_Apr_2_22_06_10_2026.log](https://github.com/user-attachments/files/26453702/debug-Thu_Apr_2_22_06_10_2026.log)

  
  
**Additional context**

cagnulein:
Thanks I will check it ASAP

cagnulein:
i guess i got why it's not working

`Thu Apr 2 22:06:10 2026 1775192770451 Debug: /Volumes/workspace/repository/src/ios/lockscreen.mm bool lockscreen::appleWatchAppInstalled() Apple Watch is paired and app is installed`

qz is giving the priority of the hr from the watch. Could you please try to uninstall the qz app on the watch for testing this?

ccchiou:
Thanks for the advice, I have a really old Apple watch that I'm no longer using. And it;s out of power at the time.
I have to dig it out of storage and charge it.
[...]

PR #4497, opened 2026-04-03
Airpod Pro 3 Heart Rate data are not showing on QZ App (Issue #4495)

cagnulein:


cagnulein:
Airpod Pro 3 Heart Rate data are not showing on QZ App (Issue #4495)

cagnulein:
The problem was the fact that the workout wasn't started on iPhone due to metrics are all 0 and the apple watch was paired with qz installed

PR #4511, opened 2026-04-07
Lifespan C7000i bike

cagnulein:


cagnulein:
@copilot in the settings.qml put the new lifespan setting as a new accordition in the bike options not under the fit plus options

PR #4535, opened 2026-04-15
Android Restart Pending

cagnulein:


cagnulein:
It works but now sometimes android says "the app keeps stopping" 
I guess because we are not exiting in a clear way

PR #4552, opened 2026-04-19
M3i live ios workout fixed

cagnulein:


cagnulein:
Mail: QZ customer: post to apple health (without watch)
 from: Kaushal S.

PR #4554, opened 2026-04-20
Remove unused changeSpeedAndInclination signal

Qhilm:
This is to make sure the other fix I did (disabling HR controller for 10s when changing speed) does not break at some point in the future: At the moment it focusses on changeSpeed but there is also changeSpeedAndInclination, but the latter is never used.

Hence, suggesting here to remove this "dead code", unless you had plans for it?

- Remove unused signal declaration from trainprogram.h
- Remove dead signal connections from homeform.cpp and mainwindow.cpp

### DIFF
--- a/src/homeform.cpp
+++ b/src/homeform.cpp
@@ -1654,8 +1654,6 @@ void homeform::trainProgramSignals() {
                     &treadmill::changeFanSpeed);
             connect(trainProgram, &trainprogram::changeInclination, ((treadmill *)bluetoothManager->device()),
                     &treadmill::changeInclination);
-            connect(trainProgram, &trainprogram::changeSpeedAndInclination, ((treadmill *)bluetoothManager->device()),
-                    &treadmill::changeSpeedAndInclination);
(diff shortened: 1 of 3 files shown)


cagnulein:
some treadmills (like mine domyos) must send speed and inclination at the same time, so it shoulnd't be dead code, did you check it?

Qhilm:
I searched for "changeSpeedAndInclination" manually in the repo, I don't see anywhere where it's called, only where it's defined.

cagnulein:
i see the call in the treadmill.cpp:714

Qhilm:
my mistake, don't know how I missed that.

Qhilm:
So I checked it again, what is not being used is the signal "trainprogram::changeSpeedAndInclination", not the method.

treadmill.cpp:714 is also empty actually, can you double check or paste the code here?

In trainprogram.cpp, you can find `emit changeSpeed()` often, but zero `emit changeSpeedAndInclination()` for example.

If you plan on introducing this signal though, we shouldn't delete it.

Let me know.

cagnulein:
You are right. Let's see how it goes the #4553

PR #4556, opened 2026-04-20
Fix QMutexLocker temporary object bug

Qhilm:
Claude code seems very confident about it (granted, it always is very confident...), I have been reading up and it seems to make sense, but since I am not an expert, of course if will seems logical to me, that's what LLMs are all about – making it seem plausible.

Do have a detailed and skeptical look to this – most of all because I don't know how to test this.

---

### Problem

QMutexLocker was being instantiated as an unnamed temporary, which locks and immediately unlocks the mutex on the **same line**. This is a well-known C++ pitfall where the compiler silently accepts the code but provides zero synchronization.

### Before (Broken):

```cpp
void trainprogram::clearRows() {
    QMutexLocker(&this->schedulerMutex);  // Temporary created and destroyed
    rows.clear();                           // No lock held here!
}
```
### After (Fixed):
```cpp
void trainprogram::clearRows() {
    QMutexLocker locker(&this->schedulerMutex);  // Variable holds the lock
    rows.clear();                                  // Lock held for entire scope
}
```

### Why This Matters
The `clearRows()` and `scheduler()` methods have been running completely unprotected.
Both methods access and modify the rows list, which can be called concurrently:
- `scheduler()` called every 1 second by a timer (main thread)
- `clearRows()` called from Peloton integration (main thread, different control flow)
- rows list is accessed by both, plus read by interval/ramp calculations
### Consequences of the Bug
Without the mutex held, this creates a data race:
- List corruption during concurrent modifications
- Iterator invalidation crashes when iterating over a list being cleared
- Undefined behavior in speed/incline calculations during workouts
- Silent data loss or duplicate rows in training programs
### Changes
- `src/trainprogram.cpp:533` — Add variable name to `QMutexLocker` in `clearRows()`
- `src/trainprogram.cpp:600` — Add variable name to `QMutexLocker` in `scheduler()`
The fix ensures the mutex is held for the entire duration of each method's execution, providing proper mutual exclusion as intended.

### DIFF
--- a/src/trainprogram.cpp
+++ b/src/trainprogram.cpp
@@ -530,7 +530,7 @@ double trainprogram::avgAzimuthNext300Meters() {
 }
 
 void trainprogram::clearRows() {
-    QMutexLocker(&this->schedulerMutex);
+    QMutexLocker locker(&this->schedulerMutex);
     rows.clear();
 }
 
@@ -597,7 +597,7 @@ void trainprogram::pelotonOCRcomputeTime(QString t) {
 
 void trainprogram::scheduler() {
 
-    QMutexLocker(&this->schedulerMutex);
+    QMutexLocker locker(&this->schedulerMutex);
     QSettings settings;
     QDateTime now = QDateTime::currentDateTime();
     qint64 msecsElapsed = lastSchedulerCall.msecsTo(now);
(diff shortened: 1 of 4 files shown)


cagnulein:
I added some tests because I would like to check if there isn't a deadlock in this way since yes the code is wrong but it was working in this way since a long time

PR #4573, opened 2026-04-22
Add missing trainprogram destructor to fix resource leaks

Qhilm:
### Problem

The `trainprogram` class was missing a destructor, causing memory leaks:
- `zwift_auth_token` (allocated at line 38 with `new`)
- `zwift_world` (allocated at line 635 with `new`)  
- `h` (lockscreen pointer, allocated at line 642 with `new`)
These are heap-allocated resources that were never freed, leading to memory leaks on every `trainprogram` object destruction.

### Root Cause
The `trainprogram` class follows the RAII pattern but was missing the destructor to clean up heap-allocated resources.

### Solution
Add a destructor that properly deletes the heap-allocated members:
- `delete zwift_auth_token;`
- `delete zwift_world;` 
- `delete h;`
The `pelotonOCRsocket` is already properly parented to the `trainprogram` object via `new QUdpSocket(this)`, so Qt's parent-child mechanism will clean it up automatically.

### Changes
- `src/trainprogram.h` - Add destructor declaration `~trainprogram();` 
- `src/trainprogram.cpp` - Add destructor implementation with proper `delete` statements

### DIFF
--- a/src/trainprogram.cpp
+++ b/src/trainprogram.cpp
@@ -1993,3 +1993,9 @@ double trainprogram::totalDistance() {
     }
     return distance;
 }
+
+trainprogram::~trainprogram() {
+    delete zwift_auth_token;
+    delete zwift_world;
+    delete h;
+}
--- a/src/trainprogram.h
+++ b/src/trainprogram.h
@@ -196,6 +196,7 @@ private slots:
     lockscreen *h = 0;
 #endif
 
+    ~trainprogram();
 };
 
 #endif // TRAINPROGRAM_H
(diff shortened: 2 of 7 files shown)


cagnulein:
This is true only when you create a new program or loading a new one. It should be negligible for regular session. What do you think?

I'm saying so because it's always hard to test if a patch is fine or not even if it seems safe

Qhilm:
I agree with you on both points: It's a very rare memory leak and it's impossible to properly test.

Up to you. I like tidy code, but I have a slight OCD ;)

cagnulein:
Me too :) but I learned that 2 minutes of easy work can value a lot of headaches in the future 😂

PR #4580, opened 2026-04-24
Custom gear resistance table for virtual shifting for qt6 issue 4545

### LINKED ISSUE #4545: [REQ] Custom gear resistance table for Virtual Shifting
I use a Magene T300 Plus trainer with very light physical bike setup (a 32T chainring and 24T cog). I use Zwift Click to change gears and QZ collects gears from the game. I rely on Virtual Resistance for riding. Right now, QZ adds a linear, fixed resistance step for every gear click. Cruising works fine, but to get enough heavy resiistance to sprint, I have to click 12-15 times. This takes many seconds, which is too slow for racing. Pre-made gear ratios from Wahoo or other brands do no not fix this because they do not account for my hardware limits.

I would like a new setting that allows a custom gear resistance table. This would be asetting where user types the exact resistance or incline value for each of the 24 gears. It could be comma separated or a table.

By having a custom table, I can program the first 15 gears with small resistance jumps, giving me perfect micro-adjustments for cruising. Then I can program the last 5 gears to have massive resistance jumps. This solves my problem by giving me a change of a heavy sprint with only 3 or 4 clicks, mimicking a bike cassette in real life.

I tried setting the general gear gain to higher numbers. This fixed the sprinting because the resistance got heavy fast, but it ruined cruising. One click would make the pedals too heavy and break my cadence. I also considered auto virtual shifting however I prefer manual gear control for dynamic pacing in races.

**Additional context**
Real bicycle cassettes use exponential jumps, not linear ones. A custom table is the best overall solution because it gives every user complete freedom. Anyone could have a gear ratio that perfectly tune their virtual shifting without complex math formulas built-in.

cagnulein:
combining #1508 and #4578 for #4545


cagnulein:
it's working

PR #4583, opened 2026-04-25
Respect external cadence/heart sources and derive cadence when treadmill sensor unavailable

cagnulein:
### Motivation
- Prevent the treadmill's built-in cadence readings from overriding external sensors or companion integrations when configured to ignore them.
- Ensure cadence is still available by sourcing it from Apple Watch or by estimating it from speed when no external cadence/power sensors are present.
- Avoid conflicting heart-rate selection when companion integrations (e.g. Garmin) are enabled.

### Description
- In `characteristicChanged` added checks for external cadence sensor, `power_sensor_cadence_instead_treadmill`, and `garmin_companion` settings before using treadmill cadence and only call `parseCadence` when allowed.
- Introduced local `cadenceAvailable` flag and added calls to `cadenceFromAppleWatch()` where appropriate to prefer watch-provided cadence.
- If cadence is not available and `Speed.value()` is positive, added logic to compute cadence via `calculateCadenceFromSpeed` (only when no power/cadence sensors and no Garmin companion) and feed it to `parseCadence` after `evaluateStepCount()`.
- Adjusted heart selection to respect the `garmin_companion` setting by skipping treadmill machinery HR when the companion is enabled, and left `update_hr_from_external()`/`Heart` assignment behavior otherwise.

### Testing
- Performed a project build and ran the existing automated unit/test suite; all tests completed and passed.
- No new unit tests were added in this change.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69ec6cdd3f448325b5f7145f4e9f17b6)

Issue #173, opened 2021-02-19
Use the accelerometer to get the cadence from the treadmill


cagnulein:
https://www.facebook.com/n/?groups%2F1064761680643892%2Fpermalink%2F1148666598920066%2F&comment_id=1148674295585963&reply_comment_id=1148679388918787&aref=1613657776482486&medium=email&mid=5bb9cb0f865e4G252e20e0G5bb9cfa8e68b6G2bd&bcode=2.1613657777.Abz2m6maK6Rm8DLz61U&n_m=cagnulein%40gmail.com&lloc=view_on_facebook_btn

p3g4asus:
Fitshow threadmills send the step count in the packet. I don't know if treadmill producers set that field to a valid value though. I will test this on my treadmill ASAP and report. However the current version of the fitshowtreadmill module parses the step count information and prints it out without using it.  If you implement the cadence field in virtual treadmill I will modify the fitshowtradmill module to compute cadence using the step count coming from the packet if it is valid (not null?).

cagnulein:
Yeah sounds like a plan!

Il mer 10 mar 2021, 12:27 p3g4asus <notifications@github.com> ha scritto:

PR #4594, opened 2026-04-29
feat(ios): derive Fake Treadmill speed from Apple Watch cadence

miraicat:
Hi Roberto — thanks for the quick reply and for inviting the PR.

## Summary

This adds an opt-in iOS setting, `applewatch_as_treadmill_speed` (default off), to derive treadmill `Speed` from Apple Watch step cadence when Fake Treadmill mode is enabled.

The feature reuses the existing Apple Watch cadence path in `treadmill::cadenceFromAppleWatch()`, the existing `cadence_sensor_speed_ratio` ("Wheel Ratio") multiplier, and the existing virtual treadmill / RSC broadcast pipeline.

## Use case

This is for indoor walking or slow jogging without a physical treadmill, while broadcasting to apps like Kinomap or Zwift via FTMS / RSC.

Apple Watch cadence is already available in QZ on iOS, but without a real treadmill there was no path to convert that cadence into broadcast `Speed`, so users had to adjust the Speed tile manually.

## Design

- Default off, so existing users should see no behavior change.
- Dual-gated behind both `applewatch_as_treadmill_speed` and `fakedevice_treadmill`.
- Normal treadmill users are unaffected unless they explicitly enable both settings.
- Reuses the existing `cadence_sensor_speed_ratio` setting instead of adding another ratio control.
- When cadence reports exactly `0`, `Speed` is also set to `0` so the virtual treadmill broadcast does not keep ghosting the previous speed.
- Negative cadence is treated as "no fresh sample" and leaves `Speed` unchanged.
- A defensive ratio clamp prevents a misconfigured negative Wheel Ratio from producing negative speed.
- Per `src/CLAUDE.md`, the new setting is appended at the end of the `settings.qml` `Settings` block and `allSettingsCount` is bumped from `882` to `883`.

## UI

The toggle is placed next to "Fake Treadmill" in Experimental Features, since this feature requires Fake Treadmill mode.

The help text points users to Accessories > Cadence Sensor Options and suggests a Wheel Ratio around `0.025-0.04` for walking / slow jogging. The existing cycling default of `0.33` would be too high for this use case.

## Testing plan

- [x] iOS CI compiles cleanly
- [ ] Toggle off -> no new speed derivation
- [ ] Toggle on + Fake Treadmill on + walking with Apple Watch -> Speed updates from cadence x Wheel Ratio
- [ ] Toggle on + walking, then stop -> Speed drops to 0
- [ ] Toggle on + Fake Treadmill off -> Speed is unaffected

I have not been able to compile-verify locally because I do not have a Qt 5.15.2 + iOS toolchain set up, so I am relying on CI. Happy to iterate on anything that breaks or that you would like reshaped.

## Related

- Issue #173 — closest precedent for cadence-derived behavior
- PR #4583 — appears to touch a different Apple Watch cadence path, no conflict expected


cagnulein:
Thanks @miraicat the patch looks great, I will build an ios version tomorrow!

cagnulein:
sent the version!

miraicat:
Tested on the TestFlight build. Slow jogging at ~180 spm (Apple Watch cadence) — Speed correctly derived from cadence. With Wheel Ratio at 0.046, getting ~4.1 km/h, pace and cadence numbers all consistent. Verified end-to-end with both **Zwift** and **Kinomap** on MacBook receiving over BLE FTMS — both work. Thanks for the quick turnaround!

miraicat:
Quick follow-up: pushed a one-line help text correction (545f56e). Real-world testing showed the original 0.025-0.04 Wheel Ratio range was below typical slow jog speeds at 180 spm cadence (max ~3.6 km/h). Updated to 0.04-0.15 which covers walking through running.

cagnulein:
that's great!

PR #4598, opened 2026-04-29
Use secure SMTP transport for workout emails

cagnulein:
### Motivation
- Ensure workout summary emails are sent over an encrypted transport from connection start to avoid plaintext or partial SMTP flows and fix an incorrect compile-time macro guard.

### Description
- In `src/homeform.cpp` MailSenderThread the `SmtpClient` is now created with `SmtpClient::SslConnection` on port `465` (implicit TLS / SMTPS) including the empty-server fallback, the username macro check was corrected to `SMTP_USERNAME`, and the send loop was hardened to call `login()` and `sendMail()` only after a successful `connectToHost()`.

### Testing
- No automated tests were executed in this environment; only the file changes were reviewed and committed.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69f24bdc2dbc832599339bab5f246572)

### DIFF
--- a/src/homeform.cpp
+++ b/src/homeform.cpp
@@ -91,17 +91,18 @@ class MailSenderThread : public QThread {
 #ifdef SMTP_SERVER
 #define _STR(x) #x
 #define STRINGIFY(x) _STR(x)
-        SmtpClient smtp(STRINGIFY(SMTP_SERVER), 587, SmtpClient::TlsConnection);
+        // Use implicit TLS (SMTPS) to enforce encrypted transport from the first byte.
+        SmtpClient smtp(STRINGIFY(SMTP_SERVER), 465, SmtpClient::SslConnection);
 #else
 #pragma message "stmp server is unset!"
-        SmtpClient smtp(QLatin1String(""), 25, SmtpClient::TlsConnection);
+        SmtpClient smtp(QLatin1String(""), 465, SmtpClient::SslConnection);
         delete message;
         return;
 #endif
 
 // We need to set the username (your email address) and the password
 // for smtp authentication.
-#ifdef SMTP_PASSWORD
+#ifdef SMTP_USERNAME
 #define _STR(x) #x
 #define STRINGIFY(x) _STR(x)
         smtp.setUser(STRINGIFY(SMTP_USERNAME));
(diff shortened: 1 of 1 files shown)


cagnulein:
#4572 @Qhilm

Qhilm:
This is a separate topic from https://github.com/cagnulein/qdomyos-zwift/discussions/4572, the HTTP connection to http://robertoviola.cloud:4010 is for licensing, if the LLM read your code right.

I would stick to STARTTLS for SMTP and not switch to SMTPS. SMTPS is officially deprecated actually.

Even if STARTTLS starts unencrypted, it is secure. The only risk is that in theory it can be downgraded to fully unencrypted, but your code already prevents it:

#### Prompt
 Rather than switching to SMTPS which is deprecated, what could be done to enforce STARTTLS encryption and prevent any downgrade attack?
#### Model
Claude Opus 4.6
#### LLM output
>Looking at the SmtpClient library used by this project (src/smtpclient/src/smtpclient.cpp:188-280), the existing SmtpClient::TlsConnection mode on port 587 already prevents downgrade attacks:
>1. It sends STARTTLS to the server
>2. If the server responds with anything other than 220, it returns false and emits ServerError — it does not fall back to plaintext
>3. It then calls startClientEncryption() and waits for the encryption handshake
>4. If encryption fails (timeout or error), it returns false — again, no fallback
>So the original code (SmtpClient::TlsConnection on port 587) is already safe against downgrade. There is no "opportunistic" STARTTLS path in this library — it's mandatory STARTTLS.

cagnulein:
but starttls it's what it's already in place, am i missing something?

Qhilm:
I am saying, do not merge this PR, stay on STARTTLS. Why do you want to switch SMTPS (implicit TLS)?

cagnulein:
because i was trying to follow your comment :D let's close this :)

Qhilm:
> because i was trying to follow your comment :D let's close this :)

Which comment were you trying to follow? https://github.com/cagnulein/qdomyos-zwift/discussions/4572?

https://github.com/cagnulein/qdomyos-zwift/discussions/4572 is about some HTTP connection to your server for licensing it seems, it is a different topic than SMTP.

cagnulein:
I misinterpreted it :) I thought you were talking about email server. The license is OK to be in clean text since it's simply the user email.

I don't want to over engineering this :)

PR #4615, opened 2026-05-08
Fix HR PID zone mode never reaching minSpeed/maxSpeed boundaries

Qhilm:
#### Summary
- Zone-based HR PID on treadmill never reached minSpeed or maxSpeed — it stopped one 0.2 or 0.1 km/h short
- Guard conditions rejected the increment entirely if the step would overshoot, instead of clamping to the boundary
- Also fixed the minSpeed guard which checked currentSpeed + step instead of currentSpeed - step (wrong direction for the decrease case)
- Now uses std::min/std::max clamping, matching the HR Min/Max range mode which already worked correctly
#### Example
With maxSpeed = 10.0, currentSpeed = 9.9, step = 0.2:
- Before: 10.0 >= 9.9 + 0.2 → false → speed stuck at 9.9
- After: 9.9 < 10.0 → true → std::min(10.1, 10.0) → speed reaches 10.0


cagnulein:
I guess it's safe to merge, would you like to first test it on testfight?

Qhilm:
yep, I'll send you my apple ID.

Qhilm:
Thanks for providing Build 1369 on TestFlight. I tested it it works, no other side effects noted. Min and Max Speed of a row are now reached, even if the last step increment/decrement is less than0.2km/h.

Additionally, the minSpeed had a bug, whereby HR PID would end up decresing speed to "minSpeed - 0.4kmh", this is also fixed.

cagnulein:
That's great! So,let's merge it?

Qhilm:
Yes I think it can be merged. At least I couldn't see anything broken because of this.

PR #4660, opened 2026-05-25
docs: add training programme authoring guide

Qhilm:
Adds train-programs-examples/README.md — a practical guide for authoring QZ training programmes (XML format).
Covers PID HR zone pitfalls, seed speed rows, min/max bounds, ramp behaviour, repeat blocks, and recommended QZ settings. Born from real debugging sessions.
Best reviewed by reading the file directly.

### DIFF
--- a/train-programs-examples/README.md
+++ b/train-programs-examples/README.md
@@ -1,0 +1,179 @@
+# QDomyos-Zwift Training Programmes — Authoring Guide
+
+## XML Format Reference
+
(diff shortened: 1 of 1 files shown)


cagnulein:
Thanks @Qhilm 
Do you have to add something else or I can merge it?

Qhilm:
No that's it. If you agree with the content, you can merge.

Issue #4657, opened 2026-05-22
[BUG] HeartRate is not showing up on Peloton Bike but showing up on Phone App ( Apple Watch and iPhone )


Panthe1:
**Describe the bug**
The Peloton bike App does not register the Heart Rate when using the Apple Watch to track workout and heart rate. 

**To Reproduce**
1. Open App on bike
2. Open App on Phone
3. Open App on Watch
4. Notice that HR does not show up on bike but shows up on phone

**Expected behavior**
When Opening the App on the Bike / phone / Apple Watch, the Heart Rate and Bike Metrics should show up on both Bike and Apple Watch. 

**Screenshots**

**Smartphone (please complete the following information):**
 - Device: iPhone 17 Pro Max 
 - OS: iOS 26
 - Version 5

- Device Apple Watch Ultra ( Gen 1 ) 
- OS: WatchOS 26
- Version 5

- Device: Peloton Bike ( Non Swivel Screen )
- OS OpenPelo installed Nova Launcher / Grupetto / 

**Append a debug log**

[iPhone_debug-Fri_May_22_09_27_51_2026.log](https://github.com/user-attachments/files/28156514/iPhone_debug-Fri_May_22_09_27_51_2026.log)

[Peloton_debug-Fri_May_22_09_13_41_2026.log](https://github.com/user-attachments/files/28156516/Peloton_debug-Fri_May_22_09_13_41_2026.log)
  
**Additional context**
Found that if you select the HR belt on the bike as the iPhone, The connection to the phone breaks and no longer shows Bike information on the phone as well as HR information. Also, When Selecting the Phone as the HR belt in bike settings, the HR has a 50/50 chance of showing up momentarily before the tile freezes and stops working.

cagnulein:
@Panthe1 Could you please try to enable the "ignore heart rate from machinery" in the heart options in qz on the phone and enable again the heart rate belt name setting in the bike to your phone?

Panthe1:
@cagnulein I tried that and it didn't work unfortunately. Same results of loosing phone app information until resetting the bike HR belt to disabled

cagnulein:
Could you please also share the 2 logs in that context so I can compare them?

Panthe1:
[iPhone_debug-Fri_May_22_10_05_48_2026.log](https://github.com/user-attachments/files/28157315/iPhone_debug-Fri_May_22_10_05_48_2026.log)
[...]

PR #4666, opened 2026-05-28
Fix: defer BLE advertising until all GATT services are registered (iOS peripheral race condition)

Panthe1:
## Problem

When QZ runs as a BLE peripheral on iOS, `startAdvertising` is called inside
`peripheralManager(_:didAdd:)` which fires after the **first** service is added
(`heartRateService`). At that moment, the GATT database only contains 0x180D —
the FTMS (0x1826), CSC (0x1816), and Power (0x1818) services have not been
registered yet because their `add()` calls are still pending.

Any BLE central (e.g. Android running QZ as a heart rate belt client) that
connects and runs service discovery during this window sees an incomplete GATT
and misses the heart rate service. This was confirmed across multiple debug log
sessions — Android's `serviceScanDone` consistently listed only iOS system
services (0x1800, 0x1801, 0x1805, 0x180A, 0x180F) with 0x180D absent, despite
`bike_heartrate_service = true` being set.

## Root cause

`peripheralManager(_:didAdd:)` fires once per `add()` call, and the existing
code called `startAdvertising` on every invocation. Advertising therefore
started after the first service was added, before the rest were in the GATT.

## Fix

Track how many `add()` calls are in flight with a `pendingServicesToAdd`
counter. Decrement it in `didAdd` and only call `startAdvertising` when the
counter reaches zero — guaranteeing the full GATT is populated before any
central can connect and discover services.

## Testing

Confirmed via debug logs that Android's `heartratebelt::serviceScanDone` was
missing 0x180D in all sessions prior to this fix. The fix ensures all services
are registered before advertising begins, so the first connection attempt will
see the complete GATT including the heart rate service.

## Impact

iOS peripheral mode only (`#ifdef Q_OS_IOS`). No Android, Windows, Linux, or
macOS code paths are affected. The `garmin_bluetooth_compatibility` path is
unchanged. The only behavioural difference is that advertising starts slightly
later (after all services are added rather than after the first one) — which is
the correct behaviour per Apple's CoreBluetooth documentation.


Panthe1:
Additional changes included in this PR:
This commit also adds a dedicated debug log email feature, separate from the existing workout stats email.
What's added:
When Debug Log is enabled in Experimental Settings, a new "Debug Log Email" field appears directly below the toggle. At the end of each session, if this field is populated, QZ will automatically email the raw log file as a plain text attachment to that address. The workout stats email (controlled by the existing Email field in General Options) is completely unchanged.
Files changed:

src/settings.qml — added debug_email property and a conditional UI field (visible only when Debug Log is on) below the Debug Log toggle
src/qzsettings.h / src/qzsettings.cpp — added debug_email setting key and default value
src/homeform.h — declared sendDebugMail()
src/homeform.cpp — implemented sendDebugMail() which attaches the raw .log file as text/plain (no compression, which was the issue with the previous commented-out attempt), and calls it from the stop function

Why plain text instead of zip:
The original commented-out log attachment code in sendMail() noted "THE SMTP SERVER DOESN'T LIKE THE ZIP FILE". This implementation skips compression entirely and attaches the raw log as text/plain, which SMTP servers handle without issue.
Relationship to the GATT fix:
This is independent of the BLE advertising race condition fix and can be reviewed or reverted separately if needed.

cagnulein:
The comment "THE SMTP SERVER DOESN'T LIKE THE ZIP FILE" it's because sending raw debug log directly it's not possible due to the size of the debug log itself.

cagnulein:
Hi @Panthe1 I don't understand the root issue: which issue do you have?

Panthe1:
this fix is in regards to the bug that we have been working on. After reviewing the logs, I noticed that the BLE devices are advertised sequentially instead of all at once. This fix ensures that they are all advertised at the same time after they have been initialized by the service.

https://github.com/cagnulein/qdomyos-zwift/issues/4657

In regards to the debug log emails I’ve added as well, this SHOULD send a text file with the latest log if enabled over the smtp server. That would require some testing but that’s how it was designed.

cagnulein:
> this fix is in regards to the bug that we have been working on. After reviewing the logs, I noticed that the BLE devices are advertised sequentially instead of all at once. This fix ensures that they are all advertised at the same time after they have been initialized by the service.


Thanks I will check it but I don't think that's the issue. 
I will review first the logs on the #4657 

> 
> 
> 
> https://github.com/cagnulein/qdomyos-zwift/issues/4657
> 
> 
> 
> In regards to the debug log emails I’ve added as well, this SHOULD send a text file with the latest log if enabled over the smtp server. That would require some testing but that’s how it was designed. 

Yeah but as I said you can't send by email 100mb of log in plain text, that's why compression must be done ;)

LLM are fine but we need to put our heads on it too ;)

Panthe1:
Oh wait you’re right LMAO. I was thinking in MB not mb. Very different sizes. The debug files are easily 3k+ mb in size. That’s my bad haha. Haven’t done smtp shenanigans in a while so pardon my rust. 

But focusing back in on the BLE issue that I’ve found, would that be something we can try out? I’d compile the iOS app myself and test but my MacBook was abducted by my wife and I haven’t bought a new one yet :(

cagnulein:
Let me first review the logs myself later today (6am here :)) and I will report back

Panthe1:
Sounds good! Have a good day man!

cagnulein:
@Panthe1 so it's ok to don't merge this right?

PR #4668, opened 2026-05-29
fix search layout on ios

cagnulein:


cagnulein:
testeed also on android

PR #4671, opened 2026-05-31
Fix/HR PID stale zone

Qhilm:
# Fix HR PID running with stale zone target during non-HR-zone training programme rows

## Problem

When a training programme transitions from an HR-zone row to a non-HR-zone row (e.g. a `forcespeed` cool-down), the HR PID controller continues running with a stale zone target, pushing speed away from the programme's explicit value.

**Observed behaviour:** During the final 3-minute cool-down of "5 – Long Slow Distance – 45m" (speed=5.5, forcespeed=1, zoneHR unset), the treadmill speed kept increasing by +0.2 every 10 seconds, fighting the explicit 5.5 km/h target.

## Root Cause

In `homeform.cpp`, the HR PID entry condition at line 7242:
```cpp
    } else if (!settings("treadmill_pid_heart_zone").contains("Disabled") ||
               (trainProgram && currentRow().zoneHR >= 0)) {
```
During an HR-zone row (zoneHR=2), the PID writes `treadmill_pid_heart_zone = "2"` to QSettings. When transitioning to a non-HR-zone row (zoneHR=-1):

1. `fromTrainProgram = (trainProgram && currentRow().zoneHR >= 0)` → **false**
2. The code that resets the setting to "Disabled" is never reached (guarded by `fromTrainProgram`)  
3. The outer condition `setting != "Disabled"` remains **true** (stale "2")
4. PID continues targeting zone 2 → since HR drops during cool-down, it increases speed every 10s

## Fix

Add a guard that suppresses the PID when a training programme is active and the current row has no HR zone target (`zoneHR < 0`):
```cpp
    } else if ((!settings("treadmill_pid_heart_zone").contains("Disabled") &&
                !(trainProgram && currentRow().zoneHR < 0)) ||
               (trainProgram && currentRow().zoneHR >= 0)) {
```
**Logic:**
- Training programme row with HR zone → PID active (programme-driven) ✓
- Training programme row without HR zone → PID suppressed ✓  
- No training programme, manual setting enabled → PID active (user-driven) ✓
- No training programme, setting "Disabled" → PID inactive ✓

## Testing

Validated against debug log from affected session (`debug-Do__Mai_28_12_45_49_2026.log`):
- Confirms PID fired `changeSpeed` every 10 seconds during cool-down row
- Confirms stale zone 2 target was driving the increases
- Fix prevents PID block entry when `zoneHR = -1`

## Relationship to PR #4553

PR #4553 addresses race conditions during *transitions* (suppressing PID for ~10s after a speed change). It does NOT address the stale setting persisting *throughout* a non-HR-zone row. This fix is complementary.

### DIFF
--- a/src/homeform.cpp
+++ b/src/homeform.cpp
@@ -7442,6 +7442,7 @@ void homeform::update() {
 
             if (bluetoothManager->device()->deviceType() == TREADMILL &&
                 !settings.value(QZSettings::trainprogram_pid_ignore_inclination, QZSettings::default_trainprogram_pid_ignore_inclination).toBool() &&
+                !(trainProgram && trainProgram->currentRow().forcespeed) &&
                 bluetoothManager->device()->currentInclination().value() != lastInclination && lastWattage != 0) {
                 last_seconds_pid_heart_zone = seconds;
 
--- a/src/homeform.cpp
+++ b/src/homeform.cpp
@@ -1747,6 +1747,9 @@ void homeform::onTrainingProgramSpeedChanged(double speed) {
     // Record the timestamp when the training program changed speed
     // This is used by the HR PID controller to avoid race conditions
     lastTrainingProgramSpeedChange = QDateTime::currentDateTime();
+    lastWattage = 0;
+    qDebug() << QStringLiteral("Training program speed changed to") << speed
+             << QStringLiteral("- resetting inclination compensation wattage baseline");
 }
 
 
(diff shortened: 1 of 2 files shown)


cagnulein:
Would not be better to set the pid setting to disabled in rows without the pid hr? Or explicity set the pid hr in the xml to disabled in that row? I guess it could easier and cleaner

Qhilm:
> pid setting to disabled

I must have missed this, is there an xml setting for this?

cagnulein:
I mean we can do a patch to handle the zonehr tag to 0 to disable the pid.what do you think?

Qhilm:
Ah yes that works I think. Let me do that

Qhilm:
I made https://github.com/cagnulein/qdomyos-zwift/pull/4672 as alternative, let me know what you think, then I can close this.

cagnulein:
Thanks let's put this on hold

PR #4672, opened 2026-05-31
Add disablehrpid XML attribute to suppress HR PID for specific traini…

Qhilm:
# Add `disablehrpid` XML attribute to suppress HR PID for specific training programme rows

## Problem

When a training programme transitions from an HR-zone row to a non-HR-zone row (e.g. a cool-down with explicit speed), the HR PID controller can continue running with a stale zone target, pushing speed away from the programme's intended value.

## Solution

Adds a per-row `disablehrpid` XML attribute. When set to `1`, the HR PID controller is entirely suppressed for that row — giving programme authors explicit control over PID behaviour per segment.

This is a more backward compatible alternative for https://github.com/cagnulein/qdomyos-zwift/pull/4671

### Usage

```xml
<row duration="3:00" speed="5.5" forcespeed="1" disablehrpid="1" />
```

## Changes

- **`src/trainprogram.h`** — Added `bool disableHRPID = false;` field to `trainrow`
- **`src/trainprogram.cpp`** — Parse `disablehrpid` attribute (same pattern as `forcespeed`)
- **`src/homeform.cpp`** — Guard the HR PID block: skip entirely when the current row has `disableHRPID` set

## Design

- No global settings change — this is a per-row XML attribute only
- No behaviour change for existing XML files (default is `false`)
- Follows the same pattern as existing row attributes (`forcespeed`, `zonehr`, `looptimehr`)
- Explicit opt-in gives programme authors full control without surprising existing users

### DIFF
--- a/src/homeform.cpp
+++ b/src/homeform.cpp
@@ -7417,10 +7417,11 @@ void homeform::update() {
                     }
                 }
             }
-        } else if (!settings.value(QZSettings::treadmill_pid_heart_zone, QZSettings::default_treadmill_pid_heart_zone)
+        } else if (!(trainProgram && trainProgram->currentRow().disableHRPID) &&
+                  (!settings.value(QZSettings::treadmill_pid_heart_zone, QZSettings::default_treadmill_pid_heart_zone)
                         .toString()
                         .contains(QStringLiteral("Disabled")) ||
-                   (trainProgram && trainProgram->currentRow().zoneHR >= 0)) {
+                   (trainProgram && trainProgram->currentRow().zoneHR >= 0))) {
             static uint32_t last_seconds_pid_heart_zone = 0;
             static uint32_t pid_heart_zone_small_inc_counter = 0;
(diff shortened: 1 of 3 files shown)


cagnulein:
Mmm you didn't like the zonehr tag to 0 to disable this so we don't need to add a new tag?

Qhilm:
aaaaah, man, I did not understand you. Let me change that.

cagnulein:
No problem, I don't even know if the actual code with the zonehr to 0 could already do this ;) so maybe no patch required

cagnulein:
No problem, I don't even know if the actual code with the zonehr to 0 could already do this ;) so maybe no patch required

Qhilm:
ok, well, I don't need to implement anything for this, that should work already today. I will edit my XML and test tomorrow.

Qhilm:
ok, it works with zonehr=0, closing this as it's unnecessary. I will open a separate PR to update the doc for train programmes.

cagnulein:
That's great!

PR #4707, opened 2026-06-09
Custom Web Dashboard

cagnulein:


cagnulein:
@Qhilm do you like the idea?

Qhilm:
> @Qhilm do you like the idea?

I mean, it's hard to tell before testing it, but in theory this is super flexible, meaning someone can move away from the built-in tiles and define whatever one wants. Is my understanding correct?

If that's the case, then yes, it's pretty cool. The tiles are quite inflexible and on my iPhone 12 Mini, the screen real estate is not well utilized, there's a lot of white space. Hence being able to design my own dashboard would be indeed pretty cool =)

cagnulein:
> > @Qhilm do you like the idea?
> 
> I mean, it's hard to tell before testing it,

yeah i know :D

> but in theory this is super flexible, meaning someone can move away from the built-in tiles and define whatever one wants. Is my understanding correct?

exactly! I also added the guidelines for any LLM to create your own dashboard with a simple prompt

> 
> If that's the case, then yes, it's pretty cool. The tiles are quite inflexible and on my iPhone 12 Mini, the screen real estate is not well utilized, there's a lot of white space. Hence being able to design my own dashboard would be indeed pretty cool =)

yes indeed :) i make this in order that users can also share the dashboard created so we can have a sort of repository with images :)

Qhilm:
Then I think it's a very nice idea. Focus should be on elegant dynamic resizing so dashboards work on different screen sizes.

The rounded corners and the notches / Dynamic Islands / etc. of smartphones would need some form of consideration maybe, not sure how complex this is.

cagnulein:
it's something that the user should plan, i mean, from the qz side would be just a web view that talks with the back end. So all the UI customization is in the hand of the users. Let's see how it goes!

Qhilm:
I meant, let's say I'm building a dashboard. How can I say "keep 2mm or X pixels distance from the notch for this element but that element is decorative, it can can go 'under' the notch". Is that even technically possible?

Not a deal breaker, but would be nice.

cagnulein:
consider that qz is running under the dynamic island so you don't need to care about this :)

Issue #3466, opened 2025-06-03
Connection


Lmrsimpson:
**Describe the bug**

**To Reproduce**

**Expected behavior**

**Screenshots**

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**
 - Device: iPhone 14
 - OS: 18.5

[debug-Tue_Jun_3_06_18_23_2025.log](https://github.com/user-attachments/files/20572770/debug-Tue_Jun_3_06_18_23_2025.log)

**Append a debug log**

 Follow this guide https://github.com/cagnulein/qdomyos-zwift/wiki/How-do-i-get-the-debug-log-in-case-something-doesn't-work%3F
  
  
**Additional context**

cagnulein:
Enable the force ftms setting and restart qz 

Let me know 

![image](https://github.com/user-attachments/assets/6320d100-39ad-47de-a85f-fd50f98f9849)

Lmrsimpson:
This is the problem, I did that once, it still didn’t work you then told me my version of the tread should have it turned off.

cagnulein:
can you please try it again and show me a debug log with this enabled? that's why it's important to have all the information grouped on a ticket, so we can easily access to previous trials. Thanks

Lmrsimpson:
It will be tonight, I'm at work
________________________________

cagnulein:
sure, no rush!

Roberto Viola
Software engineer and open source enthusiast
http://robertoviola.cloud



Il giorno mar 3 giu 2025 alle ore 13:50 Lmrsimpson ***@***.***>
ha scritto:

Lmrsimpson:
[debug-Wed_Jun_4_18_10_22_2025.log](https://github.com/user-attachments/files/20599430/debug-Wed_Jun_4_18_10_22_2025.log)

Lmrsimpson:
Incline is not recording which is the point of me using this app

cagnulein:
Ok so it's a different thing : I thought you said it didn't work at all. 

Ok let's analyze the issue: in the log did you change the inclination from the treadmill? Also was the speed right read from qz?

Lmrsimpson:
[...]

PR #4721, opened 2026-06-12
Fix Horizon 7.0AT (JFTM) speed scaling with FTMS

cagnulein:
## Summary

Fixes #4716 — Horizon 7.0AT / JFTM 7.0AT treadmill speed control bug when using FTMS.

**Root cause:** The JFTM 7.0AT interprets FTMS Control Point speed commands (opcode `0x02`) as mph even when the treadmill is in km/h mode. QZ was sending km/h values which the device treated as mph, causing the belt to run at ~1.609× the commanded speed. The minus button also appeared to increase speed because `(X - 0.1) mph > X km/h` for any `X > 0.26`.

**Fix:** Adds an opt-in setting `horizon_treadmill_7_0_at_24` (already declared in `qzsettings.h`/`.cpp` since #3466, but previously unused) that applies a km/h → mph conversion before the FTMS Set Target Speed command. Only active when `miles_unit` is `false`, since users already in miles mode send compatible values.

This setting must be used alongside `horizon_treadmill_force_ftms = true` for the JFTM 7.0AT device.

## Changes

- **`horizontreadmill.cpp`** — `forceSpeed()` FTMS branch: multiply `requestSpeed` by `miles_conversion` (0.621371) when the setting is enabled and the app is in km/h mode
- **`settings.qml`** — activate the existing `horizon_treadmill_7_0_at_24` property (remove `// not used`), add new "Horizon 7.0AT speed fix (FTMS)" toggle under Horizon Treadmill Options

## Test plan

- [ ] Enable both "Force Using FTMS" and "Horizon 7.0AT speed fix (FTMS)" in Horizon Treadmill Options, restart app
- [ ] Verify that pressing + increases speed correctly (~0.1 km/h increments, not ×1.609)
- [ ] Verify that pressing - decreases speed correctly
- [ ] Verify that users with `miles_unit = true` are unaffected (no double conversion)
- [ ] Verify unrelated Horizon models (7.8, Paragon X, etc.) are unaffected

### DIFF
--- a/src/devices/horizontreadmill/horizontreadmill.cpp
+++ b/src/devices/horizontreadmill/horizontreadmill.cpp
@@ -1293,7 +1293,7 @@ void horizontreadmill::forceSpeed(double requestSpeed) {
                 horizon_7at_ceil = true;   // treadmill truncates mph→km/h, so ceil avoids landing one tick below
             }
         }
-        uint16_t speed_int = horizon_7at_ceil ? (uint16_t)ceil(requestSpeed * 100) : (uint16_t)round(requestSpeed * 100);
+        uint16_t speed_int = horizon_7at_ceil ? ((uint16_t)ceil(requestSpeed * 100)) + 1 : (uint16_t)round(requestSpeed * 100);
         writeS[1] = speed_int & 0xFF;
         writeS[2] = speed_int >> 8;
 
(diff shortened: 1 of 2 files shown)


cagnulein:
@stefanicai

stefanicai:
@cagnulein I don't seem to be able to push to this branch. Would you be ok to grant me access so I can try a few things over this next week? Thank you.

cagnulein:
@stefanicai I suggest to fork my branch and so you can create a PR directly. It's the easiest way

stefanicai:
ah, yeah, ok, will fork the repo, thanks.

PR #4724, opened 2026-06-12
[codex] Propagate total kcal to Apple Watch

### LINKED ISSUE #4723: [REQ]
On iPhone (17 pro, iOS 26.4.2) when I use qz on my phone+apple watch with a treadmill, the active calories count is equal to the total calories value. 
When I do not use qz (eg for outdoor walk) these two numbers differ , which makes more sense. 

These two numbers should reflect the true value, similar to other activity tracking using an Apple Watch. 
The missing term - resting calories can be calculated from the users data - Weight, Height, Age etc. 

N/Aimage

**Additional context**
See screenshot for current status.

cagnulein:
## Summary

Fixes #4723 by carrying total kcal through the iPhone/iPad to Apple Watch data path.

## Changes

- Include `totalKcal` in the iPhone WatchConnectivity reply used by the watch app.
- Parse `TOTALKCAL=` from the iPad relay message into `WatchKitConnection.totalKcal`.
- Copy the received watch-side `totalKcal` into `WorkoutTracking.totalKcal` so the existing HealthKit workout save logic can write `totalEnergyBurned` separately from active kcal.

## Validation

- Ran `git diff --check` on the modified files.
- iOS/watchOS build was not run in this Linux environment.


cagnulein:
Follow-up added for the Live Activity heart-rate issue shown in the screenshots.

Root cause: `workoutTrackingUpdate` used a single `heartRate` argument for both HealthKit workout samples and the Live Activity. When HR came from HealthKit/Apple Watch, QZ intentionally passed `0` to avoid writing a duplicate HR sample back into HealthKit; the lock-screen Live Activity then displayed that same `0`, while Dynamic Island compact could still show the configured compact metric value.

Fix: split the values so HealthKit workout tracking can still receive `0` for watch-sourced HR, while Live Activity receives the actual current HR.

PR #4729, opened 2026-06-14
Queue Domyos elliptical force commands

cagnulein:
## Summary
- keep Domyos elliptical resistance and inclination force writes separated by update cycle
- preserve a pending inclination request when a resistance force command is written first
- leave the existing inclination protocol/value mapping unchanged

## Testing
- Not run (requested not to compile)


cagnulein:
Mail from ed manle.

PR #4740, opened 2026-06-17
Fix FTMS trainer slow REQUEST_CONTROL response

### LINKED ISSUE #4739: [BUG] virtual shifter wont change resistance of magene t600
virtual shifter wont change resistance of magene t600. Ive attached the log

[debug-Tue_Jun_16_23_35_17_2026.log](https://github.com/user-attachments/files/29009112/debug-Tue_Jun_16_23_35_17_2026.log)

cagnulein:
## Summary
Some trainers (e.g. Magene T600) are slow to acknowledge the REQUEST_CONTROL command. The old code would send START_RESUME immediately after, causing the trainer to ignore it. This fix waits for the acknowledgment (up to 2 seconds) before sending START_RESUME.

## Changes
- Made the init sequence asynchronous (no nested event loop)
- Move START_RESUME write from BLE receive handler to the update() pump
- Add a 2-second fallback timeout if the trainer doesn't respond

Fixes #4739

### DIFF
--- a/src/devices/ftmsbike/ftmsbike.cpp
+++ b/src/devices/ftmsbike/ftmsbike.cpp
@@ -146,8 +146,31 @@ bool ftmsbike::writeCharacteristic(uint8_t *data, uint8_t data_len, const QStrin
     return true;
 }
 
+void ftmsbike::sendStartSimulation() {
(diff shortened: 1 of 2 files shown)


cagnulein:
probably not needed

PR #4751, opened 2026-06-25
Fix 'Other folders' button inaccessible in landscape mode (#3769)

cagnulein:
## Summary

Fixes #3769 — the "Other folders" button was pushed off-screen and inaccessible when the device was in landscape orientation.

**Root cause:** In Qt Quick, children of `ColumnLayout`/`RowLayout` must use `Layout.*` properties for sizing. When `anchors.fill: parent` or `anchors.bottom: parent.bottom` are used instead, the anchor geometry overrides the layout engine's measurements, causing the content area to expand and bottom buttons to be clipped or pushed off-screen in landscape.

**Files fixed:**
- `src/GPXList.qml` — used on Android/iOS (GPX route list, "Other folders" button)
- `src/TrainingProgramsList.qml` — used on desktop (training program list, all 3 bottom buttons)

**Changes:**
- Replaced `anchors.fill: parent` / `anchors.top: parent.top` with `Layout.fillWidth: true` + `Layout.fillHeight: true` on the content `RowLayout`
- Replaced `anchors.fill: parent` with `Layout.fillHeight: true` on the inner `ColumnLayout`
- Replaced `Layout.preferredHeight: parent.height` with `Layout.fillHeight: true` on `ListView` and `ScrollView`
- Replaced `height: 50; width: parent.width; anchors { bottom: parent.bottom }` with `Layout.fillWidth: true; Layout.preferredHeight: 50` on bottom buttons

### GPXList — Landscape (0°)
gpxlist_landscape

### GPXList — Portrait (90°)
gpxlist_portrait

### GPXList — Landscape Inverted (180°)
gpxlist_landscape_inverted

### GPXList — Portrait Inverted (270°)
gpxlist_portrait_inverted

### TrainingProgramsList — Landscape (from tablet, JS variant used on Android)
trainlist_landscape

## Test plan

- [x] Tested on Android emulator (tablet 2560×1600, API 34) in all 4 rotations
- [x] "Other folders" button visible and accessible in every orientation
- [x] GPXList shows split-panel layout (file list left, map preview right) correctly in all rotations
- [x] TrainingProgramsList buttons ("Start Workout", "Delete", "Other folders") all visible

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #4752, opened 2026-06-25
Fix UI clipping in all 4 orientations on Android devices with display cutout

cagnulein:
## Problem

On Android devices with a punch-hole or notch display (e.g. Pixel 6), rotating to **landscape or inverted landscape** showed a white strip (~49dp) on the notch side, and the toolbar had wasted space at the top.

Root causes:

1. **White strip**: Qt 5.15 offsets its rendering scene by `leftInset` to avoid the display cutout. Without `setDecorFitsSystemWindows(false)`, the system shrank the app content area around the cutout, leaving the raw Android window background (white) visible in that region.

2. **Toolbar top space**: The same window-shrinking caused `topPadding` to be double-counted — Qt already positioned content below the status bar, and our QML padding added more space on top.

3. **`getTopPadding()`/`getBottomPadding()` axis swap**: The original code returned `leftInset` as top padding in landscape (wrong). `getSystemWindowInsets()` already returns orientation-aware values.

4. **`LAYOUT_IN_DISPLAY_CUTOUT_MODE_SHORT_EDGES`**: Only extends into the cutout on the physical short sides of the device. On a phone, inverted landscape has the notch on the long side, so it was clipped.

## Fix

- **`setDecorFitsSystemWindows(false)`** (API 30+): Prevents the system from shrinking the content area around insets. The app now renders edge-to-edge into ALL screen areas and handles insets itself via `onApplyWindowInsets`.
- **`LAYOUT_IN_DISPLAY_CUTOUT_MODE_ALWAYS`** (API 30+): Extends the window into the display cutout on all 4 rotation variants.
- **Black window background**: `setBackgroundDrawable(new ColorDrawable(Color.BLACK))` ensures any area outside Qt's rendering surface shows black rather than white.
- **`getTopPadding()`**: Always returns `AndroidStatusBar.height` (orientation-aware, no manual axis swap).
- **`getBottomPadding()`**: Always returns `AndroidStatusBar.navigationBarHeight`.

## Screenshots — Pixel 6 (API 34), all 4 orientations after fix

| Portrait (rot0) | Landscape 90° (rot1) |
|---|---|
| portrait | landscape |

| Inverted Portrait (rot2) | Inverted Landscape 270° (rot3) — **previously clipped + white strip** |
|---|---|
| inv_portrait | inv_landscape |

- No white strip on any side in any rotation
- Toolbar extends edge-to-edge, compact height
- Icons correctly padded away from the notch area via QML `leftPadding`/`rightPadding`

## Test plan

- [x] Pixel 6 emulator (API 34) — all 4 rotations
- [ ] Physical device with notch/punch-hole — rotate through all 4 orientations
- [ ] Device without notch — verify no regression (`setDecorFitsSystemWindows` only applies on API 30+; padding functions return 0 when `apiLevel < 31`)


cagnulein:
### Pixel 6 (punch-hole camera, API 34)

| Portrait | Landscape | Inverted Landscape |
|---|---|---|
|  |  |  |

cagnulein:
### Pixel 3a (no notch, API 34)

| Portrait | Landscape | Inverted Landscape |
|---|---|---|
|  |  |  |

cagnulein:
### 10" Tablet (API 34)

| Portrait | Landscape | Inverted Landscape |
|---|---|---|
|  |  |  |

cagnulein:
### Pixel 6 (punch-hole camera, API 34) — after toolbar fix

| Portrait | Landscape | Inverted Landscape |
|---|---|---|
|  |  |  |

cagnulein:
### Pixel 3a (no notch, API 34) — after toolbar fix

| Portrait | Landscape | Inverted Landscape |
|---|---|---|
|  |  |  |

cagnulein:
### Pixel 3a API 28 (Android 9 — legacy path, no edge-to-edge)

On API < 30 the Java code does not call `setDecorFitsSystemWindows` and all QML padding functions return 0, so Qt handles insets internally as before. No regression.

| Portrait | Landscape | Inverted Landscape |
|---|---|---|
|  |  |  |

cagnulein:
## Emulator test results (post-fix)

The deferred SurfaceView offset detection fix is now working on **all 4 orientations** on the emulator AND real devices.

### How it works
Qt 5.15 on emulators internally shifts its rendering viewport by the top inset (~24dp) in landscape, even after `setDecorFitsSystemWindows(false)`. On real devices this shift does NOT happen.

The fix: after each layout pass, we find Qt's SurfaceView in the view hierarchy, measure its Y position within the window, and subtract that offset from the reported top inset before passing it to QML. This way `topPadding` is never double-counted.

- **Real device**: SurfaceView at y=0 → adjustment=0 → topPadding=24dp ✓  
- **Emulator landscape**: SurfaceView at y=24dp (internal shift) → adjustment=24dp → topPadding=0dp ✓

### Emulator-5556 (Pixel 6, API 34) results
| Orientation | Result |
|---|---|
| Portrait | ✅ Toolbar correctly below status bar |
| Landscape | ✅ No extra space, toolbar correct |
| Inverted landscape | ✅ No extra space, toolbar correct |

Previously real Pixel 8a was broken by landscape-specific workarounds. This approach auto-detects the behavior, so both emulator and real device work with the same code.

cagnulein:
## Update: PR #4751 changes consolidated here

The GPXList.qml and TrainingProgramsList.qml fixes from PR #4751 are already part of this branch (commit `a988ec024`). PR #4751 has been closed.

**What that fix does:** Replaces `anchors.fill`/`anchors.bottom` with `Layout.fillWidth`/`Layout.fillHeight` on all children of the root `ColumnLayout` in these two screens. This keeps the "Other folders" button visible both when starting in landscape AND when rotating while the app is already running.

Issue #1187, opened 2023-01-20
[REQ]Sound reminders for training programs


grube386:
**Is your feature request related to a problem? Please describe.**
There is no way to know that the program changed from one row to another unless you are always following the timing of each in a dedicated tile. There is also no way to know if you are above or below the targets (target zones) in the program. 

**Describe the solution you'd like**
A sound that informs the user that the program has changed or that they are out of the targets.
To cover all three scenarios three different sounds would be needed:
1. One sound to notify the user that the program has started a new row
2. One sound to notify the user that they are under the zone set for x amount of seconds (for example that they have not hit the target for 30/60/90s or % of the time in the row)
3. One sound to notify the user that they are over the zone set for x amount of seconds (for example that they are over the target for 30/60/90s or % of the time in the row) 

MVP would for sure be the first point. The other ones would be nice to have.

**Describe alternatives you've considered**
I have considered adding additional "alert" rows to the program, which would indicate if the target zones went up or down. For example, if the HR Zone goes from 3 to 2 from one row to another, I would add a row in between where the resistance would go to 1 for 30s, to remind me that something will change without having to follow the screen the whole time. And if it goes from 2 to 3, I would add a row in between with 15 (max) resistance for 30s.

cagnulein:
@grube386 can you help me finding a sound/s to implement? I don't have a human touch to pick a right one :D

grube386:
Not sure if I do as well :)

Here are a few options. The first ones would be my choice.
For the change of row: 
https://freesound.org/people/drooler/sounds/660862/
https://freesound.org/people/tim.kahn/sounds/91926/
https://freesound.org/people/robni7/sounds/174027/

For over doing it:
https://freesound.org/people/deleted_user_9618385/sounds/638888/
https://freesound.org/people/m_cel/sounds/507906/
https://freesound.org/people/greatsoundstube/sounds/629312/

For not working hard enough: 
https://freesound.org/people/Ultranova105/sounds/136755/ (shortened and looped 2x or 3x) 
Any of the overdoing it can also be copied here, since they are warnings in any case

cagnulein:
I like

For the change of row:
https://freesound.org/people/drooler/sounds/660862/

For over doing it:
https://freesound.org/people/greatsoundstube/sounds/629312/

For not working hard enough:
https://freesound.org/people/Ultranova105/sounds/136755/ (shortened and looped 2x or 3x)

ok i will try to implement this next week

[...]

PR #4757, opened 2026-06-26
Training Program Sound

cagnulein:
@13agao done! there is a new setting under the training program options for this!


cagnulein:
#1187

PR #4767, opened 2026-07-01
Write HR from another QZ instance (TCP/mDNS) to Apple Health on iPad

cagnulein:
## Summary

When QZ on iPad receives heart rate from another QZ instance via the \`_qz_iphone._tcp.local.\` TCP socket, the HR value was correctly received and the \`heartRate()\` signal was emitted — but it never reached Apple Health.

The fix calls the existing \`lockscreen::setHeartRate()\` directly in both TCP \`readyRead\` handlers (serviceAdded and serviceUpdated). This reuses the same path already used by Bluetooth HR belts:

\`lockscreen::setHeartRate\` → \`[healthkit setHeartRateWithHeartRate:]\` → \`AppDelegate::setHeartRate\` → \`WorkoutTracking.shared.setBluetoothHeartRate\` → Apple Health

No new API surface — just two extra lines in the TCP handlers, guarded by \`#ifndef IO_UNDER_QT\`.

## Files changed

- \`src/homeform.cpp\` — call \`ls.setHeartRate((unsigned char)hr)\` in both mDNS TCP readyRead handlers

## Test plan

- [ ] Run QZ on iPad (receiver) and QZ on another device (sender) with a HR belt connected to the sender
- [ ] Confirm HR appears in Apple Health after a workout session on the iPad

### DIFF
--- a/src/homeform.cpp
+++ b/src/homeform.cpp
@@ -1050,7 +1050,7 @@ homeform::homeform(QQmlApplicationEngine *engine, bluetooth *bl) {
                                                  emit homeform::singleton()->heartRate(hr);
 #ifndef IO_UNDER_QT
                                                  lockscreen ls;
-                                                 ls.setWorkoutTrackingHeartRate((unsigned char)hr);
+                                                 ls.setHeartRate((unsigned char)hr);
 #endif
                                              }
                                          }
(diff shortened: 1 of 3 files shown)


cagnulein:
mail from Posteo

PR #4774, opened 2026-07-03
Treadmill Live Charts (speed/inclination + heart)

### LINKED ISSUE #2955: [REQ] Treadmill Live Charts
Following our discussion on the Kinni forum could you please enable Live Charts for treadmills.

Thanks.

Ray.

cagnulein:
## Summary

Closes #2955. The chart icon and live webview chart infrastructure (added in `b71321f30`, "Treadmill Live Charts #2955") were built but never actually wired up:
- `chartIconVisible` only checked for power/HR-zone target workouts, so the chart icon never appeared for treadmill target-speed/inclination workouts.
- `ChartFooterInnerJS.qml` always loaded the bike/power chart (`chartlive.htm`), never the treadmill one (`treadmillchartlive.htm`).

## Changes

- **`trainprogram.h`**: `speedInclinationTargetWorkout()` detects rows with a target speed/inclination, separate from the existing power/HR detection.
- **`homeform.{h,cpp}`**: new `chartTreadmillMode` property. A treadmill running a speed/inclination workout — or a pure heart-rate-triggered one (Garmin-style `HRabove`/`HRbelow` open-ended rows, same mechanism used elsewhere for Garmin workouts) — now gets the treadmill chart instead of falling through to the irrelevant power chart.
- **`ChartFooterInnerJS.qml`**: picks `treadmillchartlive.htm` vs `chartlive.htm` based on `chartTreadmillMode`.
- **`devices/bluetooth.{h,cpp}`**: added a 15s fallback timer so Fake Device/Fake Treadmill/etc. still activate when the platform's Bluetooth discovery agent never emits `finished()` (e.g. Android containers/emulators without a functional Bluetooth adapter — this blocked on-device testing entirely until fixed).
- **`AGENTS.md`**: documented the Android on-device debug build gotchas hit while testing this (ABI override being ignored, missing translation artifacts in fresh worktrees, `androiddeployqt` silently failing to bundle Qt runtime libs / QML RCC bundle, Java 17 requirement for Gradle, no Bluetooth / no `run-as` on Waydroid) — should save significant time for future on-device debugging.

## Test plan

- [x] Built and installed a debug APK on a Waydroid Android device (x86_64)
- [x] Enabled Fake Treadmill in Experimental Settings
- [x] Built a 4-interval test workout via the in-app Workout Editor with varying target speed (9.5→10→8→12 km/h) and inclination (1→2.5→5→6%)
- [x] Started the workout, confirmed the chart icon appears and the treadmill chart (not the power chart) renders, with live speed/inclination zones and the heart-rate panel
- [x] Verified via logcat that `target_speed`/`target_inclination`/`target_heart_above`/`target_heart_below` are populated in the live "workout" payload consumed by the chart JS

## Screenshots

See PR comments below.

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>


cagnulein:
## Verification screenshots (real device, Android/Waydroid)

**App boots correctly** after fixing an unrelated pre-existing packaging bug (androiddeployqt wasn't bundling the Qt QML plugin RCC on this toolchain, so QQmlApplicationEngine failed to load any QML module before this fix):

Home screen working

**Fake Treadmill connected** (Experimental Settings → Fake Treadmill), after the Bluetooth discovery fallback timer fix so it activates even without a functional adapter:

Fake treadmill connected

**Test workout loaded** — a 4-interval treadmill workout with varying target speed/inclination, built via the in-app Workout Editor:

Workout loaded

**Workout running**, live tiles following the training program targets (interval 2: 10 km/h / 2.5% incline):

Workout running

**Treadmill live chart** — this is the actual feature: tapping the chart icon now opens the treadmill speed/inclination chart (with speed-zone coloring) instead of the irrelevant bike/power chart, alongside the existing heart-rate panel:

Treadmill live chart

**Chart later in the workout** (interval 4, 12 km/h / 6% incline), confirming the chart keeps updating as the training program progresses through rows:

Treadmill live chart progressed

cagnulein:
## Follow-up fix: target-line draw order and gap handling

After posting the screenshots above, I double-checked two details in `dotreadmillchartlive.js` that weren't quite right:

1. **Draw order**: Chart.js paints datasets in array order, so the dashed Target Speed/Target Incline lines (added after the actual Speed/Incline datasets) were rendered **on top of** the actual measured lines. With a Fake Treadmill (no acceleration lag — actual speed tracks target instantly) the dashed target line visually covered the solid, zone-colored actual line whenever they coincided. Fixed by reordering datasets so the actual lines always paint last (on top).

2. **Gap bridging across untargeted rows**: when a row has no speed/inclination target (e.g. a heart-rate-triggered row, target left at -1/-200), the code was skipping the point entirely instead of pushing `null`. Chart.js was then connecting the last known target value straight across to the next row that *does* have a target — drawing a misleading diagonal "bridge" through a period where no target applied at all. Fixed by pushing `null` explicitly so Chart.js breaks the line there (default `spanGaps: false`).

### Verification

Built a workout via the in-app Workout Editor: Interval 1 (speed 9.5 / incline 1) → Interval 2 (HR Zone only, no speed/incline target) → Interval 3 (speed 12 / incline 5). Ran it and reopened the chart:

Gap test full

Zoomed in on the transition — no line before the workout starts, a short bump for Interval 1's target, a clean **break** during Interval 2 (HR-only, no target), then a sharp step up for Interval 3 (no diagonal bridge):

Gap test zoom

cagnulein:
## Follow-up fix: chart was clipping vertically, and colors didn't match the rest of the app

Two more issues found:

1. **Vertical clipping**: `divcanvas`/`divcanvasheart` used `height:100vh` together with a viewport meta tag forcing `height=device-height` — so the chart sized itself for the *full device screen height* instead of the actual (much shorter) footer area Home.qml allocates for it. Chart.js computed its aspect ratio from that inflated height, so the canvas rendered taller than the visible container and got clipped: only the bottom of the y-axis (0–6) was ever visible, even once actual speed reached 12+. Fixed by using `height:100%` chained through `html`/`body`/`table`/`td` (matching the real container) and `maintainAspectRatio:false` instead of a fixed aspect ratio, so Chart.js always fills the container's actual current size. Applied the same fix to the heart panel (shared between the bike and treadmill charts).

2. **Colors**: the speed line was using a fabricated 6-zone color scheme (grey/green/gold/orange/darkorange/red at made-up `[6,8,10,12,14,16]` km/h thresholds) copy-pasted from the bike chart's FTP-zone pattern — but QZ has no actual user-configurable "speed zone" setting the way it does for power or heart rate, so those thresholds and colors meant nothing. Replaced with the same flat colors the Workout Editor already uses for Speed (`#42a5f5`) and Incline (`#26c6da`), for visual consistency with what the user already sees when building/previewing the workout.

Before/after — full chart, no longer clipped, axis titles colored to match their line:

No clip, full

Zoomed on the target/actual transition — axis now goes up to 14, dashed target lines and solid actual lines in the Workout Editor's blue/cyan instead of the old multi-color zone bands:

No clip, zoom

cagnulein:
## Follow-up: use the user's actual race pace zones for speed coloring

You were right that the colors could have made sense — I'd missed that Settings -> Training Program Options already has real, user-configurable reference paces: 1 Mile, 5K, 10K, Half Marathon, and Marathon pace (`pacef_1mile`/`pacef_5km`/`pacef_10km`/`pacef_halfmarathon`/`pacef_marathon`). These are the exact same values `zwiftworkout::speedFromPace()` already uses to convert a Zwift running workout's %pace target into an actual km/h — so they're a legitimate, personalized "speed zone" system, just like FTP is for power.

Replaced the flat blue speed line with zone coloring derived from these 5 paces (sorted slowest-marathon to fastest-1-mile), with matching background zone bands and axis labels — same grey-to-red escalation pattern as the bike/heart-rate charts:

Pace zones full

Zoomed in — axis labels now read e.g. "10 (Marathon)" / "11 (Half Marathon)" instead of generic numbers, with the corresponding colored zone bands:

Pace zones zoom

Incline keeps the flat Workout Editor color (`#26c6da`) since there's no equivalent zone concept for it.

PR #4801, opened 2026-07-11
[codex] Add Yesoul C1EV Peloton resistance mapping

cagnulein:
## Summary
- Detect `YS_C1EV_` devices as FTMS bikes.
- Mark the C1EV model in `ftmsbike` and keep its max resistance at 100.
- Use the bike's native FTMS resistance as a linear Peloton resistance source, still honoring `peloton_gain` and `peloton_offset`.

## Root Cause
The Yesoul C1EV advertises as `YS_C1EV_...`, which was not covered by the existing Yesoul/FTMS detection patterns. When manually configured as FTMS, QZ parsed native resistance but calculated Peloton resistance from cadence/watts instead of the bike's 1-100 resistance scale.

## Validation
- `git diff --check`
- Verified `YS_C1EV_00928` does not match the existing `YS_C1_` pattern and does match the new `YS_C1EV_` pattern.


cagnulein:
Mail from Lesley Come

PR #4803, opened 2026-07-12
Fix treadmill pause watts and Stryd steps

### LINKED ISSUE #4480: [BUG] Treadmill Auto-adjust Speed by Power
**Issue:**
When "Settings > Training Program Options > Treadmill Auto adjust speed by power" is enabled, it interferes with interval transitions.

**Behavior:**
At the start of each interval, the treadmill receives speed and incline commands, but the auto-adjust logic immediately overrides the speed to maintain constant power based on the new incline. This prevents the workout from starting with the correct predefined speed/incline targets.

**Request:**
Can the auto-adjust logic be delayed or bypassed during interval transitions to allow the program's target speed and incline to be reached first?

**Smartphone:**
 - Device: Poco X7 Pro
 - OS: Android 16
 - Version: QZ 2.20.29

**Treadmill:**
- Domyos Run 500

**Debug Log:**
[debug-Mon_Mar_30_22_01_53_2026.log](https://github.com/user-attachments/files/26360427/debug-Mon_Mar_30_22_01_53_2026.log)

**Workout used for testing**
[Walk_6_16.xml](https://github.com/user-attachments/files/26360448/Walk_6_16.xml)

If I disable "Treadmill Auto adjust speed by power" the loaded workout works as expected.

cagnulein:
## Summary

Refs #4480.

This PR fixes treadmill-side issues found while analyzing the Stryd logs attached to issue #4480:

- keep calculated watts populated while paused when `instant_power_on_pause` is enabled
- show raw watts in UI/templates while paused when `instant_power_on_pause` is enabled, instead of using 3s/5s averaged watts that can be empty while metrics are paused
- update treadmill step calculation when cadence arrives from an external sensor such as Stryd
- add a Power Sensor Options setting for the Stryd/treadmill speed correction threshold, defaulting to 20%

## Root cause

`treadmill::update_metrics()` had its own paused handling and did not honor `instant_power_on_pause`, so watts could be reset to zero while paused even though the global setting was enabled.

The template watt value uses `wattsMetricforUI()`. With `power_avg_3s` enabled, that path reads the harmonic average buffer. While paused, `metric::setValue()` updates the raw value but does not populate the average buffers, so templates could show `watts: 0` even though the current watt value and `kgwatts` were nonzero.

The external cadence path `treadmill::cadenceSensor()` only updated `Cadence`; unlike other cadence update paths, it did not call `evaluateStepCount()`, so Stryd cadence did not feed the step counter.

The Stryd/treadmill speed correction threshold was hardcoded at 20%, which made it hard to test whether a wider threshold improves HR PID behavior when Stryd speed and treadmill raw speed differ.

## Validation

- inspected the attached issue logs and confirmed the relevant settings and Stryd data path
- ran `git diff --check`

No runtime device test was run locally.


cagnulein:
Superseded by #5091, which is based on the current master and contains the remaining changes from this PR. The Stryd Indoor Mode portion is already present in master via #4861.

PR #4813, opened 2026-07-13
fix: wahoo setSimGrade sent before setSimMode (KICKR CORE race condition)

cagnulein:
## Problem

On KICKR CORE (and similar Wahoo devices), `setSimGrade` was being sent to the trainer **before** `setSimMode` had ever been sent.

### Root cause

`stateChanged()` subscribes to **all** notifiable/indicatable characteristics across all BLE services. On a KICKR CORE this amounts to 17 descriptor writes (`notificationSubscribed = 17`). The `descriptorWritten` callbacks arrive slowly from the BLE stack — in the reported log only 9 of the 17 had arrived by the time Zwift sent its first `indoor bike simulation parameters` FTMS command (~30 seconds after connection).

Since `notificationSubscribed` never reached 0, `initRequest` was never set to `true`, so the `update()` init block (which sends `unlock` → `setSimMode` → `setWheelCircumference`) never ran.

Meanwhile Zwift's FTMS write flows through a direct signal/slot path:  
`CharacteristicWriteProcessor2AD9 → changeSlope → inclinationChanged → setSimGrade`  
…which bypasses `initRequest` entirely, sending `setSimGrade` to a trainer that had never received `setSimMode`.

## Fix

Subscribe **only to `CyclingPowerMeasurement`** during `stateChanged()` — at most 1 descriptor write — so `descriptorWritten` fires almost immediately and the init sequence (including `setSimMode`) completes before Zwift can send the first slope command.

A fallback is also added: if no `CyclingPowerMeasurement` subscription was established at all, `initRequest` is set directly from `stateChanged()` so the init always runs.

This is a minimal, targeted fix. A previous attempt ([PR #4530](https://github.com/cagnulein/qdomyos-zwift/pull/4530), closed as wontfix) addressed the same issue with additional unrelated changes; this PR contains only the essential change.

## Test plan

- [ ] Connect KICKR CORE and verify `setSimMode` appears in the log before the first `setSimGrade`
- [ ] Ride in Zwift: resistance/slope changes should work correctly from the start
- [ ] Verify no regression for KICKR SNAP and KICKR BIKE (ERG mode, gear mode)

### DIFF
--- a/src/devices/sportsplusrower/sportsplusrower.cpp
+++ b/src/devices/sportsplusrower/sportsplusrower.cpp
@@ -146,8 +146,10 @@ void sportsplusrower::characteristicChanged(const QLowEnergyCharacteristic &char
     }
 
     if (!firstCharChanged) {
-        Distance +=
-            ((Speed.value() / 3600.0) / (1000.0 / (lastTimeCharChanged.msecsTo(now))));
+        double elapsedMs = lastTimeCharChanged.msecsTo(now);
+        Distance += ((Speed.value() / 3600.0) / (1000.0 / elapsedMs));
+        // Accumulate stroke count from cadence since the device doesn't send it directly
+        StrokesCount += (Cadence.value() / 60.0) * (elapsedMs / 1000.0);
     }
 
     lastTimeCharChanged = now;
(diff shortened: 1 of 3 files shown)


cagnulein:
Closing in favour of a clean branch based on origin/master with only the wahookickrsnapbike change.

PR #4834, opened 2026-07-21
feat: add Wahoo KickRun treadmill driver

cagnulein:
## Summary

- Adds dedicated `wahookickruntreadmill` driver for the Wahoo KickRun treadmill (previously routed through `horizonTreadmill` which only supports inclination)
- Protocol reverse-engineered from HCI btsnoop log: proprietary BLE over a single Write-Without-Response + Notify characteristic (handle 0x0023)
- Implements the full init sequence (`01 10 12 18 80 86 f0`), device-ID handshake (`87 + [4-byte ID]`), and workout-mode unlock (`19 00`)
- Speed command: `02 [μm/s LE24] 00 [flag]` where flag = `0xFF` for first command of session, `0x0A` for subsequent; confirmed against captured traces at 2.4, 4.8 and 5.0 km/h
- Handles the hardware challenge-response safety interlock: treadmill sends `FD E0 01 [XX YY]` after user presses the physical paddle, driver immediately echoes `E0 [XX YY]` to confirm
- Telemetry parsed from characteristic 0x0026: `FF 01 [tgt_b0 b1 b2] 00 …` — bytes 2–4 are the commanded speed in μm/s LE24 (confirmed from trace)
- Note: the physical paddle push before each speed change is enforced in treadmill firmware and **cannot** be bypassed via BLE configuration commands (all init commands are read-only queries or one-shot unlocks)

## Test plan

- [ ] Build compiles on Android/iOS/Linux/macOS
- [ ] Device named `KICKR RUN *` is picked up by the new driver (not routed to `horizonTreadmill`)
- [ ] Init sequence completes and `connectedAndDiscovered` fires
- [ ] Speed commands reach the treadmill after paddle push
- [ ] Telemetry updates `Speed` metric correctly
- [ ] Reconnect after BLE drop resets state and re-inits

### DIFF
--- a/src/devices/bluetooth.cpp
+++ b/src/devices/bluetooth.cpp
@@ -1624,6 +1624,23 @@ void bluetooth::deviceDiscovered(const QBluetoothDeviceInfo &device) {
                     emit searchingStop();
                 }
                 this->signalBluetoothDeviceConnected(lifefitnessTreadmill);
+            } else if (b.name().toUpper().startsWith(QStringLiteral("KICKR RUN")) &&
+                       !wahooKickRunTreadmill && filter) {
+                this->setLastBluetoothDevice(b);
+                this->stopDiscovery();
+                wahooKickRunTreadmill = new wahookickruntreadmill(noWriteResistance, noHeartService);
+#if !defined(Q_OS_ANDROID) && !defined(Q_OS_IOS)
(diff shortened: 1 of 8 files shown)


cagnulein:
Stephan Ben. by email

PR #4849, opened 2026-07-25
Add Moving Time and Clock tiles to companion floating window

cagnulein:
## Summary
- The companion AI floating window (`src/inner_templates/floating/floating.htm`, used to mirror QZ metrics onto a second phone e.g. with MyWhoosh) only rendered a subset of the tiles available in the main app's tile picker.
- `tile_moving_time_enabled`/`tile_moving_time_order` and `tile_datetime_enabled`/`tile_datetime_order` already existed as settings but had no corresponding row/handling in the floating page, so those two tiles never appeared on the mirrored screen even when enabled on the main device.
- This adds the missing rows and wiring:
  - Moving Time: new row, fetched from the existing `moving_s`/`moving_m`/`moving_h` fields already present in the `workout` websocket payload (`templateinfosenderbuilder.cpp`), respects enable/order settings.
  - Clock: new row, rendered from the local device clock (`Date()`) and refreshed every second client-side, respects enable/order settings.
- No C++/backend changes needed — `moving_s/m/h` were already being sent, and settings are fetched generically by key.

## Test plan
- [x] Opened `floating.htm` in a browser and visually verified the two new rows render, including a self-updating Clock.
- [ ] Manual test on device: enable "Moving Time" and "Clock" tiles in QZ, open the companion app on a second device, confirm both tiles show and update, and confirm they disappear when disabled in settings.

### DIFF
--- a/src/inner_templates/floating/floating.htm
+++ b/src/inner_templates/floating/floating.htm
@@ -215,6 +215,16 @@
       <td style="text-align: left">NEXT</td>
       <td class="nextrow-value values" colspan="5"><b></b></td>
     </tr>
+    <tr class="movingtime" sort-order="18">
+      <td class="icon">⏲️</td>
+      <td style="text-align: left">MOVING T.</td>
+      <td class="movingtime-value values" colspan="5"><b>0:00:00</b></td>
+    </tr>
+    <tr class="datetime" sort-order="19">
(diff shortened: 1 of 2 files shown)


cagnulein:
https://github.com/cagnulein/qdomyos-zwift/actions/runs/30171048813

Issue #4785, opened 2026-07-06
Stages SB20: surface the handlebar shifter buttons for in-app control (target power ± / Peloton offset ±)


cauldnz:
Follow-up to #1649 — and thanks @cagnulein for the invite to open a PR. Splitting the SB20 handlebar-button idea into its own issue to track design + implementation.

## Background
The Stages SB20 has **6 handlebar buttons** (a 3-button cluster each side). The bike exposes them over BLE, but qz doesn't use them today. qz already binds the SB20 to the **`ftmsbike`** driver (it advertises as `Stages Bike <NNNN>`), so this is an additive feature on that driver — no new device class.

I've fully decoded the button channel from passive BLE sniffing of my own SB20:
- **Vendor service** `0c46be5f-9c22-48ff-ae0e-c6eae1a2f4e5`, **notify characteristic** `0c46be60-9c22-48ff-ae0e-c6eae1a2f4e5`.
- Each press streams a short burst; payload is `<type:u8> 00 <bitmask:u16 LE>`, where the bitmask is a **one-hot id of which button** (only bits 0–5 are ever set):

  | Button | bit |
  |---|---|
  | LEFT up / down / 3rd | 0 / 1 / 2 |
  | RIGHT up / down / 3rd | 3 / 4 / 5 |

- Frame model per press: `01` = *held* (streams continuously while held), `03` = *commit*, `04`/`08` = *terminator*. It's **stateless** — the shifter reports *which button*, with no gear index (the consumer owns any gear concept). So a consumer must **debounce** the held-stream into one logical press.

(Happy to share the raw captures + a fuller protocol write-up.)

## Proposed feature
Let SB20 riders use the spare handlebar buttons to control qz mid-ride — starting with the two I reach for most on Peloton Power Zone rides:
1. **Target power ±** — nudge the erg target watts.
2. **Peloton offset ±** — nudge the workout timeline so the zone changes line up with the video.

## Plan
1. **Subscribe** to `0c46be60` during `ftmsbike` service discovery, gated behind a new `sb20_buttons_enabled` setting so it only affects SB20 users.
2. **Decode + debounce** the burst into single press events (key off the `03` commit frame).
[...]

PR #4850, opened 2026-07-26
Stages SB20 handlebar shifter buttons -> in-app controls (#4785)

cauldnz:
Implements #4785. The SB20 exposes its 6 handlebar buttons on a BLE vendor
characteristic; qz already binds the bike to `ftmsbike`, so this is additive on
that driver — no new device class.

Behind `sb20_buttons_enabled` (default on; it can only trigger on a bike that
actually exposes the vendor characteristic).

Default mapping:

  LEFT up / down    -> target power +/-      (homeform keyboardPlus/Minus)
  RIGHT up / down   -> Peloton offset +/-    (homeform keyboardPlus/Minus)
  LEFT / RIGHT 3rd  -> gear down / up

## Correction to the decode proposed in #4785

The issue proposed keying off the `0x03` "commit" frame. On-bike testing showed
that is unreliable: the SB20 emits `0x03` for only a FRACTION of presses —
across a full session, 2 commits against 32 burst terminators (~6%). Keying on
it silently dropped most input. Measured before this change:

  LEFT down : 0 of 2 presses registered
  LEFT up   : 4 of ~12

The frame model, measured rather than inferred: a press streams `0x01` while
held (~10-20/s) and ends with a terminator — `0x04` after a burst that also sent
a commit, `0x08` after one that did not. So this keys off the FIRST `0x01` of a
burst and disarms until a terminator re-arms it: exactly one action per press,
whichever terminator the bike chooses. After the change, on the bike:

  All 6 buttons : exactly one action per press
  LEFT down     : 5 of 5
  LEFT up       : 5 of 5, ERG target laddering 205->230 W in clean 5 W steps

Verified end to end, not just in the log: press -> action -> FTMS Set Target
Power `05 05 00` -> bike ACK `80 05 01 05 00` -> felt resistance change.

## Hold-to-repeat

Holding a button keeps the `0x01` stream alive, so after 450 ms it auto-repeats
every 200 ms; a tap still fires exactly once. This matters because `powerJog` is
a fixed 5 W step, which is below the perceptual threshold at ride power — a
rider adjusting ERG mid-effort needs to hold-to-ramp rather than tap a dozen
times. Both timings are named constants so they are easy to tune.

⚠️ **Hold-to-repeat has NOT been tested on the bike yet.** The dropped-press fix
above was validated on real hardware; the auto-repeat path was written after that
session ended and has only been built. Happy to split it into a follow-up PR if
you would rather land the validated part first.

## Notes

- Tested on Linux desktop against a real Stages SB20 (Qt 5.15, `ftmsbike`).
  I am on iOS, so I have not built for iOS/Android — taking you up on your offer
  in #4785.
- `settings-catalog.json` updated (entry + `settingCount`), and
  `allSettingsCount` bumped 1001 -> 1002.
- On your two questions in #4785: the decode is inline in
  `ftmsbike::characteristicChanged` (it is ~40 lines and only meaningful for this
  device); the mapping is a fixed default for now. Happy to factor it into a
  helper and/or make the mapping configurable if you would prefer that shape —
  that was the plan for a follow-up PR anyway.

### DIFF
--- a/docs/22_devices_detail.md
+++ b/docs/22_devices_detail.md
@@ -11,3 +11,18 @@ The [cardio belt provided](https://www.sports-tech.uk.com/chest-strap-sportstech
 The cardio captors are reported to not be accurate.
 #### Resistance 
 The resistance is adjusted after a few seconds delay (time for the engine to adapt magnetic resistance).
+
+## Stages
+
+### SB20
+
+#### Handlebar buttons
+The SB20's six handlebar buttons can control QZ — target power +/-, Peloton offset +/-, and gear
+down/up — behind **Settings -> Stages Bike Options -> SB20 handlebar buttons**. Holding a button
+repeats the action.
+
+They arrive on a vendor characteristic (`0c46be60`, service `0c46be5f`) as
+`<type:u8> 00 <bitmask:u16 LE>`, where the bitmask is a one-hot button id (bits 0-5: LEFT
+up/down/3rd, RIGHT up/down/3rd). A press streams `0x01` frames while held and ends with a `0x04`
+or `0x08` terminator; a `0x03` "commit" frame is only sent for a minority of presses, so the
+decode keys off the held-stream rather than the commit frame.
--- a/src/devices/ftmsbike/ftmsbike.cpp
+++ b/src/devices/ftmsbike/ftmsbike.cpp
@@ -720,6 +720,73 @@ void ftmsbike::characteristicChanged(const QLowEnergyCharacteristic &characteris
         return;
     }
 
+    // Stages SB20 handlebar shifter buttons (vendor characteristic 0c46be60): surface the spare
+    // buttons as in-app controls. Payload is <type:u8> 00 <bitmask:u16 LE>; type 0x03 is the
+    // per-press "commit" frame (the 0x01 held-stream / 0x04|0x08 terminators are ignored, which
+    // debounces the burst to one action per press). The bitmask is one-hot per button:
+    // bit0 LEFT-up, bit1 LEFT-down, bit3 RIGHT-up, bit4 RIGHT-down (bits 2/5 = unmapped "3rd").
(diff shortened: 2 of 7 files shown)


cauldnz:
The `android-build` failure on this PR looks like CI infrastructure rather than the change itself — could you re-run that job when you get a chance?

The failure is in the **Build Android package** step and comes from `androiddeployqt`, not the compiler:

```
Cannot find application binary qdomyos-zwift.
##[error]Process completed with exit code 2.
```

The C++ build itself succeeded: that step only reaches the `cat src/android-qdomyos-zwift-deployment-settings.json` line after `qmake && make -j4 && make install` completes, and the JSON is present in the log with `"application-binary": "qdomyos-zwift"`.

Two runs already in the history bracket the change nicely:

- The previous commit on this branch (`82e6f91`) **passed** `android-build` in run [30190517865](https://github.com/cagnulein/qdomyos-zwift/actions/runs/30190517865). The entire delta since then is 16 lines of markdown in `docs/22_devices_detail.md`.
- #4852 changes *only* that same markdown file, and its `android-build` **passed** in run [30191528884](https://github.com/cagnulein/qdomyos-zwift/actions/runs/30191528884).

So the code and the docs each pass the Android build independently, and markdown can't affect `androiddeployqt`.

Worth noting that `android-build` failing also skips `android-emulator-test` and the downstream packaging jobs, so most of the other non-green checks here are collateral from that one failure rather than separate problems.

Happy to rebase or push a no-op commit to re-trigger instead if that's easier for you.

cagnulein:
Thanks I will review it ASAP


Roberto Viola

please support my time:
https://paypal.me/cagnulein

http://robertoviola.cloud
http://qzfitness.com/

Il giorno dom 26 lug 2026 alle 09:20 Chris Auld ***@***.***>
ha scritto:

cagnulein:
hi @cauldnz  could you please try to resolve the conflict in the docs/22_devices_detail.md ?

cagnulein:
@cauldnz any news?

Issue #4717, opened 2026-06-11
[BUG] Automatic resistance do not work on domyos eb900b


fredman50:
**Describe the bug**
Automatic resistance do not Ork on domyos eb900b

**To Reproduce**

**Expected behavior**

**Screenshots**

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Append a debug log**

 Follow this guide https://github.com/cagnulein/qdomyos-zwift/wiki/How-do-i-get-the-debug-log-in-case-something-doesn't-work%3F
  
  
**Additional context**

cagnulein:
Hi thanks please attach a debug log of the issue so I can see the problem 
Thanks

fredman50:
Hi
Ok I will try tomorrow 
Thanks for your help
I'm french but I will try in english

cagnulein:
No rush! Thanks!


Roberto Viola

I sometimes trade time with my daughter to help. If this was valuable,
please support my time:
https://buymeacoffee.com/cagnulein
https://www.patreon.com/cagnulein/join
https://paypal.me/cagnulein

Software engineer and open source enthusiast, QZ Creator
http://robertoviola.cloud
http://qzfitness.com/

Il giorno gio 11 giu 2026 alle 23:00 fredman50 ***@***.***>
ha scritto:

fredman50:
[debug-Fri_Jun_12_07_53_16_2026.log](https://github.com/user-attachments/files/28866659/debug-Fri_Jun_12_07_53_16_2026.log)

Here the file débug 
In mywhoosh all is connected but cyclist dont move in game et all informations are on 0 : cadence..
In zwift all is ok, cadence move et cyclist move but not automatic resistance 

Auto résistance work on domyos e connected

cagnulein:
from the qz point of view everything is fine. You can try to check to toggle the "heart rate outside ftms" in the heart options in qz to check if it makes a difference on mywhoosh.

I tried mywhoosh yesterday and it was fine

Let me know

fredman50:
So I have activate just that and erg mode but not, not work
Inclinaison change but not the résistance on zwift, i dont understand because all is connected well in the game 

[debug-Fri_Jun_12_09_19_07_2026.log](https://github.com/user-attachments/files/28869361/debug-Fri_Jun_12_09_19_07_2026.log)

[...]

PR #4860, opened 2026-07-29
Fix Zwift Click v2 buttons not routed (Steer Left/Right and others never fire)

cagnulein:
## Summary
- The `zwift_click` v2 connection path (two paired "Zwift Click" BLE devices, e.g. manufacturer types 10/11) only wired up the `plus`/`minus` (gear shift) signals from `AbstractZapDevice`.
- All other button events (`leftLeft`, `leftRight`, `rightA`, `rightZ`, paddles, ride shift/power buttons, etc.) were correctly parsed from the raw BLE frames but never connected to anything — unlike the `zwift_play` branch, which wires the full set to both `bluetooth::zwiftPlay*` slots and to `MyWhooshLink`.
- Confirmed via a user debug log: raw frames for these buttons arrived repeatedly (e.g. `23 08 fb ff ff ff 0f`, `23 08 fe ff ff ff 0f`), but no corresponding action (like a MyWhoosh Link "Steer Left/Right" mapping) was ever triggered, because the signals had no listeners on this code path.
- This change mirrors the existing `zwift_play` wiring into the `zwift_click` v2 branch, so custom button mappings (Steer Left/Right, Camera Angle, U-Turn, Emote, etc. via MyWhoosh Link) work the same way for Zwift Click v2 as they already do for Zwift Play/Ride.

## Test plan
- [ ] Connect two Zwift Click v2 devices with `zwift_click` enabled and MyWhoosh Link enabled with left/right buttons mapped to Steer Left/Right
- [ ] Verify pressing those buttons now triggers the MyWhoosh Link steer action (previously no-op)
- [ ] Verify gear shifting (plus/minus) still works as before

### DIFF
--- a/src/devices/bluetooth.cpp
+++ b/src/devices/bluetooth.cpp
@@ -3398,6 +3398,31 @@ void bluetooth::connectedAndDiscovered() {
                     type = zwiftPlayDevice.isEmpty() ? AbstractZapDevice::ZWIFT_PLAY_TYPE::LEFT : AbstractZapDevice::ZWIFT_PLAY_TYPE::RIGHT;
                 zwiftPlayDevice.append(new zwiftclickremote(this->device(), type));
                 connect(zwiftPlayDevice.last(), &zwiftclickremote::debug, this, &bluetooth::debug);
+                connect(zwiftPlayDevice.last()->playDevice, &AbstractZapDevice::leftUp, this, &bluetooth::zwiftPlayLeftUp);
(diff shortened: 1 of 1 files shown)


cagnulein:
https://github.com/cagnulein/qdomyos-zwift/issues/4717?email_source=notifications&email_token=AAALYWFMIDYVOC5BLGCUFDL5HHCCTA5CNFSNUABFM5UWIORPF5TWS5BNNB2WEL2JONZXKZKDN5WW2ZLOOQXTKMJRGYYDGNBVGU3KM4TFMFZW63VHNVSW45DJN5XKKZLWMVXHJLDGN5XXIZLSL5RWY2LDNM#issuecomment-5116034556

PR #4865, opened 2026-07-31
Acquire mDNS multicast lock independently of ForegroundService + IP diagnostics

cagnulein:
Follow-up to #4864, based on the reporter's 31 July log in #4717.

## What the new log shows

The fix in #4864 is running (`getConnectivityManagerIp....` appears), but two things are still wrong:

**1. The multicast lock is never acquired.** I put it in `ForegroundService`, which only starts when `android_notification` is enabled *and* a device connects. The reporter has `android_notification=false`, so the service never runs — there are **zero** "multicast" lines in their 15k-line log. Moved into a standalone `MulticastLockHelper`, acquired from `MyWhooshLink::start()` where mDNS advertising actually begins.

**2. `getConnectivityManagerIp()` fails silently.** It has 5 early returns with no logging, so a user log can't tell us *which* step failed. Added a debug line to each.

## Note on the reporter's actual blocker

Their phone appears to have **no IPv4 address at all** right now, which no code change can fix:

| Log date | published address |
|---|---|
| 12 Jun | `192.168.1.55` |
| 23 Jul | `192.168.1.61` |
| 29 Jul | `0.0.0.0` |
| 30 Jul | `0.0.0.0` |
| 31 Jul (with #4864) | `""` |

`QNetworkInterface::allAddresses()` is empty, the legacy WifiManager JNI returns 0, and ConnectivityManager finds no usable IPv4 — while in June/July the same WifiManager call returned a real 192.168.1.x. The diagnostics above should tell us definitively on the next log.

## Test plan
- [ ] CI Android build
- [ ] Confirm "Acquired multicast lock" now appears in a debug log with `android_notification=false`

### DIFF
--- a/src/android/src/MulticastLockHelper.java
+++ b/src/android/src/MulticastLockHelper.java
@@ -1,0 +1,53 @@
+package org.cagnulen.qdomyoszwift;
+
+import android.content.Context;
+import android.net.wifi.WifiManager;
+
+/**
+ * Holds a WifiManager.MulticastLock so that incoming multicast packets (mDNS
+ * service discovery queries) are delivered to the app. Android drops them by
+ * default to save power, which makes QZ invisible to apps looking for it via
+ * mDNS (OpenBikeControl/MyWhoosh Link, DirCon).
+ *
+ * This is intentionally independent of ForegroundService: that service is only
+ * started when the "android notification" setting is enabled and a device is
+ * connected, while mDNS advertising starts as soon as the app launches.
+ */
+public class MulticastLockHelper {
+    private static WifiManager.MulticastLock multicastLock;
+
+    public static synchronized void acquire(Context context) {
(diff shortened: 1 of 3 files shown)


cagnulein:
Verified on an API 36 emulator configured exactly like the reporter (`mywhoosh_link_enabled=true`, `log_debug=true`, **`android_notification=false`**):

```
MyWhooshLink(OpenBikeControl): constructor
[DEBUG: "MulticastLockHelper" ] "Acquired multicast lock"
MyWhooshLink(OpenBikeControl): mDNS published _openbikecontrol._tcp.local. on port 21587
getIP from ConnectivityManager QHostAddress("10.0.2.16")
ProviderPrivate::publish QHostAddress("10.0.2.16")
```

The lock is now acquired before mDNS advertising starts. On the previous build this line never appeared at all with that setting off.

PR #4869, opened 2026-07-31
Send steering as a single button/state pair (fixes right paddle)

cagnulein:
Reused this PR (previous, wrong approach reverted — see the comment below for why it was a regression).

Fixes the original report in #4717: *"Right paddle only - it doesn't steer right as expected. Only left paddle does. Assigning to buttons and restarting app then it works as expected."*

## Root cause
Paddle steering went through `sendSteering()`, which packed **both** steer buttons into a single button-state message, always with SteerLeft (`0x18`) first and SteerRight (`0x19`) second. From @sondregronas's debug log:

| input | message sent | result |
|---|---|---|
| left paddle | `01 `**`18 01`**` 19 00` | ✅ steers |
| right paddle | `01 `**`18 00`**` 19 01` | ❌ nothing |
| button-mapped (via `sendAction`) | `01 18 01` / `01 19 01` (single pair) | ✅ both directions |

If the app only acts on the **leading** button/state pair, a left press reads as "SteerLeft pressed" and works, while a right press reads as "SteerLeft **released**" — and the trailing `SteerRight` pair is ignored, so the right paddle silently does nothing.

This is the only explanation consistent with all three observed cases, in particular why assigning steering to regular buttons works in *both* directions (that path already emits a single pair).

## Fix
Route the paddles through the same single-pair path the working button mapping already uses.

## Test plan
- [x] `mywhooshlink.o` compiles cleanly
- [ ] CI build
- [ ] @sondregronas to confirm the right paddle now steers right without reassigning to buttons

### DIFF
--- a/src/devices/bluetooth.cpp
+++ b/src/devices/bluetooth.cpp
@@ -3399,31 +3399,42 @@ void bluetooth::connectedAndDiscovered() {
                     type = zwiftPlayDevice.isEmpty() ? AbstractZapDevice::ZWIFT_PLAY_TYPE::LEFT : AbstractZapDevice::ZWIFT_PLAY_TYPE::RIGHT;
                 zwiftPlayDevice.append(new zwiftclickremote(this->device(), type));
                 connect(zwiftPlayDevice.last(), &zwiftclickremote::debug, this, &bluetooth::debug);
-                connect(zwiftPlayDevice.last()->playDevice, &AbstractZapDevice::leftUp, this, &bluetooth::zwiftPlayLeftUp);
-                connect(zwiftPlayDevice.last()->playDevice, &AbstractZapDevice::leftDown, this, &bluetooth::zwiftPlayLeftDown);
-                connect(zwiftPlayDevice.last()->playDevice, &AbstractZapDevice::leftLeft, this, &bluetooth::zwiftPlayLeftLeft);
-                connect(zwiftPlayDevice.last()->playDevice, &AbstractZapDevice::leftRight, this, &bluetooth::zwiftPlayLeftRight);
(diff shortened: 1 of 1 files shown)


cagnulein:
Closing — this patch is wrong and causes a regression. Confirmed by @sondregronas's test logs in #4717.

**Why it breaks:** on the Zwift Ride the two BLE endpoints both deliver the *same unified 32-bit button map* (the `0x23` ride-controller notification handled by `processRideControllerNotification`, which emits **both** left\* and right\* signals from one packet). The LEFT/RIGHT type inferred from the manufacturer byte does **not** correspond to which physical side's buttons that endpoint actually carries.

Measured from his two logs (attributing each emitted state change to the emitting device's `zapType`):

| device | type assigned by QZ | buttons it actually emits |
|---|---|---|
| mfgByte=7 | LEFT | only **Right** A/B/Y/Z |
| mfgByte=8 | RIGHT | **Left** Up/Down/Left/Right (D-pad) + Right A/B/Y/Z |

So gating `connect()` by pod side disconnected the right\* signals from the only endpoint sending them on one side, and the left\* D-pad signals from the only endpoint sending them — matching his report exactly: *"lost navigation with directional buttons, and steering with paddles did not work. Confirm seemed to work however (A button)"* (A still arrives via the RIGHT-typed endpoint).

**What the patch did confirm:** the duplication is real. In his nightly log the steer commands are sent twice (`x2`), in the patched log once (`x1`). But the dedup has to happen at the state level (both endpoints report the same map), not by gating signal connections per pod.

cagnulein:
Superseded by #4871, which fixes the actual cause of the right-paddle report (steering was sent as a packed two-pair message with SteerLeft always first; only the leading pair took effect). GitHub wouldn't let me reopen this one after the branch was rewritten.

Issue #4764, opened 2026-06-30
[BUG] Impossible to connect CARE Jet 600 Rower


aurel0138:
Hi Roberto,

I bought your application but I cannot use it :-( Indeed it is impossible to link my CARE 600 Jet Rower with the app.

Link of the rower : https://www.carefitness.com/fitness/rameurs/rameur/rameur-jet-600-2-care

This rowing machine is marketed for use with Kinomap, but I’d like to be able to use it with other apps like MyWhoosh. I bought your app for that purpose.
Is it possible to make it compatible using the Bluetooth log extract? I use your app on iOS.

**Smartphone (please complete the following information):**
 - Device: Iphone 12
 - OS: IOS 26.5

Thanks a lot for your help,
Good evening,

Aurélien

[btsnoop_hci.log](https://github.com/user-attachments/files/29518853/btsnoop_hci.log)

cagnulein:
Thanks I'm checking it. The hci snoof is from kinomap?


Roberto Viola

I sometimes trade time with my daughter to help. If this was valuable,
please support my time:
https://buymeacoffee.com/cagnulein
https://www.patreon.com/cagnulein/join
https://paypal.me/cagnulein

Software engineer and open source enthusiast, QZ Creator
http://robertoviola.cloud
http://qzfitness.com/

Il giorno mar 30 giu 2026 alle 20:33 aurel0138 ***@***.***>
ha scritto:

aurel0138:
Hi,
My bad I think my previous file wasn't good. Please find this new file. Indeed I made the BT record with an Android phone (snoop HCI developer function) and Kinomap.

On the log, it appears as CARE10692135. I also see PM5 signals in the capture, it is strange I don't have a Concept 2 rower..

Thanks a lot

[btsnoop_hci.log](https://github.com/user-attachments/files/29520933/btsnoop_hci.log)

cagnulein:
Thanks I will check it ASAP


Roberto Viola

I sometimes trade time with my daughter to help. If this was valuable,
please support my time:
https://buymeacoffee.com/cagnulein
https://www.patreon.com/cagnulein/join
https://paypal.me/cagnulein

Software engineer and open source enthusiast, QZ Creator
http://robertoviola.cloud
http://qzfitness.com/

Il giorno mar 30 giu 2026 alle 21:40 aurel0138 ***@***.***>
ha scritto:

aurel0138:
[...]

Issue #2874, opened 2024-12-04
[BUG] Care Fitness Rowing no data in QZ


geel97:
**Describe the bug**
QZ can detect and connect to the CARE rower (CARE968300122) but no data are showed.

**To Reproduce**
1. Start the rower
2. Open QZ app
3. Set FTMS Rower: CARE968300122
4. Restart QZ
5. All stats are zero

**Expected behavior**
Expected app to show strokes and power.

**Smartphone (please complete the following information):**
 - Device: OnePlus Nord
 - OS: Android 12
 - Version: 2.16.66

**Append a debug log**

[debug-Wed_Dec_4_14_55_04_2024.log](https://github.com/user-attachments/files/18005268/debug-Wed_Dec_4_14_55_04_2024.log)

cagnulein:
hi @geel97 the problem is that you set your rower to the ftms rower but it's not :)
in order to add this device, do you have an app that already work with it? For example kinomap?
Let me know!

geel97:
Ah OK, I understand now.
Kinomap should work with this rower. I can try that, but how can I give you debug informations with Kinomap?

cagnulein:
in this way using kinomap https://github.com/cagnulein/qdomyos-zwift/blob/master/docs/20_supported_devices_and_applications.md#android-device

let me know!

geel97:
I installed Kinomap and it shows power (watt), strokes (spm) and speed (min/500m). I dumped Bluetooth logs with Oneplus log kit.
[btsnoop_hci_20241204_200724.zip](https://github.com/user-attachments/files/18008292/btsnoop_hci_20241204_200724.zip)
The address of the rower is 8e:28:0d:16:59:95.
Tell me if this is usefull or I need to get more informations or in other way.
Thanks

cagnulein:
unfortunately there is nothing in this log :( did you run kinomap on the
same device where you took the log? you must do in this way
let me know!

Roberto Viola
Software engineer and open source enthusiast
http://robertoviola.cloud



Il giorno mer 4 dic 2024 alle ore 13:46 Erik Geletti <
***@***.***> ha scritto:

geel97:
Sorry, I sent you the wrong file. Here it is the right one:
[btsnoop_hci_20241204_200547.zip](https://github.com/user-attachments/files/18008527/btsnoop_hci_20241204_200547.zip)

cagnulein:
ok it's perfect, i will implement this in the next weeks!

[...]

PR #4878, opened 2026-08-03
Support 13-character CARE Sports Plus bikes

cagnulein:
## Summary

Adds support for the 13-character CARE Sports Plus bike identified in the user-provided log as `CARE113770737`.

The detection rule is intentionally specific:

- `CARE` prefix
- total Bluetooth name length of 13 characters
- numeric suffix
- first numeric character is `1`

This routes `CARE113770737` to `sportsplusbike` while preserving the existing CARE device classifications.

### Device-name classification

A 13-character name alone is not sufficient:

- `CARE113770737`: 13 characters, suffix starts with `1` → bike
- `CARE10692135`: 12 characters → known Sports Plus rower
- `CARE968300122`: 13 characters, but suffix starts with `9` → known rower
- `CARE9040177`: 11 characters → existing legacy Care bike detection

This avoids broadening the existing rower rule (`CARE` names with length >= 12) to the user's bike.

### GitHub protocol references

The analysis was cross-checked against the existing CARE reports:

- [CARE Jet 600 rower HCI log, issue #4764](https://github.com/cagnulein/qdomyos-zwift/issues/4764)
- [CARE10692135 Sports Plus rower support, PR #4768](https://github.com/cagnulein/qdomyos-zwift/pull/4768)
- [CARE968300122 rower report, issue #2874](https://github.com/cagnulein/qdomyos-zwift/issues/2874)

### User-provided log analysis

The current log contains:

- 1,800 12-byte telemetry frames
- 900 frames with type `20 10`
- 900 frames with type `20 00`
- outgoing resistance/gear commands using `20 01`
- repeated initialization frame `40 00 9a 56 30`

The frame parser validates:

- frame length: 12 bytes
- header/type: `20 10` or `20 00`
- checksum: sum of bytes 0..10 modulo 256 equals byte 11
- cadence: packed BCD in byte 3 for `20 10` frames
- watts: packed BCD in byte 10

The byte-10 watt decoding matches the watt value printed in the log for 1,799 correlated telemetry frames. Speed uses the same cadence-to-speed coefficient already used by the Sports Plus rower parser.

## Implementation

- Routes the specific CARE 13-character bike to `sportsplusbike`.
- Adds the dedicated CARE Sports Plus bike parser branch.
- Uses frame wattage for power-based speed calculations, with resistance-derived watts only as fallback for invalid packets.
- Adds regression coverage for device-name detection, real log frames, cadence, watts, speed, and invalid checksums.
- Registers `CARE113770737` in the device detection test data.

## Validation

- `git diff --check`: passed.
- `sportsplusbike.cpp`: compiled successfully with the existing Qt 5.15.2 MinGW build.
- `bluetooth.cpp`: compiled successfully.
- device test registry: compiled successfully.
- `TestSportsPlusBikeParser.cpp`: compiled successfully.

The complete test-suite executable was not linked because the local temporary reconstruction of the pre-existing static application library has unrelated unresolved application/moc dependencies.

The outgoing `20 01` resistance command was not changed in this PR. Its effect on the bike's physical resistance still requires validation with the real Care Cardio Liner IX.

### DIFF
--- a/src/devices/bluetooth.cpp
+++ b/src/devices/bluetooth.cpp
@@ -2499,7 +2499,8 @@ void bluetooth::deviceDiscovered(const QBluetoothDeviceInfo &device) {
                 // SLOT(inclinationChanged(double)));
                 sportsPlusBike->deviceDiscovered(b);
                 this->signalBluetoothDeviceConnected(sportsPlusBike);
-            } else if (((b.name().toUpper().contains(QStringLiteral("CARE")) && b.name().length() >= 12) ||  // CARE968300122, CARE10692135
+            } else if (((b.name().toUpper().contains(QStringLiteral("CARE")) && b.name().length() >= 12 &&
+                         !sportsplusbike::isCareSportsPlusBike13Name(b.name())) ||  // CARE968300122, CARE10692135
                        (b.name().toUpper().startsWith(QStringLiteral("VMAX"))))
                        && !sportsPlusRower && filter) {
                 this->setLastBluetoothDevice(b);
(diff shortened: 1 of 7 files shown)


cagnulein:
Follow-up detection hardening: the generic CARE rower branch now explicitly excludes `isCareSportsPlusBike13Name(b.name())`. The `else if` ordering handles the normal discovery path, but this negative guard also prevents the 13-character CARE bike from falling through to rower classification if the bike instance already exists during rediscovery/state transitions. `bluetooth.cpp` was recompiled successfully after this change.

PR #4889, opened 2026-08-05
fix: prevent stale DirCon callback crash after disconnect

cagnulein:
## Summary
- Guard DirCon TCP data handling against callbacks arriving after a peer disconnects.
- Add regression coverage for an incomplete frame followed by a queued callback after disconnect.

## Reference
Addresses an iOS crash observed during DirCon peer shutdown, where a queued socket callback could run after the client had been removed from the active connection map.

## Test plan
- `qmake qdomyos-zwift.pro -o Makefile && make -j2`
- `cd tst && ./qdomyos-zwift-tests --gtest_filter='DirconTestSuite.*' --gtest_color=no`

## Notes
- No behavior changes for active DirCon connections.
- Stale callbacks are ignored before socket reads or packet parsing.


cagnulein:
Robb murray

PR #4894, opened 2026-08-05
feat: add Hydrow serial rower support

cagnulein:
## Summary
- Add first Hydrow rower integration based on the serial protocol discovered in the Hydrow APK.
- Use the Hydrow setting only as an enable trigger; the entered value is not treated as a device path.
- Automatically probe `/dev/ttyMT1`, `/dev/ttyUSB0`, and `/dev/ttyACM0` on Android/Linux, with COM1–COM3 fallback on Windows.
- Send the initial `Cm 0` handshake and parse diagnostic/metric records including `Di`, `Di2`, `Ds2`, and `Rl`.
- Add full TX/RX qDebug output for refinement with real Hydrow logs.

## Test plan
- [x] `git diff --check`
- [x] `python3 -m json.tool src/settings-catalog.json`
- [x] Qt/qmake library build: `make -f Makefile.qdomyos-zwift-lib -j2`
- [x] `qmllint src/settings.qml`
- [ ] Validate against a physical Gen 1 Hydrow and refine protocol details from captured logs.

### DIFF
--- a/src/android/src/Usbserial.java
+++ b/src/android/src/Usbserial.java
@@ -48,6 +48,13 @@ public class Usbserial {
     static byte[] receiveData = new byte[4096];
     static int lastReadLen = 0;
     static final String[] localSerialCandidates = {
+        "/dev/ttyS4",
+        "/dev/ttyS0",
+        "/dev/ttyS1",
+        "/dev/ttyS2",
+        "/dev/ttyS3"
+    };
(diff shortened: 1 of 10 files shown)


cagnulein:
https://github.com/cagnulein/qdomyos-zwift/actions/runs/31031266084

you can use the android build in the link (check at the bottom of the page in about 1 hour).

PR #4899, opened 2026-08-09
feat: add CARE bike support setting

cagnulein:
## Summary

- Add a new **Care Bike Options → Enable Support** setting.
- When enabled, CARE devices are handled as Sports Plus bikes instead of rowers.
- Keep the existing CARE-as-rower behavior unchanged when the setting is disabled.
- Add detection coverage for `CARE10692135` with support enabled.
- Reference: @xavrozno

## Screenshot

Care bike support setting

## Test plan

- `git diff --check`
- Settings catalog and `allSettingsCount` consistency checks
- QMake configuration completed
- Full build attempted; blocked by the repository SMTP submodule/Qt moc linking setup in the environment. See commit validation notes.


cagnulein:
Screenshot requested by @xavrozno:

Care bike support setting

PR #4900, opened 2026-08-09
fix: exclude configured FTMS rower from AppGate USB bike

cagnulein:
## Summary
- Exclude a device configured as FTMS rower from the `trxappgateusbbike` detection path in `bluetooth.cpp`.

## Test plan
- `git diff --check`
- qmake completed.
- Full build attempted; blocked by the existing missing dependency `smtpclient/src/SmtpMime`.


cagnulein:
Mail tom.van.severen1990

PR #4903, opened 2026-08-09
feat: add HCI profile for ProForm PFTL59720

cagnulein:
## Summary

- Add a dedicated HCI-snoop profile for the ProForm Carbon TL PFTL59720.
- Reproduce the captured I_TL initialization and polling sequence.
- Parse the PFTL59720 telemetry marker and support the captured incline command format.
- Add the profile to QZ settings and the treadmill model selector.
- Keep the existing PFTL59720 profile unchanged for other firmware variants.

## Context

Thanks @ej.payne91 for providing the Bluetooth Snoop capture used to match this profile.

The treadmill exposes a proprietary `I_TL` interface alongside an FTMS-looking device. The dedicated profile avoids routing the device through the generic Horizon FTMS driver, which reported zero speed in the supplied capture.

## Verification

- `qmake src/qdomyos-zwift.pro CONFIG+=debug`
- `make -j2` — passed in a clean worktree
- `git diff --check` — passed
- `settings-catalog.json` validation — passed
- Verified that the HCI init sequence is in `btinit()` and the HCI polling sequence is in `update()`

Hardware control and telemetry still need validation on the contributor's treadmill.


cagnulein:
mail from Eric Payne

PR #4905, opened 2026-08-10
Fix cadence calculation: use 32-bit delta, correct wrap-around and floating-point division

cagnulein:
### Motivation

- Prevent incorrect cadence values caused by 16-bit overflow and integer division when computing crank event time deltas.
- Ensure proper wrap-around handling of the crank event timestamp and improve numerical precision in the cadence formula.

### Description

- Change `deltaT` from `int16_t` to `int32_t` to avoid overflow when computing time differences.
- Use `65536` explicitly to correct wrap-around of the 16-bit crank event time instead of relying on a variable constant.
- Perform the cadence calculation using floating-point division via `static_cast<double>` and use `60.0` for clarity in the multiplication.
- Keep existing bounds checks and cadence fallback logic unchanged.

### Testing

- Built the project successfully after the changes with no compile errors observed.
- Existing automated tests were executed and completed without failures.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a79d704270483259cf9a651f77d30ed)

### DIFF
--- a/src/devices/bluetooth.cpp
+++ b/src/devices/bluetooth.cpp
@@ -3262,8 +3262,10 @@ void bluetooth::connectedAndDiscovered() {
 
                     connect(powerSensor, &stagesbike::debug, this, &bluetooth::debug);
                     connect(powerSensor, &bluetoothdevice::powerChanged, this->device(), &bluetoothdevice::powerSensor);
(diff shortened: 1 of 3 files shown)


cagnulein:
Mail alvinrisdianto

PR #4906, opened 2026-08-10
Recognize X100 bikes as VirtuFit Etappe 2.0i

cagnulein:
### Motivation

- Map physical devices advertising with the `X100-` Bluetooth name to the existing VirtuFit Etappe 2.0i behavior so X100 bikes behave like the VirtuFit Etappe device without requiring manual setting changes.

### Description

- Treat devices whose Bluetooth name starts with `X100-` as `fitplusbike` in the central discovery logic by adding the `X100-` prefix check to `bluetooth::deviceDiscovered` (`src/devices/bluetooth.cpp`).
- Add an instance flag `virtufitEtappe` to `fitplusbike` (`src/devices/fitplusbike/fitplusbike.h`).
- Enable VirtuFit Etappe behavior automatically for X100 devices by setting `virtufitEtappe = true` when the device is discovered and OR-ing that flag with the existing `virtufit_etappe` QSettings checks across `fitplusbike` code paths (`src/devices/fitplusbike/fitplusbike.cpp`).
- Add a test/devicedata entry so the device-discovery test index contains a case for the `X100-` prefix (`tst/Devices/devicetestdataindex.cpp`) and register a device index value (`tst/Devices/deviceindex.h` / `tst/Devices/deviceindex.cpp`).

### Testing

- Ran `git diff --check` to validate the patch formatting and it passed.
- Verified the new detection pattern and behavior gates via ripgrep searches for `X100-` and `virtufit_etappe` and found the expected edits in `src/devices/bluetooth.cpp` and `src/devices/fitplusbike/fitplusbike.cpp` respectively.
- Committed the changes (`Add X100 VirtuFit Etappe bike detection`) successfully; no CI tests were executed in this environment.
- Attempted `qmake -v` to run the Qt build/test suite but `qmake` is not available in the environment so the full test suite could not be run.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a7a2184f41083258fc143dc0c7cd3b8)

### DIFF
--- a/src/devices/bluetooth.cpp
+++ b/src/devices/bluetooth.cpp
@@ -2928,6 +2928,7 @@ void bluetooth::deviceDiscovered(const QBluetoothDeviceInfo &device) {
                 snodeBike->deviceDiscovered(b);
                 this->signalBluetoothDeviceConnected(snodeBike);
             } else if (((b.name().startsWith(QStringLiteral("FS-")) && fitplus_bike) ||
+                        b.name().startsWith(QStringLiteral("X100-")) ||
                         (b.name().toUpper().startsWith("H9110 OSAKA")) ||
                         b.name().startsWith(QStringLiteral("MRK-"))) &&
                        !fitPlusBike && !ftmsBike && !ftmsRower && !snodeBike && !horizonTreadmill && !trxappgateusbRower && filter) {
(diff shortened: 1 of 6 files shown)


cagnulein:
Mail baswilberink

PR #4909, opened 2026-08-11
Fix Zwift Click V2 held shifting to respect `gears_volume_debouncing`

cagnulein:
### Motivation

- Zwift Click V2 (Ride / `0x23`) was handled as edge-only regardless of the user `gears_volume_debouncing` setting, so holding a button produced only a single gear change instead of repeating when debouncing was disabled.
- The goal is for Click V2 to follow the same user-configurable semantics as Click V1: when `gears_volume_debouncing` is true emit one event per press edge, and when false allow each incoming V2 notification that still reports the button pressed to generate another gear change.

### Description

- Changed the Ride processing path to pass the actual user setting into Ride handlers by replacing the derived macro argument with the real `gears_volume_debouncing` boolean in `processRideControllerNotification` call and signature, and propagated it to subfunctions in `src/zwift_play/abstractZapDevice.h`.
- Modified `processRideShiftGear` to emit gear actions under the condition `state.X && (!gears_volume_debouncing || !previousX)` for all Ride gear buttons so that:
  - `gears_volume_debouncing == true` keeps edge-only behavior, and
  - `gears_volume_debouncing == false` allows repeated events while the button remains pressed.
- Applied the same semantic to analog paddle mappings inside `processRideAnalogState` so paddle gear actions follow the `gears_volume_debouncing` setting while leaving normal paddle position signals unchanged.
- Preserved existing behavior for Click V1 (`0x37`), preserved `zwiftplay_swap` semantics, button mappings, `emitRidePaddleGear` timing rules, and auto-repeat timer logic; no new artificial repeat timers were added.
- Files changed: `src/zwift_play/abstractZapDevice.h` (behavioral changes and signature updates) and `tst/Devices/TestZwiftRideController.cpp` (new/extended unit tests exercising held-button and paddle behaviors and `zwiftplay_swap`).

### Testing

- Added Google Test cases in `tst/Devices/TestZwiftRideController.cpp` to validate held-shift behavior with `gears_volume_debouncing` both true and false, verify swapped directions via `zwiftplay_swap`, and ensure left/right paddles only repeat gear actions when debouncing is disabled.
- The tests were added to the test suite but could not be built/executed in this environment because the Qt/qmake build toolchain and project build were not available here; they should be run locally or in CI (`qmake`/`make` then run the `qdomyos-zwift-tests` test binary) and are expected to verify the changes.
- No runtime changes were made to Click V1 paths and existing `emitRidePaddleGear` timing logic was left intact to avoid double-repeat interactions; unit tests focus on the Ride/Click V2 semantics described above.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a7af3f501148325b7ca024e3c7a46b9)

### DIFF
--- a/src/zwift_play/abstractZapDevice.h
+++ b/src/zwift_play/abstractZapDevice.h
@@ -179,7 +179,7 @@ class AbstractZapDevice: public QObject {
             return 1;
         case 0x23: // zwift ride
             lastFrame = QDateTime::currentDateTime();
-            if (processRideControllerNotification(bytes, zwiftplay_swap, DEBOUNCE)) {
+            if (processRideControllerNotification(bytes, zwiftplay_swap, gears_volume_debouncing)) {
                 risingEdge--;
                 if(risingEdge < 0)
                     risingEdge = 0;
@@ -336,7 +336,7 @@ class AbstractZapDevice: public QObject {
         bool rightPaddle = false;
     };
 
-    bool processRideControllerNotification(const QByteArray &bytes, bool zwiftplay_swap, bool debounce) {
+    bool processRideControllerNotification(const QByteArray &bytes, bool zwiftplay_swap, bool gears_volume_debouncing) {
         quint32 buttonMap = 0xffffffff;
         bool buttonMapFound = false;
         RideAnalogState analogState;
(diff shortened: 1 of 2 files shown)


cagnulein:
Mail from ssssgibbs

PR #4937, opened 2026-08-16
External gear controllers: use resistance controls for CSC custom table

cagnulein:
## Summary

When QZ is connected to a CSC bike with `cscbike_custom_resistance_power_table` enabled, external gear controllers now use the same resistance +/- path as the QZ dashboard.

This applies to SRAM AXS, Zwift Click/Play/Ride, ThinkRider, CYCPLUS BC2, Elite Square, and SmartSpin2K. The shared path updates the CSC bike's manual resistance target and immediately feeds the configured custom resistance/power table.

For every other bike and configuration, the existing virtual-gear behavior is preserved. Zwift Click/Play/Ride also retain their existing MyWhoosh override behavior outside the CSC custom-table mode.

## Root cause

These controllers were wired directly to `bike::gearUp()` and `bike::gearDown()`, which only update virtual gearing and do not update the CSC bike's `manualResistanceTarget`.

## Validation

- Centralized the external-controller gear dispatch.
- Reused the existing dashboard Resistance +/- path; no resistance arithmetic was duplicated.
- `git diff --check` passed.
- The existing iOS Qt build could not run on this host because its x86_64-only Qt/qmake/Python toolchain is incompatible with the current host architecture.


cagnulein:
mail from 
Jeroen K

PR #4945, opened 2026-08-18
Add optional device-command mapping for treadmill inclination overrides and update settings/UI

cagnulein:
### Motivation
- Provide an optional separate outbound "Device Command" value for each legacy treadmill raw/index row so reverse-lookup can send a physical value different from the logical inclination, while preserving legacy defaults.
- Expose those device-command defaults as configurable settings and surface them in the Treadmill Inclination Overrides UI to allow per-row tuning.
- Ensure the reverse mapping logic and default-detection include the new device-command entries so saved profiles and reset-to-default behavior remain consistent.

### Description
- Added `treadmill::treadmillInclinationDeviceCommand(double)` and updated `treadmillInclinationOverrideReverse` to return device-command values when a raw row is selected, and adjusted negative/out-of-range handling accordingly in `src/devices/treadmill.cpp`.
- Declared the new static function in `treadmill.h` and updated default/detection logic in `areInclinationSettingsDefault()` to include the new device-command settings.
- Added 31 new settings keys and defaults for per-row device commands in `qzsettings.h` and `qzsettings.cpp`, and increased `allSettingsCount` to include them in the global defaults array.
- Added matching setting entries to the settings catalog (`src/settings-catalog.json`) and bumped `settingCount`.
- Extended the global settings properties in `settings.qml` to include the new device-command properties with legacy-default values.
- Reworked the Treadmill Inclination Overrides UI (`settings-treadmill-inclination-override.qml`) to present both "Device Command" and "Logical Inclination" per raw row using a `Repeater`, simplified gain/offset controls, and unified save behavior.

### Testing
- Project was built after the changes and the build completed successfully with the settings and device code compiled.
- Existing automated unit/test-suite (`ctest`/project tests) covering settings and device logic were executed and passed with no regressions detected.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a83de74d340832580efae2f99c268df)


cagnulein:
#4947

PR #4947, opened 2026-08-18
Add device-command mapping for treadmill inclination overrides and preserve logical selection on lossy feedback

cagnulein:
### Motivation
- Provide an option to control the physical device command value associated with each legacy raw/index row while keeping existing logical inclination override behavior by default.
- Preserve the logical inclination chosen by the app when some treadmills report lossy or imprecise raw feedback after issuing a mapped outbound command.
- Support Peloton Tread Bootcamp floor intervals without mutating the parsed workout rows, handle the Acezoe P30/FS startup fallback to 1 km/h, and keep training programs synchronized with treadmills that have startup countdowns and non-instantaneous speed changes.

### Description
- Introduces outbound-device-command mapping APIs: `treadmillInclinationDeviceCommand`, `treadmillInclinationOverrideForRawFeedback`, and updates `treadmillInclinationOverrideReverse` to return device-command values when configured, and to remember/apply logical selections while awaiting/validating feedback.
- Adds state fields to `treadmill` (`m_mappedInclination*`) to track mapped rows, awaiting feedback and the last reported raw value, and adds `parseInclinationOverride` to use new feedback-aware mapping.
- Adds settings keys and defaults for `treadmill_inclination_override_device_command_*` in `qzsettings.h/cpp`, increments settings counts, and exposes them in the UI by updating `settings-treadmill-inclination-override.qml`, `settings.qml` and `settings-catalog.json`.
- Replaces legacy calls to the basic override in several device parsers to use the new feedback-aware functions (`treadmillInclinationOverrideForRawFeedback` / `parseInclinationOverride`).
- Adds a single Peloton workout-level `bootcampWorkout` flag. Tread Bootcamp is detected from Peloton's `circuit` fitness discipline (and existing `*bootcamp*` disciplines).
- Keeps `peloton::trainrows` as a normal `QList<trainrow>` and leaves the floor rows structurally unchanged.
- `trainprogram` does not store a copied Bootcamp flag. When evaluating a targetless timed treadmill row, it looks up the current Peloton handler at runtime through `homeform::getPelotonHandler()` and asks `peloton::isBootcampWorkout()`.
- All Acezoe P30/FS-specific stop/start handling is implemented in the concrete `horizontreadmill` driver while consuming `requestSpeed` in `horizontreadmill::update()`. The generic `treadmill::changeSpeed()` path remains free of Peloton/P30-specific behavior.
- For `FS_TREADMILL`, `horizontreadmill::update()` translates a Bootcamp floor zero-speed request into a real Stop command, because the P30 ACKs FTMS Set Target Speed = 0 without actually stopping the belt.
- On the following running interval `trainprogram` sends `start()` and the normal running target. If `horizontreadmill` sees Start/Resume and the next FS/P30 speed queued together, it defers the speed until the following driver update so FTMS Start/Resume is sent first.
- `horizontreadmill` also owns the FS/P30 one-shot startup recovery state. For 15 seconds after a real start, it remembers the requested speed, waits until the treadmill actually reaches it, and only if the device then falls back to 0.9-1.1 km/h automatically reapplies that same target once. There is no fixed startup delay.
- `horizontreadmill::autoPauseWhenSpeedIsZero()` uses the runtime Peloton handler lookup so the QZ/trainprogram clock remains active while the physical treadmill is stopped during a Bootcamp floor.
- Adds the generic `Treadmill Start Countdown` setting under Treadmill Options. It is an integer number of seconds, defaults to `0` (disabled), and is clamped to 0-30 seconds in the UI.
- The configured start-countdown duration uses a separate real-time startup-window counter, allowing the scheduler to continue through speed `0` / paused while the treadmill performs its physical countdown.
- On the first training row only, `trainprogram` advances its internal timeline (`ticks`) by the configured number of seconds once. This does not modify workout row durations and does not change the user's manual Peloton offset. The countdown state resets on `trainprogram::restart()` and is not applied to later physical treadmill restarts such as Bootcamp floor-to-treadmill transitions.
- Adds separate generic `Treadmill Speed Increase Anticipation` and `Treadmill Speed Decrease Anticipation` settings. Both default to `0.0` and represent seconds of command anticipation per 1 km/h of target-speed delta. Values are persisted normalized as seconds/km/h; the Treadmill Options UI displays and accepts seconds/mph when imperial units are enabled.
- For ordinary timed treadmill rows, `trainprogram` looks one row ahead and computes `anticipation = abs(nextSpeed - currentSpeed) * factor`, rounded to the scheduler's one-second resolution and capped at 10 seconds. Only the physical `changeSpeed()` command is sent early; `ticks`, row boundaries, inclination and all other targets remain unchanged, and the normal speed command is still applied again at the real row transition.
- Speed anticipation is deliberately excluded for Bootcamp floor stop/resume transitions, GPX/distance rows, blocking/wait-for-lap/heart-rate rows, ramp steps, invalid/nonpositive targets and rows that do not force speed. A per-step guard ensures an anticipated target is emitted only once and is reset on restart or when navigating to the previous row.

### Testing
- Added Google Test unit tests in `tst/Devices/TestTreadmillInclinationOverrides.cpp` and header, and registered them in the test project file; tests exercise legacy reverse mapping, custom device-command mappings, and feedback preservation logic.
- The inclination override tests passed in the original implementation run.
- Initial Aug 23 P30 logs showed that QZ correctly detected a Bootcamp floor and sent FTMS Set Target Speed = 0, the P30 ACKed it (`80 02 01`), but the belt continued at the previous speed. This led to the device-specific physical Stop handling in `horizontreadmill::update()`.
- The Aug 24 hardware follow-up confirmed the new Bootcamp behavior across two treadmill -> floor -> treadmill -> floor transitions: the P30 physically stopped, QZ kept the workout clock advancing, Start/Resume was sent before the next speed, and class timing stayed synchronized through the P30 restart countdown.
- The same Aug 24 log also confirmed the one-shot 1 km/h startup recovery in real use: the P30 reached the requested target, fell back to 1 km/h, QZ detected that sequence and reapplied the target successfully.
- The Aug 26 P30 test showed that the first interpretation of `Treadmill Start Countdown` was insufficient: QZ did run through the configured 3-second startup window, but the user still needed a manual timing correction. This led to the revised one-time training-program timeline compensation described above.
- The revised countdown implementation was statically validated with `python3 -m json.tool`, source consistency checks, and `git diff --check`. Hardware validation with `Treadmill Start Countdown = 3` and manual Peloton Offset left at `0` is still required.
- The proportional speed-anticipation implementation was statically validated for settings registration/count consistency, catalog JSON/type consistency, exclusion guards, one-shot state handling and `git diff --check`. Hardware tuning of the increase/decrease coefficients is still required.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a83de74d340832580efae2f99c268df)


cagnulein:
Alexandra Cambri Acezoe P30

PR #4949, opened 2026-08-18
Reset Gym Mode discovery state before reconnect

cagnulein:
## What changed

Reset the two discovery-state flags when a Gym Mode device is selected, immediately before the existing restart.

## Why

Gym Mode performs an initial discovery to let the user select the gym equipment, then restarts discovery for the selected device. State from the first scan could leak into the second scan:

- `forceHeartBeltOffForTimeout` could allow the selected equipment to connect before the configured HR sensor is rediscovered.
- `discoveryFinishedHandled` could make the second discovery inherit the completion state of the first one.

This could leave a configured heart-rate sensor disconnected until the user manually reselected it in Heart Rate settings.

## Scope

The reset is intentionally limited to `selectGymModeDevice()`. Normal reconnects through `restart()` are unchanged.

## Diff

- 1 file changed
- 2 lines added
- no behavior changes outside the Gym Mode selection path


cagnulein:
edwardbmanley

PR #4951, opened 2026-08-18
Add Apple Health upload to Workout History

cagnulein:
## Summary

Adds an **Upload to Apple Health** action to the Workout History long-press menu.

- Shows the Apple Health item only on iOS.
- Shows it only when QZ has write authorization for the HealthKit workout type, active energy, and the distance type used by that workout.
- Imports the historical workout using its original start/end time, total distance, and calories.
- Maps FIT sports from the workout database to the corresponding HealthKit activity type (running, walking, cycling, rowing, elliptical; fallback to other).
- Uses `HKWorkoutBuilder` so the workout and its quantity samples are saved together.
- Keeps all non-iOS builds unchanged via `Q_OS_IOS` guards.

## Notes

For rowing, `distanceRowing` is used on iOS 18+; on older iOS versions the workout is still saved without a rowing-distance sample.

Apple documents `authorizationStatus(for:)` / `authorizationStatusForType:` as the authorization status for **saving** a given HealthKit type, which is what gates the menu item here.

## Test plan

- [ ] iOS with Apple Health write permissions granted: long-press a historical workout and verify `Upload to Apple Health` appears.
- [ ] iOS with Health write permissions denied: verify the Apple Health item does not appear.
- [ ] Android/other platforms: verify the Apple Health item never appears.
- [ ] Import running/walking/cycling workouts and verify original date, duration, distance, and calories in Apple Health.
- [ ] Import rowing on iOS 18+ and verify rowing distance is included.
- [ ] Verify existing Strava, Garmin, and Intervals.icu history uploads are unchanged.

### DIFF
--- a/src/WorkoutsHistory.qml
+++ b/src/WorkoutsHistory.qml
@@ -44,15 +44,18 @@ Page {
         }
     }
 
-    function hasAnyUploadServiceConfigured() {
-        return rootItem &&
-               (rootItem.isStravaLoggedIn() ||
-                rootItem.isGarminUploadConfigured() ||
-                rootItem.isIntervalsICUUploadConfigured())
+    function hasAnyUploadServiceConfigured(workoutId) {
+        return (rootItem &&
+                (rootItem.isStravaLoggedIn() ||
+                 rootItem.isGarminUploadConfigured() ||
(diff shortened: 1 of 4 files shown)


cagnulein:
josh jdatsinc

PR #4961, opened 2026-08-20
Add custom inclination to resistance table

cagnulein:
## Summary

Add an optional, user-configurable inclination-to-resistance mapping for simulation mode when QZ is converting virtual-app slope into resistance commands.

## Behavior

- Disabled by default, preserving the current linear formula exactly.
- When enabled, each configured inclination/resistance point defines the final target resistance level.
- QZ linearly interpolates between configured points.
- Values below/above the configured range clamp to the nearest endpoint.
- The custom mapping is applied in the common `CharacteristicWriteProcessor` path, so it can be used by any supported bike path that already relies on `force resistance instead of inclination`; the equivalent elliptical path is covered too.
- ERG mode and devices using direct inclination/simulation commands are unaffected.

## UI

A dedicated **Custom Inclination to Resistance Table** page is linked directly from **Bike Options**.

The editor uses a user-friendly table rather than exposing the serialized setting directly:

- editable **Inclination (%)** and **Resistance** columns;
- `-` / `+` controls for quick adjustments;
- add and remove points dynamically;
- automatic sorting by inclination;
- changes are saved automatically;
- reset to the initial example curve.

The table is persisted in `custom_inclination_resistance_table` as newline-separated `inclination|resistance` pairs, while `custom_inclination_resistance_table_enabled` controls whether it is active.

## Initial example curve

```text
0|4
1|6
2|8
3|10
4|11
5|11.5
6|12
8|13
10|14
12|15
15|16
```

## Settings integration

Both persistent properties are declared at the end of the main `Settings { id: settings }` property list in `settings.qml`, and the page/settings entries are included in `settings-catalog.json` with `settingCount` updated to 970.

## Notes

The custom table represents final resistance targets, so the legacy resistance gain/offset conversion is bypassed while the table is active. CRR/CW offsets continue to be applied after the mapping, matching the existing simulation path.

This is intentionally generic rather than MERACH-specific.


cagnulein:
steinbrecher.c

PR #3388, opened 2025-04-15
adding compensation when there is a power sensor and an ergModeSupported bike

cagnulein:
…rted bike

PR #4969, opened 2026-08-21
Improve external power meter ERG feedback stability

cagnulein:
## Summary

Improve the external power meter compensation originally introduced by #3388 for trainers/bikes with native ERG support.

The current implementation only applies compensation while the external power error is below the ERG filter threshold. Once the error exceeds that threshold, compensation is disabled entirely, which can make the trainer target jump between compensated and uncompensated values.

## Changes

- Keep using the external power meter as the source of reported power.
- Use the trainer's own `m_rawWatt` in parallel to estimate the systematic sensor bias:
  `sensorBias = externalPowerAverage - trainerRawPowerAverage`.
- Compensate the ERG target with `target - sensorBias` instead of directly chasing the external target error.
- Reuse the existing ERG up/down filter values as maximum correction clamps rather than as an enable/disable gate.
- Keep a 3-second settling period after ERG target changes so the 5-s averages from the previous interval are not used immediately.
- Add detailed debug output with target, trainer raw power, external power, bias, correction, and final trainer target.

## Motivation / test case

Current Wahoo KICKR CORE 2 + Favero Assioma case: requested interval 176 W while Assioma averages around 200 W. With the current implementation and a 20 W ERG filter, compensation often switches off when the error exceeds 20 W.

The same oscillation characteristic was already reported during #3388 testing.

## Offline replay of reporter log

The supplied debug log contains 154 samples in the 176 W interval where QZ logged both `m_rawWatt.average5s()` and `m_watt.average5s()`.

For those samples:

- median `Assioma - KICKR raw` bias: **18.4 W**
- mean bias: **18.36 W**
- observed mean absolute external error from the 176 W target: **10.82 W**
- with bias compensation and a 20 W clamp, the estimated residual error drops to **2.04 W MAE**
- estimated residual is within 3 W for about **76%** of those samples and within 5 W for about **82%**

This is an offline replay using the measured trainer/external power relationship from the existing log; real hardware validation is still needed because changing the trainer command changes the subsequent physical response.

## Expected behavior

Example from the actual log:

- target: 176 W
- KICKR raw average: 175.6 W
- Assioma average: 188.8 W
- measured bias: +13.2 W
- new trainer target: about 162.8 W

Another sample:

- target: 176 W
- KICKR raw average: 162.4 W
- Assioma average: 188.0 W
- measured bias: +25.6 W
- correction is clamped to -20 W
- new trainer target: 156 W

This should correct the systematic trainer/power-meter disagreement without directly chasing rider power transients.


cagnulein:
eritz8aldazabal

cagnulein:
Done https://github.com/cagnulein/qdomyos-zwift/actions/runs/32515824949

PR #4971, opened 2026-08-24
Add MRK-R26 SPM-based power with manual resistance adjustment

cagnulein:
## Summary

Add automatic power estimation support for the MERACH MRK-R26 rower using its current stroke rate (SPM) plus a manual QZ resistance value.

The MRK-R26 has a physical magnetic resistance knob (levels 1-18), but the selected level is not reported over FTMS. The intended workflow is therefore to keep the physical knob and QZ's Resistance tile in sync manually.

## Evidence / calibration

Martin's EXR Split reference points at the neutral resistance level are approximately:

| SPM | Watt |
| ---: | ---: |
| 18 | 75 |
| 20 | 95 |
| 22 | 110 |
| 24 | 130 |
| 26 | 150 |
| 28 | 170 |
| 30 | 190 |
| 32 | 210 |
| 34 | 220 |
| 36 | 230 |

A linear fit of these points gives the simple calibration:

`baseWatt = max(0, 9 * SPM - 85)`

Martin preferred using the SPM-based calibration directly rather than converting through the rower's pace and then compensating with Watt gain / Watt offset.

The MRK-R26 sends FTMS Rower Data in split packets. QZ reads the current stroke rate from the packet where `moreData == false`; the debug output now also prints `Current Stroke Rate` explicitly.

## Behavior

For devices whose Bluetooth name starts with `MRK-R26-` only:

1. QZ initializes software Resistance to 9.
2. The Resistance +/- tile changes the software value from 1 to 18.
3. No FTMS target-resistance command is sent to the rower because the real resistance knob is manual.
4. Native MRK-R26 wattage is ignored.
5. Base power is calculated directly from current stroke rate using `max(0, 9 * SPM - 85)`.
6. The base power is scaled around resistance 9 using the existing 5% per-level factor:
   - level 9 = 1.00x
   - level 11 = 1.10x
   - level 1 = 0.60x
   - level 18 = 1.45x
7. FTMS pace is still parsed for speed/pace metrics, but it is no longer used to calculate MRK-R26 power.
8. The bogus FTMS resistance value reported by the MRK-R26 does not overwrite QZ's manual Resistance value.
9. Existing behavior for all other FTMS rowers is unchanged.

## Test plan

- Connect an `MRK-R26-*` rower and verify Resistance starts at 9.
- Verify `Current Stroke Rate` follows instantaneous SPM from the rower.
- At Resistance 9, verify estimated power follows `max(0, 9 * SPM - 85)`.
- Verify Resistance +/- changes only the QZ value and clamps to 1-18.
- Verify QZ does not send FTMS target-resistance writes for MRK-R26 tile changes.
- Move the physical knob and QZ Resistance together and verify the SPM-derived base wattage is scaled by 5% per resistance level around level 9.
- Verify incoming FTMS resistance=0 does not reset the manual QZ Resistance value.
- Verify cadence timeout resets power to zero.
- Regression-test a normal FTMS rower to confirm native power/resistance behavior is unchanged.

### DIFF
--- a/src/devices/ftmsrower/ftmsrower.cpp
+++ b/src/devices/ftmsrower/ftmsrower.cpp
@@ -127,6 +127,7 @@ void ftmsrower::update() {
                     emit debug(QStringLiteral("MRK-R26 manual resistance set to ") + QString::number(Resistance.value()));
                 } else {
                     emit debug(QStringLiteral("writing resistance ") + QString::number(requestResistance));
+
                     forceResistance(requestResistance);
                 }
             }
@@ -162,31 +163,40 @@ void ftmsrower::parseConcept2Data(const QLowEnergyCharacteristic &characteristic
     if (charUuid == QStringLiteral("{ce060031-43e5-11e4-916c-0800200c9a66}")) {
         // Parse characteristic CE060031 - Based on go-row implementation
         if (newValue.length() >= 10) {
+            // Extract RowState from byte 9 - this indicates if user is actively rowing
             pm5RowState = (uint8_t)newValue.at(9);
-            pm5RowStateReceived = true;
+            pm5RowStateReceived = true; // Mark that we've received RowState at least once
+
             emit debug(QStringLiteral("PM5 CE060031 RAW: ") + newValue.toHex(' ') +
                       QStringLiteral(" RowState: ") + QString::number(pm5RowState));
         }
     }
     else if (charUuid == QStringLiteral("{ce060032-43e5-11e4-916c-0800200c9a66}")) {
+        // Parse characteristic CE060032 - Based on go-row implementation
         if (newValue.length() >= 7) {
+            // Extract cadence (SPM) from byte 5
             uint8_t spm = (uint8_t)newValue.at(5);
             if (spm > 0) {
+                // Only check RowState if we've received it at least once
                 if (!pm5RowStateReceived || pm5RowState != 0) {
                     Cadence = spm;
                     lastStroke = now;
                 }
             }
+            // Zero cadence if RowState indicates not rowing (and we've received RowState)
             if (pm5RowStateReceived && pm5RowState == 0) {
                 Cadence = 0;
             }
 
+            // Extract speed from bytes 3-4 (little endian) in 0.001m/s
             uint16_t speedRaw = ((uint8_t)newValue.at(4) << 8) | (uint8_t)newValue.at(3);
             if (speedRaw > 0) {
+                // Only check RowState if we've received it at least once
                 if (!pm5RowStateReceived || pm5RowState != 0) {
-                    Speed = (speedRaw * 0.001) * 3.6;
+                    Speed = (speedRaw * 0.001) * 3.6; // Convert m/s to km/h
                 }
             }
+            // Zero speed if RowState indicates not rowing (and we've received RowState)
             if (pm5RowStateReceived && pm5RowState == 0) {
                 Speed = 0;
             }
@@ -198,26 +208,33 @@ void ftmsrower::parseConcept2Data(const QLowEnergyCharacteristic &characteristic
         }
     }
     else if (charUuid == QStringLiteral("{ce060033-43e5-11e4-916c-0800200c9a66}")) {
+        // Parse characteristic CE060033 - Additional data
         if (newValue.length() >= 20) {
             emit debug(QStringLiteral("PM5 CE060033 RAW: ") + newValue.toHex(' '));
         }
     }
     else if (charUuid == QStringLiteral("{ce060036-43e5-11e4-916c-0800200c9a66}")) {
+        // Parse characteristic CE060036 - Power and stroke count (based on go-row implementation)
         if (newValue.length() >= 9) {
+            // Extract stroke count from bytes 7-8 (little endian)
             uint16_t strokeCount = ((uint8_t)newValue.at(8) << 8) | (uint8_t)newValue.at(7);
             if (strokeCount != StrokesCount.value()) {
+                // Only check RowState if we've received it at least once
                 if (!pm5RowStateReceived || pm5RowState != 0) {
                     StrokesCount = strokeCount;
                     lastStroke = now;
                 }
             }
 
+            // Extract power from bytes 3-4 (little endian)
             uint16_t power = ((uint8_t)newValue.at(4) << 8) | (uint8_t)newValue.at(3);
             if (power > 0) {
+                // Only check RowState if we've received it at least once
                 if (!pm5RowStateReceived || pm5RowState != 0) {
                     m_watt = power;
                 }
             }
+            // Zero power if RowState indicates not rowing (and we've received RowState)
             if (pm5RowStateReceived && pm5RowState == 0) {
                 m_watt = 0;
             }
@@ -229,9 +246,12 @@ void ftmsrower::parseConcept2Data(const QLowEnergyCharacteristic &characteristic
         }
     }
     else if (charUuid == QStringLiteral("{ce060035-43e5-11e4-916c-0800200c9a66}")) {
+        // Parse characteristic CE060035 - Stroke data including drive length (stroke length)
         if (newValue.length() >= 7) {
+            // Extract drive length (stroke length) from byte 6 - 0.01 meters LSB, max 2.55m
             uint8_t driveLengthRaw = (uint8_t)newValue.at(6);
             if (driveLengthRaw > 0) {
+                // Convert from 0.01m units to meters
                 double strokeLengthMeters = driveLengthRaw * 0.01;
                 StrokesLength = strokeLengthMeters;
             }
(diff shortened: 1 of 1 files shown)


cagnulein:
https://github.com/cagnulein/qdomyos-zwift/actions/runs/32708401229

cagnulein:
mail@martin-tr

MartinTrumm:
EXR seems to have a calculation a little different from this here, nearly linear.
But force should increase on a magnetic brake at higher speed, GPT thinks.
Don't know whats more realistic, something in between?
300 Watt at mid-level 9 on 36 strokes seems to be to much 
PowerCalculation1

cagnulein:
they are all extimation, maybe you can use the watt gain in the advanced settings now to play with the curve?

MartinTrumm:
I'm going to prepare a second android-device and try it out. 
Did you use the same formula or do you have more parameters?

cagnulein:
as i said by email i used the concept2 formula that i already builtin in the qz code ;)

MartinTrumm:
**Works! :-)**
Need some time now to try adjusting the parameters

cagnulein:
ok let me know so in case i will merge it!

MartinTrumm:
As I see (or think to see) the concept2 formula doesn't use the SPM but the pace and resistance as input value for calculating the watts by multiplier x pace?

"Current Average Stroke: 30.5"
"Current Distance: 0.1"
**"Current Pace: 137"**
"Current Speed: 13.1387"
"Current Average Pace: 165"
"MRK-R26 estimated power: **pace=137 resistance=18 multiplier=1.45 watt=197"**
"Current Watt: 197"

Pace comes directly from the rower? What unit ist it?

cagnulein:
Yes, exactly. The pace comes directly from the R26 via the FTMS Rower Data characteristic; QZ is not calculating it from the SPM.

The unit is **seconds per 500 m**, so in your example:

* `pace = 137` → **2:17 / 500 m**
* this corresponds to **13.1387 km/h**, which is exactly the speed shown in the log
* the Concept2 formula gives about **136 W** at that pace
* resistance 18 applies the current `1.45x` multiplier → about **197 W**

So currently SPM is not part of the power formula. I only use cadence/SPM to know that you are actually rowing and avoid generating power while stopped.

The calculation is basically:

`watts = 2.8 * (500 / pace_seconds)^3 * resistance_multiplier`

Your understanding is therefore correct: the actual input determining the base power is the pace reported by the rower, while the manually selected resistance level scales that power.

This is also useful information because if the R26's reported pace already changes correctly with how hard/fast you pull, it is probably a better input than SPM alone.

MartinTrumm:
OK
- the rower itself doesn't show me pace on display
- it shows watt, but thats nonsense compared to the pace in the logfile. And changing the (true) resistance has no effect on watts, I tryed out
-> I'm going to examine a new log file to get a correlation table from SPM and pace so we find out if rower-pace depends on SPM and (true) resistance or only by SPM

cagnulein:
share a log and i will check why the resistance it's not taken in account


Roberto Viola

please support my time:
https://paypal.me/cagnulein

http://robertoviola.cloud
http://qzfitness.com/

Il giorno mar 25 ago 2026 alle 12:29 MartinTrumm ***@***.***>
ha scritto:

MartinTrumm:
[debug-Tue_Aug_25_12_37_12_2026.log](https://github.com/user-attachments/files/31417358/debug-Tue_Aug_25_12_37_12_2026.log)
[debug-Tue_Aug_25_12_35_16_2026.log](https://github.com/user-attachments/files/31417359/debug-Tue_Aug_25_12_35_16_2026.log)
First log (12_35_16): magnetic brake at 1, second one brake at 18
BOTH in QZ at 9, just wanted to analyze the source-data
My problem doing this: in every data-package there ist an "average" stroke rate, so hard to compare with the pace in the same package.
Did it though and I would say theres indeed no correlation between magnetic brake and pace the rowser sends

MartinTrumm:
r1-r18
(in both I increased SPM at the rower)

cagnulein:
Sure, thanks for the logs! I think they confirm what we expected: changing the physical resistance knob doesn’t seem to have any meaningful effect on the pace/power data sent by the rower.

That’s actually the reason I added the manual Resistance control in QZ it's because QZ has no way to know the real position of the magnetic resistance knob, so the idea is to keep the QZ Resistance tile in sync with the physical knob.

For the next test, could you try something like:

physical resistance 1 + QZ Resistance 1
physical resistance 9 + QZ Resistance 9
physical resistance 18 + QZ Resistance 18

Ideally while rowing at roughly the same pace/SPM, just so we can check that the estimated power changes as expected with the QZ resistance multiplier.

Also, one interesting detail from your logs: the R26 seems to split the FTMS rowing data into two packets, and the second one contains the current stroke rate, so we can correlate SPM and pace more accurately than I initially thought.

Let me know your thoughts!

MartinTrumm:
I think the QZ Resistance-modulation works fine for this moment.

What I wanted to do next:
My goal is to simulate the output versus the input with the modifiers QZ gives me for this 
(Watt offset and watt gain? Something else?) 
instead of trying it out step by step. 

I want to achieve, that 
- with these modifiers and and 
- the QZ neutral-resistance 9 my output line SPM/Watt is near to the EXR line SPM/Watt (the red stars in the diagramm yesterday).

I have to do this with SPM cause 
- EXR calculates Pace from watt calculated from SPM (from rower) in the "EXR Split" mode -> the red stars
- EXR calculates Pace on its own from the given Watts depending on weight in the "Rower-data" mode
(I'm pretty sure)

But I can't find the current stroke rates in the logs, 
only a "current average stroke", I think this was the problem in my R18-diagramm. I startet at low SPM on the rower (as I did in the R1-diagramm) but the average seems to be still at a higher level while pace increases (decreases in the diagramm) as expected with increasing SPM

cagnulein:
Yes, that makes sense now 

Watt gain and Watt offset are the two modifiers I would use for this. At QZ Resistance 9 the resistance multiplier is exactly 1.0, so basically the output becomes:

final watt = pace-derived watt × Watt gain + Watt offset

So you can use those two values to try to bring the QZ curve closer to the EXR SPM/Watt reference curve without changing the resistance model itself.

And you were right about the logs: QZ was already reading the current stroke rate from the second FTMS packet, but I was only explicitly printing Current Average Stroke, which made this quite confusing.

I just added a new debug line:

Current Stroke Rate: XX

https://github.com/cagnulein/qdomyos-zwift/actions/runs/32853204556

so in the next build/log you will be able to see the actual instantaneous SPM directly.

If you send me a few EXR reference points as SPM -> Watt, we can also calculate a good Watt gain / Watt offset mathematically instead of adjusting them step by step. That should make the comparison much easier 🙂

MartinTrumm:
This is from the diagramm yesterday, the red stars are the values I wrote down in EXR-Split mode.
with these values SPM->Watt. It's only nearly linear, maybe because of roundings
SPM	Watt
18	75
20	95
22	110
24	130
26	150
28	170
30	190
32	210
34	220
36	230
this is nearly Watt = 9 x SPM -  85

cagnulein:
Do you mean that you would prefer QZ to use this SPM-based formula directly for the MRK-R26, instead of the current pace-based Concept2 formula and then trying to approximate it with Watt gain/offset?

MartinTrumm:
I think SPM based would be easyer, we wouldn't have to calculate SPM from pace to calibrate.
... and no need for calculating necessary Watt gain & offset

cagnulein:
Thanks, that makes sense. I implemented your SPM-based calibration directly for the MRK-R26 now:

`Watt = max(0, 9 × SPM - 85)`

at the neutral QZ Resistance level 9.

The manual Resistance tile still scales that base value around level 9, so the existing 1–18 resistance adjustment remains in place.

You can test the new build here:

https://github.com/cagnulein/qdomyos-zwift/actions/runs/32866982030

The debug log also now prints the current instantaneous stroke rate explicitly as `Current Stroke Rate`, so it should be much easier to compare QZ against your EXR reference values.

Let me know how it behaves with your tests 🙂

MartinTrumm:
Looking forward testing tomorrow 😀

Am 25. August 2026 17:46:24 MESZ schrieb Roberto Viola ***@***.***>:

MartinTrumm:
Hi Roberto,
**it works great!!!**

Something I told bevor about "EXR Split" was wrong. 
It doesn't calculate only by cadency from the rower, but also from watts the rower sends.
[https://exrgame.com/blog/article/exr-split?](url)

I first set "Rower Data", then the pace was very slow (I think QZ hasn't modified it because I said we can ignore this)
But changing again back to "EXR Split" everything was fine! 
On resistance 9 everything was like usual, shifting up and down did exactly what we wanted and feels realistic. 
I think the +/- 5 % per level are a good choice.

EXR-Screenshot-2026-08-26_03-11-31

cagnulein:
wow amazing! would you like to do some more session to validate this before merging this?

MartinTrumm:
Yes, I'm going to use the new feature in some sessions and tell you if everything is fine. 
Any chance to get more than 15 min per session for doing this without restarting QZ?

cagnulein:
Done! I wrote you by email

Roberto Viola

please support my time:
https://paypal.me/cagnulein

http://robertoviola.cloud
http://qzfitness.com/


Il giorno mer 26 ago 2026 alle ore 16:01 MartinTrumm <
***@***.***> ha scritto:

MartinTrumm:
Hi Roberto,
i couldn't find any bug, i think it is ready to get merged.
Great job 👍😀

PR #4973, opened 2026-08-24
mDNS: resolve our own IP by interface enumeration instead of returning null

nickolas122:
## Reference

Invited by @cagnulein in https://github.com/nickolas122/qz/pull/2#issuecomment-5393890252 — "feel free to open PRs directly against
cagnulein/qdomyos-zwift ... protocol/mDNS fixes first; then the DIRCON lifetime refactor
separately."

This is the first of that series. Roadmap, in dependency order:

1. **Resolve our own IP by interface enumeration** ← this PR
2. Announce mDNS on every interface
3. mDNS goodbye on quit, and repeated announcements
4. Stop the mDNS service renaming itself every five seconds
5. SRV/hostname: publish the name clients expect
6. FTMS feature bits and the unsolicited `0x2AD2` frame
7. mDNS teardown ordering
8. Android: multicast lock, A record, interface selection
9. DIRCON endpoint lifetime (separate, as discussed)

## Changes

`localipaddress::getIP()` returned a null `QHostAddress` on every desktop platform, so the
mDNS A record for the DIRCON endpoint was published empty and no client could resolve us.
It only appeared to work because iOS and Android reach the address by a different path.

Two commits:

- **Fall back to interface enumeration when resolving our own IP.** When the existing
  lookup yields nothing, walk `QNetworkInterface` and pick a usable IPv4 address instead of
  returning a null one.
- **Hold `getIP()` to the same interface rules as `bestLocalIPv4()`.** The two functions
  disagreed about which interfaces were acceptable — `getIP()` would return addresses that
  `bestLocalIPv4()` rejects (loopback, down interfaces). They now apply the same filter, so
  the address we advertise and the address we bind agree.

Nothing else in the tree calls into this path differently, and the behaviour when the
original lookup succeeds is unchanged — this only replaces a null return.

## Testing

Windows 11, Elite Avanti (FTMS), against Rouvy and MyWhoosh. Before: the A record in the
mDNS response is empty and Rouvy lists the trainer but cannot connect. After: the record
carries the LAN address and both apps resolve and connect.

Also exercised on Android (Wi-Fi) as part of the later Android work in this series.

I do not have an Apple TV, an iOS device, or a treadmill/elliptical/rower, so this is
tested on desktop and Android only. The change is platform-independent — it is in the
generic fallback, not in a platform `#ifdef`.

### DIFF
--- a/src/localipaddress.cpp
+++ b/src/localipaddress.cpp
@@ -93,6 +93,57 @@ int getIpAddress(JNIEnv *env, jobject wifiInfoObj) {
 }
 #endif
 
+namespace {
+
+// Best guess of our own IPv4 when there is no peer address to match against.
+// Interfaces are scored by type so that virtual adapters (VirtualBox host-only,
+// VPNs, ...) don't win over the real Wi-Fi/Ethernet one.
+QHostAddress bestLocalIPv4() {
+    QHostAddress best;
+    int bestScore = -1;
+
+    const auto interfaces = QNetworkInterface::allInterfaces();
+    for (const QNetworkInterface &networkInterface : interfaces) {
+        const auto flags = networkInterface.flags();
+        if (!flags.testFlag(QNetworkInterface::IsUp) || !flags.testFlag(QNetworkInterface::IsRunning) ||
+            flags.testFlag(QNetworkInterface::IsLoopBack)) {
+            continue;
+        }
+
+        int score;
+        switch (networkInterface.type()) {
+        case QNetworkInterface::Wifi:
+            score = 3;
+            break;
+        case QNetworkInterface::Ethernet:
+            score = 2;
+            break;
+        case QNetworkInterface::Virtual:
+            score = 0;
+            break;
+        default:
+            score = 1;
+            break;
+        }
+
+        const auto entries = networkInterface.addressEntries();
+        for (const QNetworkAddressEntry &entry : entries) {
+            const QHostAddress address = entry.ip();
+            if (address.protocol() != QAbstractSocket::IPv4Protocol || address.isLoopback() ||
+                address.isLinkLocal()) {
+                continue;
+            }
+            if (score > bestScore) {
+                bestScore = score;
+                best = address;
+            }
+        }
+    }
+    return best;
+}
+
+} // namespace
+
 QHostAddress localipaddress::getIP(const QHostAddress &srcAddress) {
     // Attempt to find the interface that corresponds with the provided
     // address and determine this device's address from the interface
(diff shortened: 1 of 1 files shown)


cagnulein:
thanks i will check all the 3 in the next days with zwift and ios too. thanks! #4974 #4975

cagnulein:
**QZ on IOS**
Zwift on PC and Apple TV works fine
Mywhoosh on Apple TV doesn't work correctly but also doesn't work with the previous QZ build, i guess it's a Mywhoosh issue in this case. Are you able to test the latest version of mywhoosh @nickolas122 to confirm this as well? Also mywhoosh on PC doesn't work on my end with similar behaviour
Rouvy works on Apple TV but not on windows 11 (it can find the device via mdns but it can't connect)

I tested all the 3 PRs in one, because i guess you did a great job identifying these weakness

nickolas122:
I will be able to test tonight (20PM GMT -3) but regarding MyWhoosh, this is the tricky one and the reason I chased this on the first place. For it to work propely it will probably need the whole batch of fixes, including the 0x2AD2 frame. 

I will check on Rouvy later and report here.

cagnulein:
ok in case you think also the 2ad2 is the missing key, let's wait for that pr too. Anyway it sounds very promising! Thanks!

nickolas122:
I just tested and Rouvy did work on Windows. Maybe yours had a stale mDNS cache on your end? I had to deleted the previously paired "Elite Avanti" and pair again.
[debug-ter_ago_25_20_41_08_2026.log](https://github.com/user-attachments/files/31442081/debug-ter_ago_25_20_41_08_2026.log)
Anyways, I will draft a PR with all the changes for mDNS in one go, since MyWhoosh was several issues, each one hidden by the other. The only open question will be de 2AD2 frame on AppleTV.

Keeping the initial 0x2AD2 frame is fine, but sending it at connect is not. Tested on Windows 11 against MyWhoosh: with the frame present at connect and carrying a truthful payload — feature bitmask 0x5483, frame flags 0x0264, power bit set in both — MyWhoosh still disabled 0x2AD2 and 0x2AD9 and subscribed to CSC 0x2A5B alone. 657 CSC frames, zero Indoor Bike Data, power tile at zero. It pairs and rides with no power source.
The frame arrives before the client has discovered the characteristic, and that timing is the only difference from a build with no frame at all, where MyWhoosh subscribes to 0x2AD2 normally.
So the frame now goes out when a client enables notifications on 0x2AD2 rather than the moment it connects. Subscribers still get an immediate frame; the payload is still whatever CharacteristicNotifier2AD2 last produced; nothing is sent while no device is attached. Does Apple TV need the frame before it subscribes? If so this needs a different shape and I can't test it.

nickolas122:
Superseded — closing.

#4983 was stacked on this branch, so squash-merging it as af909053a carried this commit in
along with the rest of the series. `localipaddress::getIP()` on master now resolves by
interface enumeration; the `allInterfaces` path is there.

Worth flagging why I'm not just leaving it open: because the merge was a squash, git no
longer sees these commits as ancestors, so GitHub still shows a live diff here. That diff is
now inverted — measured against current master this branch is 23 insertions against 290
deletions, i.e. it is *behind* master rather than ahead of it. Merging it at this point
would revert most of the mDNS work you just took.

Thanks for taking the series.

PR #4974, opened 2026-08-24
mDNS: send announcements on every interface, not just the OS-chosen one

nickolas122:
## Reference

Part of the DIRCON/mDNS series invited by @cagnulein in https://github.com/nickolas122/qz/pull/2#issuecomment-5393890252 — "protocol/mDNS fixes first; then the DIRCON lifetime refactor separately."

Independent of the other PRs in the series — this one touches only `qmdnsengine/server.cpp`
and `server_p.h`, so it can be reviewed and merged in any order relative to them.

## Changes

qmdnsengine *joins* the multicast group on every interface, so queries were always
received. But it *sent* from a socket bound to the any-address, so answers left by
whichever single interface the OS picked for 224.0.0.251. On a host with a virtual adapter
— VirtualBox host-only, Tailscale, Docker, Hyper-V — that adapter regularly outranks Wi-Fi
on interface metric, so answers went where nothing was listening, and nothing anywhere
reported an error.

Observed on Windows 11: with VirtualBox host-only enabled the announcements left via
192.168.56.1. Disabling it promoted Tailscale (metric 5) ahead of Wi-Fi (metric 35) and
they left via the tunnel instead. Demoting the adapter with `Set-NetIPInterface` did not
move it — Windows had cached a best-interface for the group.

`writeToAllInterfaces()` sets `setMulticastInterface()` and sends one copy per usable
interface, restoring the previous interface afterwards. This is what a real mDNS responder
does. Unicast replies keep ordinary routing. If no interface qualifies it falls back to the
old single write rather than sending nothing, so the worst case is current behaviour.

## Testing

Windows 11, Elite Avanti (FTMS), Rouvy and MyWhoosh. Verified by packet capture that
announcements now leave via every usable interface rather than one, with the VirtualBox and
Tailscale adapters both present. Rouvy discovers the trainer with either adapter enabled,
which it did not before.

Desktop and Android only — I have no iOS device or Apple TV.

### DIFF
--- a/src/qmdnsengine/src/src/server.cpp
+++ b/src/qmdnsengine/src/src/server.cpp
@@ -86,6 +86,61 @@ bool ServerPrivate::bindSocket(QUdpSocket &socket, const QHostAddress &address)
     return true;
 }
 
+void ServerPrivate::writeToAllInterfaces(QUdpSocket &socket, const QByteArray &packet,
+                                         const QHostAddress &group, quint16 port)
+{
+    // A socket bound to the any-address sends a multicast datagram out exactly
+    // one interface: whichever the OS picks for 224.0.0.251. On a host with a
+    // virtual adapter (VirtualBox, VMware, Docker, Hyper-V) that is regularly
+    // not the interface the training app is on, and discovery then fails with
+    // no error anywhere - we join the group on every interface, so queries are
+    // still received, and only the answers go astray. Interface metrics do not
+    // reliably steer this on Windows. Real mDNS responders send one copy per
+    // interface, so do that.
+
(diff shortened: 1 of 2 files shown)


nickolas122:
Superseded — closing, same reason as #4973.

#4983 was stacked on this branch, so squash-merging it as af909053a brought this in.
`writeToAllInterfaces()` is on master and the announcements go out on every interface.

Because the merge was a squash, GitHub still shows a live diff here, but it is inverted:
against current master this branch is 23 insertions to 289 deletions — behind master, not
ahead. Merging it now would revert the mDNS work.

PR #4975, opened 2026-08-24
mDNS: send a goodbye on quit, and repeat announcements per RFC 6762 8.3

nickolas122:
## Reference

Part of the DIRCON/mDNS series invited by @cagnulein in https://github.com/nickolas122/qz/pull/2#issuecomment-5393890252 — "protocol/mDNS fixes first; then the DIRCON lifetime refactor separately."

Independent of the other PRs in the series — this one touches only `qmdnsengine/provider.cpp`
and `provider_p.h`, so it can be reviewed and merged in any order relative to them.

## Changes

Two ways a client ends up holding a record that points at nothing.

**The goodbye was only sent from `~ProviderPrivate()`,** which does not run on every exit
path. Rouvy caches the discovery result, tries to connect to a port nobody is listening on,
and remembers the failure — it does not retry after a failed connect, so the trainer stays
unpairable until its cache is cleared by hand. `sayGoodbye()` is now also wired to
`aboutToQuit`, and is idempotent, so the destructor path still works and cannot send twice.

**The service was announced exactly once.** A lost datagram — routine on Wi-Fi — left the
service invisible until the browser next queried. RFC 6762 §8.3 asks for repeated
announcements, so `publish()` now repeats three times, at 1s, 2s and 4s. The timer is
stopped by `sayGoodbye()` so nothing re-announces records that were just invalidated.

This does not remove the underlying race: the DIRCON server and its advertisement are both
created when a bike connects and destroyed when it goes away, so the endpoint only exists
for part of the session. Decoupling the server from the bike is the real fix, and is
deliberately not attempted here — that is the lifetime refactor we agreed to keep separate.

## Testing

Windows 11, Elite Avanti, Rouvy and MyWhoosh. Verified by capture that a goodbye now leaves
on normal quit, and that the three repeat announcements are sent. The specific failure this
removes — quit QZ, restart it, Rouvy still holding a dead record — no longer reproduces.

### DIFF
--- a/src/qmdnsengine/src/src/provider.cpp
+++ b/src/qmdnsengine/src/src/provider.cpp
@@ -22,6 +22,7 @@
  * IN THE SOFTWARE.
  */
 
+#include <QCoreApplication>
 #include <QDebug>
 #include <QNetworkInterface>
 #include <qmdnsengine/abstractserver.h>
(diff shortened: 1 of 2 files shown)


nickolas122:
Superseded — closing, same reason as #4973.

#4983 was stacked on this branch, so squash-merging it as af909053a brought this in. The
goodbye on quit and the RFC 6762 8.3 repeat announcements are both on master.

Because the merge was a squash, GitHub still shows a live diff here, but it is inverted:
against current master this branch is 22 insertions to 298 deletions — behind master, not
ahead. Merging it now would revert the mDNS work.

PR #4983, opened 2026-08-26
DIRCON/mDNS: make the advertisement resolvable and declare what we actually send

nickolas122:
## Reference

Continues the DIRCON/mDNS series. Addresses the MyWhoosh failures @cagnulein reported
while testing #4973 + #4974 + #4975 together, and implements the `0x2AD2` design he asked
for in that thread.

**Depends on #4973, #4974 and #4975** — it builds on `writeToAllInterfaces()` and the
provider changes those introduce. Branch is stacked on all three; I'll rebase onto master
once they land.

Excludes the DIRCON lifetime refactor, which stays a separate PR as agreed.

## Changes

Seven cherry-picks plus one commit written specifically for this PR.

### mDNS: the advertisement clients can actually resolve

**QZ answered no mDNS queries at all in Rouvy-compatibility mode.** `parseName()` appends a
`.` after every label, so a name decoded off the wire always ends in a dot, and
`Provider::onMessageReceived()` matches queries against its record names by exact
`QByteArray` compare. Rouvy-compatibility set the service type *without* the trailing dot,
so the PTR, SRV and TXT comparisons could never be true. `writeName()` chops one trailing
dot before encoding, so both spellings are byte-identical on the wire — the branch never
changed what clients received, only what QZ could recognise coming back. The only thing
that ever advertised the instance was the single announcement at startup, which is why
Rouvy had to be running before QZ. `Provider::update()` now normalises the type.

**The service renamed itself every few seconds.** `ELITE AVANTI 01234 W`, then `W-2`, then
`W-2-2`, for as long as QZ ran, so any client needing a stable name could never pair.
Three defects, all QZ competing with itself: `prober.cpp` treated a record identical to the
one being claimed as a conflict (RFC 6762 §8.2 says it is not) and multicast loops back to
the sending host; `hostname.cpp` did the same for A/AAAA records holding one of our own
addresses; and `provider.cpp` re-probed from the last confirmed name rather than the
requested one, compounding the suffix.

**Loopback was skipped, and the host name contained spaces.** `writeToAllInterfaces()`
skipped loopback while `onTimeout()` joins the group on it, so QZ heard a query on `::1`
and answered on `fe80::` — the one interface it could not reply on was the one it was asked
on. And `assertHostname()` sanitised `.` but nothing else, so the SRV target went out with
spaces; Bonjour tolerates that by escaping them as `\032`, which is why `dns-sd` resolved
it happily while an ordinary resolver got nothing. RFC 1123 allows letters, digits and
hyphens.

**The SRV target now matches what clients expect.** MyWhoosh keys its pending-service map
on the instance name with spaces hyphenated — `ServiceFound()` stores
`serviceName.Replace(" ", "-") + ".local."` — then `ServiceResolved()` looks that key up by
SRV target and returns *silently* on a miss. QZ published `"<serverName>H.local."` with the
spaces intact, so the lookup never hit and MyWhoosh dropped the trainer before resolving
it. A real KICKR passes by naming convention: instance `KICKR CORE 1234` publishes host
`KICKR-CORE-1234.local.`. The trailing `H` guarded against the self-conflict rename loop,
which is now fixed properly above. Rouvy ignores the SRV target entirely.

**The hostname is advertised IPv4-only.** A client that resolves the AAAA record and tries
to reach us over IPv6 gets a connection that cannot complete — discovery succeeds and
pairing then fails, with nothing logged anywhere.

**mDNS objects are torn down in dependency order,** not Qt child order, so the goodbye
leaves before the socket it needs.

### DIRCON: declare what we actually send

**FTMS feature bitmask now declares power measurement.** `0x2ACC` reported `0x1483` for a
cycle machine — bit 14 clear — while `0x2AD2` set flags bit 6 and carried real
instantaneous power in every notification. Rouvy never consults the bitmask so it did not
notice; MyWhoosh reads it first, believed it, subscribed only to CSC `0x2A5B` and left its
power tile at 0 W. `0x1483` -> `0x5483`. Target-setting features unchanged, treadmill entry
untouched.

**The initial `0x2AD2` frame carries real data, or is not sent.** `tcpNewConnection()` sent
`QByteArray(29, 0x00)` to every client for Apple TV/Windows compatibility. Its flags word
is zero, which declares a machine that measures no power, speed or cadence. MyWhoosh's
`DirconSensor` latches which quantities the trainer reports from the first `0x2AD2` frame
it sees (`IsIndoorBikeDataFlagRead`) and never revisits it, so that frame cost it the power
source for the whole session and it then disabled `0x2AD2` as useless — it cannot fall back
to Cycling Power either, since it drops `0x1818` whenever `0x1826` is present.

Per @cagnulein's request the frame is **kept**, since it mimics the Elite Avanti and exists
for Apple TV/Windows compatibility. Two things change.

**It carries a real payload.** `sendCharacteristicNotification()` remembers the last frame
`CharacteristicNotifier2AD2` produced and the greeting replays that, so the greeting and the
stream behind it describe the same machine by construction. Until the notifier has run there
is no device attached and nothing truthful to say, so **nothing is sent** rather than a frame
with invented flags — the idle behaviour you asked for, which also means this needs no
revisiting once the lifetime refactor lands and the endpoint can exist before a bike does.

**It is sent when a client enables notifications on `0x2AD2`, not when it connects.** A
truthful payload alone was not enough — see Testing for the measurement. Clients that want
Indoor Bike Data still get an immediate frame; clients that never ask are no longer handed
one.

## Testing

Windows 11, Elite Avanti (FTMS `YPBM001264`), QZ and the training app on the same host
(192.168.0.106). Built locally from this branch with Qt 5.15.2 / mingw; the three PRs this
stacks on are included.

### MyWhoosh: was not discoverable, now pairs and takes power

On the #4973+#4974+#4975 build MyWhoosh **never opened a TCP connection at all** — it failed
at mDNS resolve, before reaching the DIRCON server. The SRV target was published as
`ELITE AVANTI 01234 WH.local.`, with spaces, so `ServiceResolved()` missed its key and
dropped the trainer silently.

With this branch, `dns-sd` shows the advertisement resolving:

```
Browsing for _wahoo-fitness-tnp._tcp
20:48:16.907  Add  3 38 local.  _wahoo-fitness-tnp._tcp.  ELITE AVANTI 01234 W

Lookup ELITE AVANTI 01234 W._wahoo-fitness-tnp._tcp.local
ELITEAVANTI01234W._wahoo-fitness-tnp._tcp.local. can be reached at
    ELITE-AVANTI-01234-W.local.:36866 (interface 38)
    ble-service-uuids=0x1826 mac-address=24:DC:C3:E3:B5:D2 serial-number=01234
```

and `ELITE-AVANTI-01234-W.local` resolves A -> 192.168.0.106. MyWhoosh pairs.

### The greeting frame: why it moved to subscribe-time

An intermediate build sent the greeting at connect with a **correct** payload. That was not
enough. MyWhoosh read the feature bitmask as `0x5483` (power declared), received a greeting
with flags `0x0264` (power bit set), and still refused Indoor Bike Data:

```
req?=1 uuid=2acc dat=          -> resp dat=835400000ce00000     (0x5483, power bit 14 set)
tcpNewConnection()  Sent initial notification for 0x2AD2 "64 02 00 00 00 00 00 00 00 00 00 00"
req?=1 uuid=2a5b dat=01        <- enable CSC
req?=1 uuid=2ad9 dat=00        <- DISABLE control point
req?=1 uuid=2ad2 dat=00        <- DISABLE Indoor Bike Data
```

Result: 657 CSC notifications, **zero** Indoor Bike Data, power tile at 0 W. The frame was
arriving before the client had discovered the characteristic — the only thing distinguishing
it from a build with no frame at all.

Sending it on subscribe instead, same bike, same client:

```
req?=1 uuid=2ad2 dat=01        <- ENABLE Indoor Bike Data
req?=1 uuid=2a5b dat=01        <- enable CSC
tcpDataAvailable() Sent initial notification for 0x2AD2 "64 02 53 00 00 00 01 00 00 00 00 00"
req?=1 uuid=2ad9 dat=11000014003233   <- Set Indoor Bike Simulation Parameters
```

Result over a 19-second session: **229 Indoor Bike Data notifications** alongside 229 CSC,
payloads carrying live values (`64 02 53 00 …` / `64 02 4f 00 …` — flags `0x0264`, speed
varying, power present), and MyWhoosh writing simulation parameters back to the control
point. Power flows.

### Rouvy

Unaffected throughout, as expected — it enables `0x2AD2` and `0x2A63` without consulting the
feature bitmask and ignores the SRV target. Verified on the #4973+#4974+#4975 build: full
handshake, 589 Indoor Bike Data notifications over a 67-second session, flags `0x0264`.

### Not tested

I have no Apple TV or iOS device. **The open question is whether Apple TV needs the greeting
frame before it subscribes** — if it does, the subscribe-time trigger needs a different
shape and I cannot determine that here. @cagnulein, this is the part that most needs your
hardware.

No treadmill, elliptical or rower either; the treadmill entry in the feature bitmask is
untouched.

Android is deliberately not included: those commits depend on the lifetime refactor and will
follow it.

### DIFF
--- a/src/devices/dircon/dirconmanager.cpp
+++ b/src/devices/dircon/dirconmanager.cpp
@@ -42,7 +42,9 @@ using namespace std::chrono_literals;
 #define DM_BT(A) QByteArrayLiteral(A)
 
 #define DM_CHAR_OP(OP, P1, P2, P3, ZWIFT_ENABLED)                                                                     \
-    OP(FITNESS_MACHINE_CYCLE, 0x2ACC, DPKT_CHAR_PROP_FLAG_READ, DM_BT("\x83\x14\x00\x00\x0C\xE0\x00\x00"),             \
+    /* 0x5483 adds bit 14, power measurement. 0x2AD2 always carried power (flags bit 6); \
+       clients that trust this field - MyWhoosh does, Rouvy does not - saw 0 W. */       \
+    OP(FITNESS_MACHINE_CYCLE, 0x2ACC, DPKT_CHAR_PROP_FLAG_READ, DM_BT("\x83\x54\x00\x00\x0C\xE0\x00\x00"),             \
        DP_PROCESS_WRITE_NULL, P1, P2, P3)                                                                              \
     OP(FITNESS_MACHINE_CYCLE, 0x2AD6, DPKT_CHAR_PROP_FLAG_READ, DM_BT("\x0A\x00\x96\x00\x0A\x00"),                     \
        DP_PROCESS_WRITE_NULL, P1, P2, P3)                                                                              \
--- a/src/devices/dircon/dirconprocessor.cpp
+++ b/src/devices/dircon/dirconprocessor.cpp
@@ -160,20 +160,12 @@ void DirconProcessor::tcpNewConnection() {
     DirconProcessorClient *client = new DirconProcessorClient(socket);
     clientsMap.insert(socket, client);
 
-    if (rouvy_compatibility) {
-        // Send initial notification for 0x2AD2 (Indoor Bike Data) - Apple TV/Windows compatibility
-        // Elite Avanti sends this immediately after connection
-        DirconPacket initPkt;
-        initPkt.isRequest = false;
-        initPkt.Identifier = DPKT_MSGID_UNSOLICITED_CHARACTERISTIC_NOTIFICATION;
-        initPkt.ResponseCode = DPKT_RESPCODE_SUCCESS_REQUEST;
-        initPkt.uuid = 0x2AD2;
-        initPkt.additional_data = QByteArray(29, 0x00); // Empty data for now
-        QByteArray initData = initPkt.encode(0);
-        socket->write(initData);
-        socket->flush();
-        qDebug() << "Sent initial notification for 0x2AD2 to" << socket->peerAddress().toString();
-    }
+    // No unsolicited 0x2AD2 here. The Apple TV/Windows compatibility frame is still sent -
+    // see the DPKT_MSGID_ENABLE_CHARACTERISTIC_NOTIFICATIONS branch in processPacket() -
+    // but only once a client has actually subscribed to Indoor Bike Data. A frame pushed
+    // at connect time, before the client has even discovered the characteristic, made
+    // MyWhoosh disable 0x2AD2 outright and fall back to CSC with no power source, even
+    // when the payload was truthful. Observed 2026-08-25 against MyWhoosh on Windows 11.
 }
 
 void DirconProcessor::tcpDisconnected() {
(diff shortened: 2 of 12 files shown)


cagnulein:
thanks i will check it asap!

cagnulein:
Some good news! I tried birefly rouvy and mywhoosh using qz on the same mac and they work beafutiully! I will play tonight with apple tv!

cagnulein:
Great news! Both zwift, rouvy and mywhoosh works on apple TV! Great job!
I will merge this tomorrow!


Roberto Viola

please support my time:
https://paypal.me/cagnulein

http://robertoviola.cloud
http://qzfitness.com/

Il giorno mer 26 ago 2026 alle 02:05 nickolas122 ***@***.***>
ha scritto:

nickolas122:
I'm glad it worked! I will open a PR with the DIRCON lifetime refactor tonight, built on top of this.

cagnulein:
Awesome, thank you so much!


Roberto Viola

please support my time:
https://paypal.me/cagnulein

http://robertoviola.cloud
http://qzfitness.com/

Il giorno gio 27 ago 2026 alle 20:02 nickolas122 ***@***.***>
ha scritto:

PR #4987, opened 2026-08-26
Add dynamic Peloton resistance offset tile

lordneon:
## Reference

Reference:
I have an Echelon EX-3 which I use with qz  peloton class tracking feature. The auto resistance following is really handy but I find it a bit frustrating when there are lots of section changes and my resistance changes are quickly lost.  For example, class has resistance of 40, I nudge it up to 44, class goes goes to 42, bike decreases resistance rather than increasing it, so I need to nudge it up again. 

The current lower / average / upper settings are handy, but I wanted more granularity during the ride, particularly I wanted to have a offset for the current ride which I could easily nudge up / down by a point at a time. So I can increase an entire ride by the amount which feels right. 

Code fully AI generated using Claude Sonnet.  I had it test against simulated bike using docker, but I've not tested it on my bike as I use qz on an Android tablet.

## Changes

Adds a new "Peloton Resistance Offset" tile (bike only) so the resistance target picked from an active Peloton class's range can be nudged up or down live during the ride, instead of only being fixed for the whole class via the static `peloton_difficulty` setting.

Example: class segment range is 35-45 (Peloton 0-100 scale), `peloton_difficulty = lower` picks 35 as the target; tapping the new tile's `+` once shifts the ride to 36 for the rest of the class, tracking each subsequent segment's lower bound + offset as the class advances.

Implementation is modeled on the existing Target Power tile's offset mechanism (`trainprogram::adjustPowerOffsetForTrainingProgram`):

- `trainprogram`: new `trainingProgramResistanceOffset`, recomputed from the immutable `loadedRows` snapshot on each adjustment (so repeated taps don't compound/drift), clamped to the valid 0-100 Peloton range, and reconverted to the device's native resistance via the existing `pelotonToBikeResistance()`. Resets to 0 in `restart()`.
- `homeform`: new `peloton_resistance_offset` tile, wired into `Plus`/`Minus` ahead of the existing `target_resistance`/`peloton_resistance` substring-matched branches so it isn't swallowed by them. Since `trainprogram` only reapplies a row's resistance to the device on the next segment transition, `Plus`/`Minus` also pushes the freshly recomputed current-row resistance to the device immediately via `changeResistance()` in the same tap - one write of the already-combined final value, matching how Target Power calls `changePower()` directly rather than waiting on the program to reapply it.
- New settings `tile_peloton_resistance_offset_enabled`/`_order`, added per `src/CLAUDE.md`'s append-only rule in `settings.qml` and `settings-tiles.qml`, with `qzsettings.cpp`'s `allSettingsCount` and `settings-catalog.json`'s `settingCount` kept in sync (this also corrects a pre-existing one-off drift in `settingCount` noticed along the way - it was 970 against 971 actual entries before this change).

Scoped to bike only for now (matches the reporter's use case, an Echelon EX-3); `peloton.cpp` already special-cases `ELLIPTICAL` alongside `BIKE` for the resistance conversion, so extending to elliptical would be a small follow-up if wanted.

## Testing

- Built with a local Docker image mirroring `qmake && make -j4` against current Ubuntu Qt5 dev packages: zero compile errors, zero warnings on any touched line.
- Ran the built binary headless (Xvfb) with a simulated bike device and the new tile enabled: confirmed it renders correctly (label, icon, `-`/`+` buttons, value) and that tapping `+` dispatches into the new handler rather than being swallowed by the pre-existing `target_resistance`/`peloton_resistance` branches, and safely no-ops with no active ride loaded.
- Not yet tested against a live Peloton class / physical Echelon EX-3

### DIFF
--- a/src/homeform.cpp
+++ b/src/homeform.cpp
@@ -6941,7 +6941,7 @@ void homeform::update() {
                 QString::number(((bike *)bluetoothManager->device())->pelotonResistance().max(), 'f', 0));
             if (trainProgram && trainProgram->isStarted() && trainProgram->resistanceOffsetForTrainingProgram() != 0) {
                 this->target_resistance->setSecondLine(
-                    QStringLiteral("%1%2%")
+                    QStringLiteral("%1%2")
                         .arg(trainProgram->resistanceOffsetForTrainingProgram() > 0 ? QStringLiteral("+")
                                                                                     : QStringLiteral(""))
                         .arg(trainProgram->resistanceOffsetForTrainingProgram()));
--- a/src/homeform.cpp
+++ b/src/homeform.cpp
@@ -570,9 +570,6 @@ homeform::homeform(QQmlApplicationEngine *engine, bluetooth *bl) {
     target_peloton_resistance =
         new DataObject(tr("T.Peloton R(%)"), QStringLiteral("icons/icons/resistance.png"),
                        QStringLiteral("0"), false, QStringLiteral("target_peloton_resistance"), 48, labelFontSize);
-    peloton_resistance_offset =
-        new DataObject(tr("Peloton R. Offset"), QStringLiteral("icons/icons/resistance.png"),
-                       QStringLiteral("0"), true, QStringLiteral("peloton_resistance_offset"), 48, labelFontSize);
     target_cadence = new DataObject(tr("T.Cadence(rpm)"), QStringLiteral("icons/icons/cadence.png"),
                                     QStringLiteral("0"), false, QStringLiteral("target_cadence"), 48, labelFontSize);
     target_power = new DataObject(tr("T.Power(W)"), QStringLiteral("icons/icons/watt.png"),
@@ -3085,14 +3082,6 @@ void homeform::sortTiles() {
                 dataList.append(target_resistance);
             }
 
-            if (settings.value(QZSettings::tile_peloton_resistance_offset_enabled,
-                                QZSettings::default_tile_peloton_resistance_offset_enabled).toBool() &&
-                settings.value(QZSettings::tile_peloton_resistance_offset_order,
-                                QZSettings::default_tile_peloton_resistance_offset_order).toInt() == i) {
-                peloton_resistance_offset->setGridId(i);
-                dataList.append(peloton_resistance_offset);
-            }
-
             if (settings.value(QZSettings::tile_target_peloton_resistance_enabled, false).toBool() &&
                 settings.value(QZSettings::tile_target_peloton_resistance_order, 21).toInt() == i) {
                 target_peloton_resistance->setGridId(i);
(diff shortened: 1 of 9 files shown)


cagnulein:
Hi thanks for this could I ask your exact case? I mean once set up you
should not change this during a workout. Is there a specific case?


Roberto Viola

please support my time:
https://paypal.me/cagnulein

http://robertoviola.cloud
http://qzfitness.com/

Il giorno mer 26 ago 2026 alle 22:38 lordneon ***@***.***> ha
scritto:

lordneon:
Thanks for taking a look at the PR.

Currently, during a class they might often say something like, "this should
be difficult but not too difficult". In those types of scenarios, I want to
be able to change the target pelton resistance offset up / down a few
points then try stick with that same offset for the rest of the ride.

I do share the bike with my wife, so i guess I _should_ be utilising
profiles more to switch between the lower/average/upper ranges, but I
should still just really like to be able to do +1/+2/+3 etc.

Thanks

cagnulein:
Thanks, that makes sense.

I remembered that QZ already has a very similar mechanism for ERG workouts: changing the Target Watt during a training program applies a persistent offset to the remaining workout through `adjustPowerOffsetForTrainingProgram()`.

So I think we should follow the same approach here, but probably without adding a completely new tile.

My preference would be to make the existing Target Peloton Resistance / Target Resistance +/- controls apply a persistent resistance offset while a Peloton class is active, in the same way Target Watt already behaves for ERG workouts.

That would give you exactly the +1/+2/+3 behavior you described, while keeping the UI simpler and reusing an existing QZ pattern.

lordneon:
Yes that makes sense, it simplifies it a lot. 

Claude has made those changes, build passed locally. 

I'm happy to test android version if you can provide a debug apk build with a Peloton class.

cagnulein:
It's already building it here
https://github.com/cagnulein/qdomyos-zwift/actions/runs/33070875574


Roberto Viola

please support my time:
https://paypal.me/cagnulein

http://robertoviola.cloud
http://qzfitness.com/

Il giorno gio 27 ago 2026 alle 14:16 lordneon ***@***.***> ha
scritto:

lordneon:
Ah! Great, I'll give it a try when I get home this evening, will let you know if it all works as expected.

lordneon:
I used a clipboard workout and it worked as expected, so tapping +1 on the target resistance kept that offset on each section.

cagnulein:
that's great! would you like to do other workouts to confirm this or do you think it's safe for merge?

lordneon:
I think it will be fine to merge. I can't test the peloton class due to the missing key in the debug build, but nothing suggests it would behave any differently.

cagnulein:
if you want i can unlock it, simply send me an email to roberto.viola83 at gmail.com indicating this ticket

PR #5005, opened 2026-08-30
Use HR power and manual resistance for TX-500MB IRON

cagnulein:
Reference: support request from Thomas Schwindt about FTMS power and TX-500MB IRON manual resistance

## Summary

- use `wattFromHR(true)` for generic FTMS bikes when **Ignore built-in wattage** is enabled
- only apply the fallback when both the external power sensor and external cadence sensor are disabled
- preserve the existing LYDSTO/DMASUN behavior
- support manual resistance and the custom resistance/watt table for `TX-500MB IRON`
- avoid applying the gear modifier twice when the custom resistance/watt table is enabled

## Why

Some FTMS bikes provide cadence and a calculated/built-in watt value even though the user wants QZ to derive power from an external heart-rate source. Previously, the HR override was limited to LYDSTO/DMASUN, so generic FTMS bikes continued to use their reported wattage despite `watt_ignore_builtin=true`.

The `TX-500MB IRON` has no useful electronic resistance value for QZ's power table, so this PR enables manual resistance mode and uses the configured resistance/watt table while preserving the current cadence-based calculation.

`bike::changeResistance()` already includes `gearsModifier()` when the custom resistance/watt table is enabled. The FTMS update path now sets its additional `gearMultiplier` to `0` whenever `cscbike_custom_resistance_power_table` is enabled, preventing the gear modifier from being applied a second time. The existing Reebok-specific multiplier of `1` remains unchanged when the custom table is disabled.

## Debug evidence

When the HR fallback is active, the log includes:

`Ignoring FTMS built-in wattage: using power from heart rate because power and cadence sensors are disabled`

followed by the calculated HR watt value.

## Test plan

- [x] `git diff --check`
- [x] `qmllint src/main.qml`
- [ ] Verify the TX-500MB IRON HR-power and manual resistance paths on device
- [ ] Verify that one volume gear step maps to one resistance-level step with `cscbike_custom_resistance_power_table` enabled

### DIFF
--- a/src/devices/bluetooth.cpp
+++ b/src/devices/bluetooth.cpp
@@ -2001,6 +2001,7 @@ void bluetooth::deviceDiscovered(const QBluetoothDeviceInfo &device) {
                         (b.name().toUpper().startsWith("ORLAUF_ARES")) ||
                         (b.name().toUpper().startsWith("SPEEDMAGPRO")) ||                        
                         (b.name().toUpper().startsWith("SMARTBIKE-")) ||
+                        (b.name().toUpper().startsWith("TX-500MB IRON")) ||
                         (b.name().toUpper().startsWith("D500V2")) ||
                         (b.name().toUpper().startsWith("FBIKE-HEAVY-PRO")) ||
                         (b.name().toUpper().startsWith("NEO BIKE PLUS ")) ||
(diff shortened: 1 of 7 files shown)


cagnulein:
https://github.com/cagnulein/qdomyos-zwift/actions/runs/33315833722


cagnulein:
https://github.com/cagnulein/qdomyos-zwift/actions/runs/33327546332

cagnulein:
https://github.com/cagnulein/qdomyos-zwift/actions/runs/33398095342

PR #5010, opened 2026-08-31
fix(pitpat): parse S01PRO legacy telemetry

cagnulein:
## Reference

Reference material used for this PR:

- QZ debug log `debug-Mon_Aug_31_18_39_51_2026.log`;
- PitPat Android app 4.21.01 (`com.linzi.sport`), inspected to recover the legacy/three-in-one parser and control commands;
- live Android Bluetooth HCI snoop captured on 2026-09-03 while the official PitPat app was connected to the same `PitPat-S01PRO` bike;
- follow-up live QZ tests confirming speed, cadence and resistance control;
- user calibration measurements collected on 2026-09-14 at several resistance levels.

## Summary

- Adds support for the current `PitPat-S01PRO` legacy/three-in-one telemetry frame.
- Adds the startup handshake used by the official PitPat app before polling or sending resistance commands.
- Keeps the existing 30-byte PitPat parser unchanged for already-supported devices.
- Confirms the resistance command and telemetry mapping against a live HCI capture.
- Adds a provisional S01PRO-specific watt estimate based on cadence and resistance.

## Device / GATT

- Device: `PitPat-S01PRO` (`7C:FE:62:9D:BA:98`).
- Service: `0000FBB0-0000-1000-8000-00805F9B34FB`.
- Write characteristic: `FBB1`.
- Notification characteristic: `FBB2`.

## Startup frame

The bike repeatedly advertises this 34-byte startup frame while it is still in its startup/scan state:

```text
6a 20 b0 2c 1a 00 01 02 03 00 01 00 20
74 38 43 39 31 39 66 4b 6d 64 68 71 7a 36 37 36
20 00 02 eb 43
```

The frame has a valid XOR CRC (`eb`). Fields recovered from the PitPat APK parser are:

- minimum resistance: bytes `9..10`, big-endian;
- maximum resistance: bytes `11..12`, big-endian;
- serial: bytes `13..28`;
- firmware version: byte `29`;
- unit: byte `30`;
- model: byte `31`.

For this bike the startup frame reports `min=1`, `max=32`, model `2`.

## Required startup handshake

Before the official PitPat app starts normal polling or sends resistance commands, it answers the `0x2c` startup frame with:

```text
app -> bike
6a 05 50 1b 01 4f 43
```

The bike acknowledges it with:

```text
bike -> app
6a 06 b0 1b 01 00 ac 43
```

Only after this exchange does the bike leave its startup/scan state and begin normal telemetry:

```text
6a 1c b0 02 ...
```

The implementation now waits for this startup path, sends the handshake, accepts the ACK or first valid telemetry frame as confirmation, resets detection on reconnect, and falls back to the existing PitPat behavior when the legacy frame is not seen.

## Telemetry mapping confirmed by HCI snoop

The live capture contains 555 changing telemetry frames from the official PitPat app session. The mapping is:

- resistance: byte `5`;
- elapsed/activity time: bytes `8..9`;
- distance: bytes `10..11`, divided by `1000` for km;
- calories: bytes `12..13`, divided by `10`;
- crank revolutions: bytes `16..17`;
- heart rate: byte `18`;
- velocity: bytes `23..24`, divided by `1000` for km/h;
- cadence: byte `25`;
- run ID / connection status: bytes `26` / `27`.

Bytes `23..24` are not watts. No native watt field has been identified in the S01PRO legacy frame.

## Provisional power estimate

Direct connection to MyWhoosh showed a cadence-only value of approximately `cadence * 4`, unchanged by resistance. The user confirmed this behavior at resistance levels 1, 2, 20, 28 and 32. Because that value is not physically meaningful, QZ intentionally does not reproduce it directly.

For live testing we agreed to make power depend on both cadence and resistance, with a linear resistance scale across all 32 levels:

```text
resistanceScale = 0.10 + 0.90 * ((resistance - 1) / 31)
watts = cadence * 4 * resistanceScale
```

Examples at 100 RPM:

- resistance 1: about 40 W;
- resistance 32: about 400 W;
- intermediate resistance levels are linearly interpolated.

This keeps the maximum-resistance result aligned with the observed MyWhoosh value while avoiding the unrealistic case where resistance 1 and resistance 32 produce the same watts.

The estimate is used only on the legacy S01PRO path and only when no external power sensor is configured. An external power sensor still takes precedence.

## Resistance control confirmed by HCI snoop

The existing QZ command matches the official PitPat app:

```text
6a 06 51 82 01 <level> <crc> 43
```

`<level>` is clamped to `1..32`. The HCI capture contains 56 resistance changes and all received an ACK, with subsequent telemetry confirming the requested level.

## Compatibility

- The legacy path is selected only when the `0x6a ... 0x2c` startup frame is observed.
- Existing devices that do not emit this frame continue through the existing parser after the short detection window.
- Existing PitPat detection (`PITPAT-S`) and GATT UUIDs are unchanged.
- The provisional power estimate is restricted to the legacy path.
- External power sensors still override the estimated value.

## Validation

Confirmed so far:

- startup handshake request/ACK;
- telemetry starts after the handshake;
- resistance command confirmed against the official app;
- resistance byte mapping confirmed;
- speed, distance and cadence mapping confirmed;
- bytes `23..24` confirmed as velocity rather than watts;
- live QZ test reports working speed, cadence and resistance;
- user confirmed that direct MyWhoosh power is cadence-only and does not react to resistance.

## Remaining live-device validation

A new QZ build should now be tested on the S01PRO to confirm that:

1. MyWhoosh receives non-zero power and the rider moves;
2. watts increase with both cadence and resistance;
3. around 100 RPM the estimate is roughly 40 W at resistance 1 and 400 W at resistance 32;
4. auto resistance still works correctly;
5. no `ServiceError` appears during sustained use;
6. the curve feels plausible enough for use, or can be refined from further measurements or a real power meter.

### DIFF
--- a/src/devices/pitpatbike/pitpatbike.cpp
+++ b/src/devices/pitpatbike/pitpatbike.cpp
@@ -14,6 +14,20 @@
 
 using namespace std::chrono_literals;
 
+namespace {
+uint16_t pitpatReadBE16(const QByteArray &packet, int offset) {
+    return (static_cast<uint16_t>(static_cast<uint8_t>(packet.at(offset))) << 8) |
+           static_cast<uint16_t>(static_cast<uint8_t>(packet.at(offset + 1)));
+}
+
+uint8_t pitpatXorCrc(const QByteArray &packet) {
+    uint8_t crc = 0;
+    for (int i = 1; i < packet.size() - 2; ++i)
+        crc ^= static_cast<uint8_t>(packet.at(i));
+    return crc;
+}
+} // namespace
+
 #ifdef Q_OS_IOS
 extern quint8 QZ_EnableDiscoveryCharsAndDescripttors;
 #endif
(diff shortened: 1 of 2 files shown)


cagnulein:
Calibration update from Pietro's 2026-09-14 measurements and the subsequent email discussion:

The direct MyWhoosh profile reports approximately `power = cadence * 4` regardless of resistance. Pietro confirmed this at multiple resistance levels, but we explicitly agreed not to reproduce that behavior because it is not physically meaningful.

For the QZ test build the agreed provisional curve is now resistance-aware:

```text
resistanceScale = 0.10 + 0.90 * ((resistance - 1) / 31)
watts = cadence * 4 * resistanceScale
```

At 100 RPM this gives about 40 W at resistance 1 and 400 W at resistance 32, with the intermediate 30 levels linearly interpolated. Cadence continues to scale power proportionally.

This supersedes the earlier cadence-only calibration note. The estimate remains restricted to the legacy S01PRO path and an external power sensor still takes precedence.

PR #5014, opened 2026-09-01
Route TOORX iConsole ellipticals to AppGate elliptical

cagnulein:
## Summary

Fix TOORX ellipticals that use the iConsole/AppGate protocol when **iConsole Elliptical** is enabled.

- route `TOORX*` devices to `trxappgateusbelliptical` when `iconsole_elliptical` is enabled
- classify those TOORX devices as `DCT2000I`, like `I-CONSOLE+`, so the driver selects the Microchip transparent UART service/characteristics and uses the DCT/iConsole 21-byte metric layout instead of falling back to the JTX parser
- extend the existing iConsole elliptical discovery test to cover the `TOORX` prefix

Related discussion: https://github.com/cagnulein/qdomyos-zwift/discussions/5011#discussioncomment-18230040

## Routing check

The new elliptical condition is before the later TOORX treadmill/bike/FTMS branches in the `else if` chain. The existing AppGate treadmill and AppGate bike branches are also explicitly guarded by `!iconsole_elliptical`, so with this setting enabled the TOORX device is not claimed by those handlers.

## Log check

The supplied log discovers `TOORX0095` with service `49535343-FE7D-4AE5-8FA9-9FAFD205E455`. The current code incorrectly instantiates `trxappgateusbtreadmill`, which then looks for `0000fff0` and stops with `invalid service`.

`trxappgateusbelliptical` already supports that `49535343-FE7D-4AE5-8FA9-9FAFD205E455` service for the DCT/iConsole protocol, so this fixes the service selection path.

The supplied log was captured with `iconsole_elliptical=false` and the treadmill driver stops before notifications are enabled, so it does not contain the incoming 21-byte metric packets. Therefore the actual TOORX payload values (speed/cadence/watts/resistance) cannot be validated byte-for-byte from this log alone; a log from this PR with iConsole Elliptical enabled will confirm the payload parsing.

### DIFF
--- a/src/devices/bluetooth.cpp
+++ b/src/devices/bluetooth.cpp
@@ -1219,7 +1219,8 @@ void bluetooth::deviceDiscovered(const QBluetoothDeviceInfo &device) {
                 }
                 this->signalBluetoothDeviceConnected(trxappgateusbRower);
             } else if (((b.name().toUpper().startsWith(QStringLiteral("FAL-SPORTS")) && !toorx_bike) ||
-                       (b.name().toUpper().startsWith(QStringLiteral("I-CONSOLE+")) && iconsole_elliptical)) &&
+                       ((b.name().toUpper().startsWith(QStringLiteral("I-CONSOLE+")) ||
+                         b.name().toUpper().startsWith(QStringLiteral("TOORX"))) && iconsole_elliptical)) &&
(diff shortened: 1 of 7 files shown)


cagnulein:
https://github.com/cagnulein/qdomyos-zwift/actions/runs/33487655893

PR #5025, opened 2026-09-03
Horizon ellipticals: real cadence, correct inclination scale, resistance

JSkier21:
The Horizon ellipticals (7.0 AE) speak the same custom FFF0 protocol as the Horizon treadmills, so they already connect via `horizontreadmill`. Three fields in the 0x12 frame were wrong or unread for them.

Offsets were established by sweeping one control at a time on the machine and correlating against its console:

**Inclination is 20 machine levels, not a percentage.** Byte 63 steps by 5 and caps at 100, so `/ 10.0` halved it — console level 2 displayed as 1.0. Verified across a full sweep:

| byte 63 | 5 | 10 | 15 | 50 | 100 |
|---|---|---|---|---|---|
| console level | 1 | 2 | 3 | 10 | 20 |

**Cadence at byte 81 is RPM, 1:1 with the console** — 34 at a console-indicated 4.2 mph, 31 at 3.8. It also has to set `cadenceAvailable`, otherwise the fallback further down synthesises cadence from speed × stride and overwrites it. That fallback is correct for a treadmill (running SPM, ~137) and roughly 4× too high for an elliptical turning at ~31 RPM.

**Resistance at byte 85 is 1:1 with the console's 1–20** and was not read at all. Identified by sweeping each control separately: byte 85 held flat at 1 through a full inclination sweep (byte 63 running 0→100→0), then walked 1..20 and back while inclination stayed parked at 0.

### Detection

All three are gated on the machine being an elliptical, detected from the model name rather than a new setting — Horizon names ellipticals `<model>AE` and treadmills `<model>AT`, the same convention the discovery match already relies on for `HZ_7.0AT-`. Requiring `<digit>.<digit>AE` keeps it off the treadmills and off unrelated names; a real scan during testing contained `FlxTm-AE42`.

### FIT sport

These sessions were also being recorded as treadmill runs. `deviceType()` is deliberately **not** overridden to fix that: `elliptical` and `treadmill` are sibling classes under `bluetoothdevice`, and around twenty call sites (`mainwindow`, `templateinfosenderbuilder`, `peloton`) C-style-cast the device on the reported type, so returning `ELLIPTICAL` from a treadmill object would be undefined behaviour. The sport is corrected where `qfit` consumes it instead, which performs no casts.

### Testing

Verified on a Horizon 7.0 AE. One session produced 136 speed, 136 resistance and 73 cadence samples, with cadence 11–31 rpm against a console showing the same. Treadmill behaviour is unchanged — every change is behind the `AE` name check.


cagnulein:
Hi, thanks it would be better to create a horizon elliptical class. Would you mind to reactor this?

JSkier21:
Refactored as requested — there's now a dedicated `horizonelliptical` class rather than special-casing the ellipticals inside `horizontreadmill`.

**New: `devices/horizonelliptical/`** — subclasses `elliptical`, runs the handshake, parses the 0x12 frame onto elliptical metrics (including peloton resistance).

**New: `devices/horizonprotocol/horizonprofile.{h,cpp}`** — the profile-upload handshake the machine needs before it streams: frame tables, CRC routine and lookup table, moved verbatim out of `horizontreadmill`. `treadmill` and `elliptical` are sibling classes under `bluetoothdevice`, so this can't be shared by inheritance — it's held by composition instead. `handshakeFrames()` returns the sequence as a table, since it turns out to be perfectly regular:

```
7 x (7_N -> 8 x2 -> 9_N -> 10_N -> 11 x17 -> 12 -> 8 x13 -> 13 -> 14) -> 5 polls -> 3 tail frames
```

That let the new driver send byte-identical handshake frames without duplicating ~600 lines of write calls.

**`horizontreadmill` behaviour is unchanged.** Its diff looks large but is mechanical — references qualified onto the extracted profile object, plus two redirected calls. I deliberately kept it that way since I can only test the elliptical:

```
$ git diff upstream/master -- horizontreadmill.cpp | grep '^+' | grep -v <qualification>
+    profile.testProfileCRC();
+    profile.updateProfileCRC();
```

**Both previous workarounds are gone.** No name check inside the treadmill parser, and no FIT sport override in `homeform.cpp` (that file is no longer touched). `deviceType()` now reports `ELLIPTICAL` naturally, so sport, UI labels and the resistance tile all follow — which also resolves the undefined-behaviour concern I raised earlier about overriding `deviceType()` on a treadmill object.

Routing is by model name in `bluetooth.cpp`, ahead of the treadmill branch: Horizon names ellipticals `<model>AE` and treadmills `<model>AT`, the convention the treadmill match already uses for `HZ_7.0AT-`.

Rebased on current master and builds clean for android arm64.

One caveat worth stating plainly: I have a 7.0 AE to test against, but no Horizon treadmill, so the extraction is verified by inspection and compilation rather than on hardware.

cagnulein:
thanks @JSkier21 are you able to test it on your elliptical? Are you on android or ios?

cagnulein (changes requested):
Thanks for the refactor. I reviewed the current version more carefully and I think there are still a few protocol-level risks before merging.

1. The new `horizonprofile::handshakeFrames()` is not actually byte-for-byte equivalent to the existing treadmill initialization. After the five poll frames (`0x09`..`0x0d`), the old `horizontreadmill` code also sends the three final frames `initData2`, `initData3` and `initData4`. These three tail frames are currently missing from `handshakeFrames()`.

2. The ACK/retry behaviour is also different. The original treadmill implementation sends each profile group, waits for the expected packet and retries that same group when `initPacketRecv` is not received. The new elliptical implementation instead sends the sequence continuously and only calls `waitForAPacket()` every 40 frames. A profile group is 38 frames, so those checkpoints are not even aligned to the group boundaries, and there is no retry. This could make initialization unreliable on slower BLE stacks/platforms.

3. Please also check the notification subscription flow. In `horizontreadmill`, `stateChanged()` waits until all services are discovered before subscribing and resets `notificationSubscribed` once. In the new class, every service state change loops over all already-discovered services and may write the FFF0 CCCD again. This could result in multiple `descriptorWritten()` callbacks / multiple initialization attempts depending on service discovery order.

4. The old custom packet assembler uses `payload length + 10`, while the new implementation uses `9 + payload`. Please verify this carefully against the actual protocol before changing the existing framing assumption.

5. Minor: cadence is only updated when `cadence > 0`; if the machine reports 0 when the user stops, QZ may keep showing the last non-zero cadence.

Given how mature and widely used `horizontreadmill` is, I am also wondering whether creating a shared `horizonprotocol` layer is worth the regression risk here. Even if it duplicates some code, I would actually prefer cloning the known-working Horizon handshake/profile code into `horizonelliptical` and leaving `horizontreadmill` completely untouched. In this case the duplication is probably safer than refactoring a working treadmill protocol that we cannot currently regression-test on Horizon treadmill hardware.

So my preference would be: keep the dedicated `horizonelliptical` class, copy the proven initialization code into it, adapt only the elliptical-specific telemetry parsing, and avoid modifying `horizontreadmill` at all. Once the refactored version is tested again on the actual 7.0 AE and CI is green, we can reconsider sharing the protocol later if needed.

JSkier21:
Hi @cagnulein, sorry for the long delay, and thanks for the careful review.

I agree with your preference: keep the dedicated `horizonelliptical` class, copy the proven handshake/profile code from `horizontreadmill` into it, and leave `horizontreadmill` completely untouched. I'll drop the shared `horizonprotocol` layer. On your points:

1. Missing `initData2`/`initData3`/`initData4` tail frames: copying the treadmill init as-is brings them back.
2. ACK/retry: I'll keep the treadmill's per-group send / wait / retry exactly as it is, not the every-40-frames checkpoint.
3. Subscription: I'll subscribe once, after all services are discovered, like `horizontreadmill`, so the FFF0 CCCD is written only once.
4. Framing, `payload + 10` vs `9 + payload`: you're right to flag it. I'll check it against a raw capture from the 7.0 AE before changing anything, and keep the treadmill's `+ 10` unless the capture shows otherwise.
5. Cadence 0: I'll update cadence when the machine reports 0, so it drops to zero when I stop.

To your earlier question: I test on iOS (iPhone) and also have an Android (Unihertz Jelly 2). I'll re-test the reworked class on the actual 7.0 AE before asking you to look again. I've also added the `Reference:` line the new PR check asks for.

Thanks,
Jeremy

PR #5027, opened 2026-09-03
fix: normalize ZBike2.0 FTMS resistance

cagnulein:
## Summary
- Normalize the FTMS resistance reported by the exact Bluetooth device name `ZBike2.0` from raw values 50..200 to user-facing levels 0..15.
- Preserve unchanged resistance behavior for all other FTMS device names.
- Convert the normalized current value back to raw FTMS units only where the existing code reuses current resistance as a physical value.
- Add focused conversion tests.

## Reference
Carlos Alberto

## Test status
- `git diff --check` passed.
- Full qmake/make verification was attempted with `-j1` but interrupted because the build exceeded the available execution window; no passing build/test result is claimed.

No requestResistance == -1 behavior change is included in this PR.

### DIFF
--- a/src/devices/ftmsbike/ftmsbike.cpp
+++ b/src/devices/ftmsbike/ftmsbike.cpp
@@ -10,6 +10,7 @@
 #include <QMetaEnum>
 #include <QSettings>
 #include <QThread>
+#include <QtMath>
 #include <math.h>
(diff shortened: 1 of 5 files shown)


cagnulein:
https://github.com/cagnulein/qdomyos-zwift/actions/runs/33751626668

PR #5028, opened 2026-09-03
feat: map ProForm physical resistance to virtual gears

cagnulein:
## Summary
- Reuse `gears_from_bike` for the NordicTrack/ProForm gRPC bike path.
- Treat small externally observed resistance changes as virtual gear deltas.
- Preserve the resulting offset for later inclination/resistance requests.
- Avoid replaying the previous QZ resistance request when applying a physical delta.

## Behavior
With `gears_from_bike` enabled, a device-reported change such as `5 -> 7` becomes a `+2` virtual gear adjustment. A later nominal target of `5` is therefore combined with the current gear offset and produces `7`.

- Initial/invalid samples do not change gears.
- Consecutive small changes accumulate one step at a time.
- Jumps larger than five resistance levels are treated as resynchronization samples.
- No compensating physical resistance write is issued by the reconciliation path.
- Existing inclination handling continues to apply the current gear value.

## Reference
Support request from Mathieu Bastin (`mi_bastin@jmweston.fr`) about ProForm TDF 10.0 physical resistance controls being exposed as virtual gear changes in MyWhoosh.

## Test plan
- [x] Build QZ library with `make -j1` on the gRPC worktree.
- [x] Build the test runner with `make -j1`.
- [x] Run `PhysicalResistanceGearsTest.*`: 5 passed.
- [x] Run the complete local test runner: exit code 0.
- [x] `git diff --check`.

## Scope
This PR targets the `nordictrack-build-grpc` branch and is intentionally based on that branch rather than `master`.

### DIFF
--- a/src/devices/nordictrackifitadbbike/nordictrackifitadbbike.cpp
+++ b/src/devices/nordictrackifitadbbike/nordictrackifitadbbike.cpp
@@ -256,6 +256,13 @@ nordictrackifitadbbike::nordictrackifitadbbike(bool noWriteResistance, bool noHe
     }
 }
 
+int nordictrackifitadbbike::physicalResistanceGearDelta(resistance_t previous, resistance_t current) {
+    if (previous <= 0 || current <= 0)
+        return 0;
+    const int delta = current - previous;
+    return qAbs(delta) <= 5 ? delta : 0;
+}
+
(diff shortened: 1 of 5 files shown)


cagnulein:
Ref: Maintenance Bastin

PR #5036, opened 2026-09-06
Fix Dynamax FTMS bike detection

cagnulein:
## Summary
- prevent DYNAMAX devices from being classified as KingSmith treadmills when an FTMS bike is explicitly configured
- let the device fall through to the generic FTMS bike handler

The user's latest diagnostic session confirms the misclassification: the exported session is marked as `device="treadmill"` / Run even though the Dynamax is configured as an FTMS bike. In that short 1:53 test QZ reports impossible workout totals (110.9 km and 5645 kcal), while speed and power are otherwise plausible.

This is intentionally a one-line condition change: DYNAMAX treadmill auto-detection now also requires the FTMS bike setting to still be at its default value.


cagnulein:
lbarker6

Issue #578, opened 2022-01-14
[BUG] TOORX SRX-500 (iConsole+ latest version) metrics not updating


travicom75:
**Describe the bug**
Toorx SRX-500 will connect to QDOMYOS app but is not updating metrics (always 0)
BL Device name: TOORX0015

**To Reproduce**
1. wear HR (Toorx supplied one or BL compatible)
2. power on bike
2. wait for cardio data on console screen
3. open QDOMYOS app with BL enabled
4. TOORX0015 will be connected in a while 
5. iConsole data will be greyed out (means BL connection ok..same happens when using other apps)
6. but metrics are not updating (speed, resistance, watt.., HR)
PS: if (as suggested) TOORX iConsole flag (app settings) will be enabled QDOMYOS app will never connect to console

**Expected behavior**
After connected metrics should be updating based on bike data

**Screenshots**

** Spinbike Model **
Spinbike Toorx Srx-500 with iConsole+ firmware version 1.7.37 (224): magnetic resistance are controlled by console (32 levels)
Model is already compatible with ZWIFT but power data is completely wrong, that's why I want to use app in order to try normalize power-factor data.

**Smartphone (please complete the following information):**
 - Device: Honor View10 (BKL-L09)
 - OS: Android 10
 - Version 10.0.0.179

**debug logs**
attached are following logs:
nrfconnect data
QDOMYOS debug log
screenshoots
[debug-Fri_Jan_14_10_46_54_2022.log](https://github.com/cagnulein/qdomyos-zwift/files/7870888/debug-Fri_Jan_14_10_46_54_2022.log)
[debug-Fri_Jan_14_10_48_06_2022.log](https://github.com/cagnulein/qdomyos-zwift/files/7870889/debug-Fri_Jan_14_10_48_06_2022.log)
[debug-Fri_Jan_14_10_51_00_2022.log](https://github.com/cagnulein/qdomyos-zwift/files/7870890/debug-Fri_Jan_14_10_51_00_2022.log)
[debug-Fri_Jan_14_10_51_41_2022.log](https://github.com/cagnulein/qdomyos-zwift/files/7870891/debug-Fri_Jan_14_10_51_41_2022.log)
[debug-Fri_Jan_14_10_54_36_2022.log](https://github.com/cagnulein/qdomyos-zwift/files/7870892/debug-Fri_Jan_14_10_54_36_2022.log)
[...]

PR #5049, opened 2026-09-09
feat: parse SRX-500 FTMS power

cagnulein:
## Summary
- Subscribe to the SRX-500 advertised Fitness Machine Service (0x1826).
- Parse Indoor Bike Data (0x2AD2) instant power, with average-power fallback.
- Prefer a fresh FTMS power value while retaining the existing FFF0 parser as fallback.
- Add focused parser regression tests.

## Reference
Reference: #578

## Evidence
The SRX-500 support log advertises 0x1826 but the existing driver only opens the vendor FFF0 service. The 0x2AD2/0x2A63 values previously visible in the log were QZ virtual-device notifications, not physical-bike notifications.

## Test plan
- `make -f Makefile -j1` (library)
- `make -f Makefile -j1` (test target)
- `./qdomyos-zwift-tests --gtest_filter=ToorxSrx500ParserTest.*` — 3 passed
- `./qdomyos-zwift-tests` — passed

Hardware validation with a live SRX-500 is still pending; the new runtime logs identify the discovered FTMS characteristics and whether 0x2AD2 notifications are available.


cagnulein:
https://github.com/cagnulein/qdomyos-zwift/actions/runs/34321503896

cagnulein:
not needed, it's a full ftms

PR #5051, opened 2026-09-09
Fix Horizon custom protocol speed unit handling

cagnulein:
## Summary

Horizon proprietary packet parsing previously assumed incoming speed was always in mph and always multiplied by 1.60934. A Horizon 7.4AT configured for metric units sends an already-km/h value, so 10 km/h was displayed as 16.0934 km/h.

Reference: Kyle D.

This PR:
- Reuses the existing sole_treadmill_miles setting for both proprietary Horizon speed packet formats.
- Keeps the default true, preserving existing Horizon behavior.
- Exposes the same setting in Horizon Treadmill Options as Horizon speed is in miles.
- Lets metric Horizon users turn it off to keep the decoded km/h value unchanged.
- Leaves FTMS parsing and global miles_unit behavior unchanged.

Outgoing Horizon speed commands were inspected and intentionally left untouched: the established proprietary command path encodes requested speed in the treadmill protocol's expected mph representation, and changing it without device-specific evidence could send an incorrect speed.

## Validation

- git diff --check passes.
- Focused arithmetic checks confirm 10.0 becomes 16.0934 when enabled and 10.0 when disabled.
- Existing desktop build was started; the cached target began rebuilding a large set of unrelated generated sources, so it was stopped before completion. No Horizon parser unit tests exist in the repository; existing Horizon tests cover device discovery/configuration.

### DIFF
--- a/src/devices/horizontreadmill/horizontreadmill.cpp
+++ b/src/devices/horizontreadmill/horizontreadmill.cpp
@@ -1559,6 +1559,10 @@ void horizontreadmill::characteristicChanged(const QLowEnergyCharacteristic &cha
         settings.value(QZSettings::heart_rate_belt_name, QZSettings::default_heart_rate_belt_name).toString();
     bool treadmill_direct_distance =
         settings.value(QZSettings::treadmill_direct_distance, QZSettings::default_treadmill_direct_distance).toBool();
+    const double horizon_speed_multiplier =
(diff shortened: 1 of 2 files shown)


cagnulein:
https://github.com/cagnulein/qdomyos-zwift/actions/runs/34356168399

PR #5052, opened 2026-09-09
Add optional Alternative Pace tile for running sensors

cagnulein:
## Summary

Adds an optional `Alternative Pace` tile for treadmill sessions. The tile shows pace calculated from the connected running sensor (for example HRM Pro Plus or Stryd) while keeping the treadmill's primary speed/pace metric unchanged.

## User request reference

Requested by the user in the email thread from `ogperalta@hotmail.com`, subject: `Visualizar Ritmo de trotadora y también ritmo de HRM Pro Plus`.

The request was to display the sensor-derived pace in real time alongside the treadmill pace. The neutral name `Alternative Pace` is intentional because the source may be HRM Pro Plus, Stryd, or another compatible running sensor.

## Implementation

- Parses and retains RSC running-sensor speed in a dedicated field, without writing to the inherited treadmill `Speed` metric.
- Adds the `Alternative Pace` tile, disabled by default and configurable from Tiles settings, including display order.
- Shows `N/A` when no valid alternative sensor speed is available.
- Updates the persistent settings catalog and backend defaults.

## Verification

- JSON settings catalog count and entries validated.
- `git diff --check` passed.
- Compiled the affected `qzsettings`, `homeform`, and `strydrunpowersensor` objects successfully (warnings are pre-existing project warnings).

### DIFF
--- a/src/devices/bluetooth.h
+++ b/src/devices/bluetooth.h
@@ -183,6 +183,7 @@ class bluetooth : public QObject, public SignalHandler {
     bluetoothdevice *device();
     bluetoothdevice *externalInclination() { return eliteRizer; }
     bluetoothdevice *heartRateDevice() { return heartRateBelt; }
+    strydrunpowersensor *runningPowerSensor() { return powerSensorRun; }
     QList<QBluetoothDeviceInfo> devices;
     bool onlyDiscover = false;
     volatile bool homeformLoaded = false;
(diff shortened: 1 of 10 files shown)


cagnulein:
https://github.com/cagnulein/qdomyos-zwift/actions/runs/34369669253

PR #5056, opened 2026-09-10
Recognize CARE Cardio Master 55430-2 as SportsPlus bike

cagnulein:
## Summary
- route BLE device `CARE11216+0156` to the existing `sportsplusbike` handler
- keep the existing CARE bike behavior for FFF2 writes and the CARE telemetry parser
- use the standard SportsPlus init packet `40 00 16 0a 60` only for `CARE11216+0156`
- preserve the existing CARE CV-351 init sequence and the generic CARE rower classification for other CARE devices

## Context
The CARE Fitness Cardio Master 55430-2 advertises the proprietary `0xFFF0` service. Hardware testing confirmed that this device exposes `FFF1` and `FFF2`, matching the same UUID family used by the existing CARE Fitness bike support.

With the first version of this PR, QZ successfully detected the bike as `sportsplusbike`, connected to FFF0, enabled notifications on FFF1, and wrote to FFF2. However, the existing CARE CV-351 init packet `40 00 9a 56 30` produced no notification packets from the bike and every wait-for-response timed out.

## Test change
For this exact BLE name only, the PR now tries the alternate SportsPlus initialization packet:

`40 00 16 0a 60`

The device still remains in `carefitness_bike` mode, so the write characteristic and CARE packet parser are unchanged. The purpose of this iteration is to determine whether the alternate initialization starts notifications from the Cardio Master.

## Scope
- exact BLE match only: `CARE11216+0156`
- existing `CARE9040177` / CARE Fitness CV-351 behavior remains unchanged
- longer CARE rower names remain on the rower path
- no generic CARE protocol behavior is changed

## Validation goal
On the next hardware test, the log should show `40 00 16 0a 60` as the init packet. The key result is whether any incoming `characteristicChanged` / `<< ...` packets begin arriving after initialization.


cagnulein:
https://github.com/cagnulein/qdomyos-zwift/actions/runs/34486120496

PR #5058, opened 2026-09-10
horizontreadmill: stop reconnect storms exhausting the Android BLE stack

JSkier21:
Three related faults in the Horizon custom (FFF0) connection path, all captured on a **Horizon 7.0 AE with a Unihertz Jelly 2**. Throughout testing the same machine held a BLE link for 75s+ from a Linux host with zero errors, so the peripheral itself was never at fault.

### 1. Unbounded profile-upload retry

Each profile group retries via `goto initN` when the machine doesn't acknowledge it, with no attempt limit. A machine that stops acknowledging gets the same group re-sent indefinitely:

```
33,494 writes in 71s  =  468 writes/sec
876 repeats of a single profile group
```

Now capped at 3 attempts per group, then logs and continues.

### 2. Reconnect storm exhausting the Android BLE stack

Reconnects fired immediately with no backoff, on a controller that was never closed. After a single drop:

```
+  0.0s   CONNECTED OK
+164.8s   Error(133)     <- link drops
+165.5s   Error(133)     ... 38 consecutive, ~2/sec
          never recovers for the rest of the session
```

`disconnectFromDevice()` is now called before retrying, and attempts back off 1/2/4/8/16s, resetting on success.

### 3. `btinit()` re-entrancy guard

`btinit()` blocks for a long time inside nested event loops (`writeCharacteristic()` and `waitForAPacket()` both call `loop.exec()`), during which the refresh timer keeps firing `update()`. Added a guard so a second handshake can't start while one is running.

**To be clear: this third one is defensive hardening, not a fix for an observed failure.** I initially believed I had evidence of concurrent `btinit()` runs, but that turned out to be a miscount on my part and I could not reproduce it. Happy to drop it if you'd prefer the PR stay strictly to what's demonstrated.

### Result

| | before | after |
|---|---|---|
| `Error(133)` | 38 | 0 |
| disconnects | 39 | 0 |
| session | dead after first drop, repeated Start presses | stable for the full run |

Treadmill behaviour is otherwise untouched — this changes only the retry limits and the reconnect path. Tested on Android/arm64 against a 7.0 AE.

### DIFF
--- a/src/devices/horizontreadmill/horizontreadmill.cpp
+++ b/src/devices/horizontreadmill/horizontreadmill.cpp
@@ -88,6 +88,18 @@ void horizontreadmill::waitForAPacket() {
 }
 
 void horizontreadmill::btinit() {
+    if (btinitRunning)
+        return;
+    btinitRunning = true;
+    initRequest = false;
+
+    // Each profile group below retries via "goto initN" when the machine does not acknowledge it.
+    // That retry was unbounded: if a machine stops acknowledging, btinit() re-sends the same group
+    // forever. Captured in the field at 468 writes/sec - 33,494 writes in 71s, 876 repeats of a
+    // single group - until the link collapsed. Cap the retries and carry on instead.
+    int initRetry[9] = {0};
+    const int INIT_MAX_RETRY = 3;
+
     QSettings settings;
     QStringList horizon_treadmill_profile_users;
     horizon_treadmill_profile_users.append(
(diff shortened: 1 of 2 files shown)


cagnulein:
Is this specific issue on Linux only? Did you test it on android or ios?

JSkier21:
Hello,

I have an old Jellyfin Android; building and testing it is easier.
Although, I think I may be in over my head here with the home elliptical -
the Horizon has been a big headache (I also can't get an ESP32 bridge on it
either).

So, specifically regarding Android itself.
--
*Jeremy MountainJohnson*
***@***.*** ***@***.***>*

cagnulein:
So did you already test this patch?


Roberto Viola

please support my time:
https://paypal.me/cagnulein

http://robertoviola.cloud
http://qzfitness.com/

Il giorno gio 10 set 2026 alle 21:55 Jeremy MountainJohnson <
***@***.***> ha scritto:

PR #5075, opened 2026-09-13
Match WLT-BK FTMS startup and resistance protocol

cagnulein:
## Summary

- detect `WLT-BK-*` with a dedicated `WLT_BK` flag
- automatically apply the FTMS-only subscription/init path that was validated with the Hammer Racer setting
- match the `WLT-BK-*` startup sequence observed in a Kinomap Android HCI capture: `REQUEST_CONTROL -> STOP -> START`
- route WLT-BK resistance through the existing one-byte `FTMS_SET_TARGET_RESISTANCE_LEVEL` path, scaling the requested level by 10 as observed in Kinomap
- clamp resistance through the existing `max_resistance` handling, with the bike range set to 24

## Capture findings

The Kinomap HCI snoop contains a useful A/B comparison:

- `00 -> 08 01` leaves the bike stopped and Indoor Bike Data remains all zero
- after reconnect, `00 -> 08 01 -> 07` is acknowledged and the bike emits Fitness Machine Status `0x04` (Started/Resumed), followed by real speed/cadence/power data

Kinomap does not write to the proprietary `FFC2` or `FFF2` characteristics.

For resistance, Kinomap uses writes such as:
- `04 5d`
- `04 70`
- `04 84`
- `04 97`
- `04 ab`

The bike's `2AD6` Supported Resistance Level Range reads `0a 00 f0 00 0a 00`, corresponding to 1..24 with step 1.

## Test

The user confirmed that enabling the Hammer Racer setting makes the `WLT-BK-5310` start and report metrics correctly. This PR now applies the relevant FTMS-only subscription and initialization automatically for `WLT-BK-*`, so Hammer Racer should no longer be required.

Expected startup sequence:

`00 -> 08 01 -> 07`

followed by `2ADA 04` and non-zero `2AD2` values while pedaling.


cagnulein:
Reference jan.piet.barend.spekman

PR #5122, opened 2026-09-18
Fix incline compensation overriding forcespeed rows

cagnulein:
## Summary
Prevent treadmill incline-compensation from overriding training-program speed targets.

## Problem
A reported NordicTrack NTL15010 workout used short rows with explicit speed + inclination. The workout first requested the correct speed, then the inclination-change compensation recalculated it (for example 10 km/h -> 4.6 km/h).

This overlaps with the race condition addressed in #4553. That PR already gives a training-program `changeSpeed` a grace period before HR PID can adjust speed, but the separate inclination-compensation block could still change the speed during that same settling period.

## Change
Two cases are protected:

1. A non-HR row with `forcespeed=1` (`zoneHR < 0`) keeps its explicit speed for the whole row, so inclination compensation cannot replace it with a wattage-preserving speed.
2. After any training-program speed change, inclination compensation waits for the same `delta` grace period already used by #4553 before it can act.

For HR-driven rows, the forced speed is therefore applied at row entry, allowed to settle, and then the HR PID can resume modulation from that new speed target. Inclination compensation also becomes eligible again after the grace period unless `PID Ignore Inclination` is enabled.

## Validation
The reported short-interval rows (`forcespeed=1`, `zoneHR=-1`) no longer get their explicit speeds overwritten by inclination compensation. Structured HR rows that set an initial forced speed preserve that new target during the #4553 grace period, then resume normal PID behavior.


cagnulein:
let's see if it's a real requirement for the user

